PR #5, opened 2018-08-21
Feature/pelagos 3243 move metadata status to dataset

snguyen1:


src/Pelagos/Bundle/AppBundle/Controller/UI/DatalandController.php
```diff
@@ -88,7 +88,7 @@ public function defaultAction($udi)
      *
      * @param string $udi The UDI of the dataset to return metadata for.
      *
-     * @throws BadRequestHttpException When the metadata for the dataset has not been accepted.
+     * @throws BadRequestHttpException When the dataset status is not accepted accepted.
```
mickel1138: Double accepteD?

PR #8, opened 2018-08-22
Feature/pelagos 3296 modify rabbitmq publish datasets symfony command

praneethpr:
To test: 
Please use the Datacite EZ API credentials 
bin/console dataset-doi:pub-all-appr-nonres-datasets accepted
or
bin/console dataset-doi:pub-all-appr-nonres-datasets submitted


praneethpr:
@fightingtexasaggie : I think we need to make some temporary changes, because we shouldn't be publishing datasets which are not accepted.

praneethpr:
Instead of modifying existing symfony command which publishes approved datasets, I thought i'll create a new pre-migration doi symfony command. So, i'm going ahead and deleting this branch

PR #9, opened 2018-08-24
Added the removal of non numeric characters in the string

praneethpr:


src/Pelagos/Bundle/AppBundle/Handler/UploadHandler.php
```diff
@@ -381,7 +381,7 @@ protected function removeDir($dir)
      */
     protected function toBytes($str)
     {
-        $val = trim($str);
+        $val = preg_replace('/\D/', '', trim($str));;
```
mickel1138: Cool, I didn't know what \D did... But is is add \S also in brackets like [\S\D] you don't have to use the trim.

mickel1138: Actually, the \S is not needed either. You do it all with the existing preg_replace, just remove the trim.

snguyen1:
There are a few gcs errors in this file.

PR #11, opened 2018-08-24
Feature/pelagos 3299 report for existing datasets i e udi dataset status doi status doi

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/ReportDoiStatusDatasetStatusCommand.php
```diff
@@ -0,0 +1,198 @@
...
+     * @param OutputInterface $output An class that handles output files.
+     *
+     * @return Dataset
+     */
+    protected function openIO(OutputInterface $output)
+    {
+        if ($this->fileOutput === null) {
+            $handle = fopen($this->outputFileName, 'w');
+            $this->fileOutput = new StreamOutput($handle);
+            $this->entityManager = $this->getContainer()->get('doctrine.orm.entity_manager');
```
mickel1138: You store the entityManger for the entire class in a protected variable, but only use it here once. So this could be a local scope variable. And the property can be removed.

src/Pelagos/Bundle/AppBundle/Command/ReportDoiStatusDatasetStatusCommand.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Executes the current command.
+     *
+     * @param InputInterface  $input  An InputInterface instance.
+     * @param OutputInterface $output An OutputInterface instance.
+     *
+     * @throws Exception Exception thrown when openIO function fails to generate report.
+     *
+     * @return integer Return 0 on success, or an error code otherwise.
```
mickel1138: The return is always only 0. There are no other returns here. The fact that it throws an error is described in @throws.

mickel1138: The return is always only 0. There are no other returns here. The fact that it throws an error is described in @throws.

src/Pelagos/Bundle/AppBundle/Command/ReportDoiStatusDatasetStatusCommand.php
```diff
@@ -0,0 +1,198 @@
...
+
+        return $datasets;
+    }
+
+    /**
+     * Generate report for doi migration.
+     *
+     * @param array $datasets Collection of datasets.
+     *
+     * @return integer Return 0 on success, or an error code otherwise.
```
mickel1138: Same here, return only returns 0.

src/Pelagos/Bundle/AppBundle/Command/ReportDoiStatusDatasetStatusCommand.php
```diff
@@ -0,0 +1,198 @@
...
+        }
+        return 0;
+    }
+
+    /**
+     * Prepare the output file for writting.
+     *
+     * @param OutputInterface $output An class that handles output files.
+     *
+     * @return Dataset
```
mickel1138: Datasets? Or array or Dataset(s)?

src/Pelagos/Bundle/AppBundle/Command/ReportDoiStatusDatasetStatusCommand.php
```diff
@@ -0,0 +1,198 @@
...
+     *
+     * @param Dataset $dataset The dataset instance.
+     *
+     * @throws Exception Exception thrown when doi metadata method fails.
+     *
+     * @return array
+     */
+    private function getDoiStatus(Dataset $dataset)
+    {
+        $doiStatus = null;
```
mickel1138: Not used anywhere.

PR #15, opened 2018-08-30
Added post-migration script which recreates DOIs for DIF approved datasets

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/DoiDatacitePostMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $datasets = $entityManager->getRepository(Dataset::class)->findBy(array(
+            'metadataStatus' => array(
+                DatasetSubmission::METADATA_STATUS_NONE
+            )
+        ));
+
+        foreach ($datasets as $dataset) {
+            // Creating DOI's which are DIF approved datasets.
+            if ($dataset->getIdentifiedStatus() === DIF::STATUS_SUBMITTED) {
```
fightingtexasaggie: Why aren't we querying for DIF::STATUS_APPROVED instead?

praneethpr: Good catch! :)

src/Pelagos/Bundle/AppBundle/Command/DoiDatacitePostMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+            ->setDescription('Recreate DOIs for DIF approved datasets.');
+    }
+
+    /**
+     * Executes the current command.
+     *
+     * @param InputInterface  $input  An InputInterface instance.
+     * @param OutputInterface $output An OutputInterface instance.
+     *
+     * @throws \Exception When dataset not found.
```
mickel1138: These exceptions are not in this functions.

src/Pelagos/Bundle/AppBundle/Command/DoiDatacitePostMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        $datasets = $entityManager->getRepository(Dataset::class)->findBy(array(
+            'metadataStatus' => array(
+                DatasetSubmission::METADATA_STATUS_NONE
+            )
+        ));
+
+        foreach ($datasets as $dataset) {
+            // Creating DOI's which are DIF approved datasets.
+            if ($dataset->getIdentifiedStatus() === DIF::STATUS_APPROVED) {
+                $doi = $dataset->getDoi();
```
mickel1138: Instead of doing all this, I would just publish the issue command for the consumer.

`$this->get('old_sound_rabbit_mq.doi_issue_producer')->publish($dif->getDataset()->getId(), 'issue');`

mickel1138: If you look at issueDoi on the doiConsumer you can see that this is nearly the same code.

praneethpr: Yeah, the pre-migration script needed changes in the consumer, so i called the Util methods directly from the script, followed the same here :P. Calling the consumer makes more sense in this script. 👍

src/Pelagos/Bundle/AppBundle/Command/DoiDatacitePostMigrationCommand.php
praneethpr: Removed it

src/Pelagos/Bundle/AppBundle/Command/DoiDatacitePostMigrationCommand.php
```diff
@@ -0,0 +1,61 @@
...
+                DatasetSubmission::METADATA_STATUS_NONE,
+            )
+        ));
+
+        $rabbitProducer = $this->getContainer()->get('old_sound_rabbit_mq.doi_issue_producer');
+        foreach ($datasets as $dataset) {
+            // Creating DOI's which are DIF approved datasets.
+            if ($dataset->getIdentifiedStatus() === DIF::STATUS_APPROVED) {
+                $rabbitProducer->publish($dataset->getId(), 'issue');
+                $output->writeln('Attempting to issue DOI for Dataset ' . $dataset->getId());
```
mickel1138: This message should maybe now say "adding to queue" or something.

praneethpr: Hmm.. I think its fine, just a one time script and that message is just for us

PR #18, opened 2018-09-04
Added error handling for the exceptions thrown for DIF

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
```diff
@@ -606,59 +607,99 @@ function updateDIF(form)
...
+                    OK: function () {
+                        $(this).dialog("close");
+                        scrollToTop();
+                        treeFilter();
+                        return $.Deferred().resolve();
+                    }
                 }
-            }
-        });
+            }).parent().addClass("ui-state-error");
```
mickel1138: Instead of using this class (which is reserved for jQuery validate, just change the imgInfo in the message to another icon.

mickel1138: That way you can also just use one of these dialog sections.

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
```diff
@@ -606,59 +607,99 @@ function updateDIF(form)
...
+            } else if (status.statusCode == 2) {
+                var title = "DIF Updated and Approved";
+                var message = '<div><img src="' + imgInfo + '">' +
+                    "<p>The application with DIF ID: " + udi + " was successfully updated and approved!" +
+                    "<br></p></div>";
+            }
+        } else if (status.message === "error") {
+            if (status.statusCode == 3) {
+                var title = "Unable to approve DIF";
+                var message = '<div><img src="' + imgInfo + '">' +
```
mickel1138: User imgWarning here, for example. Or another icon.

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
praneethpr: Removed this

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
praneethpr: Changed it to ImgCross, tried to scale the height and width, doesn't look good. So just showing the regular cross image. I think it should be fine 👍

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
```diff
@@ -651,14 +672,15 @@ function updateDIF(form)
...
-                OK: function() {
+                OK: function () {
                     $(this).dialog("close");
                     scrollToTop();
                     treeFilter();
                     return $.Deferred().resolve();
                 }
             }
         });
+
```
mickel1138: Could delete...

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
```diff
@@ -648,12 +650,12 @@ function updateDIF(form)
         } else if (status.message === "error") {
             if (status.statusCode == 3) {
                 var title = "Unable to approve DIF";
-                var message = '<div><img src="' + imgCross + '">' +
+                var message = '<div><img src="' + imgCancel + '">' +
```
mickel1138: 😀 Much nicer!

PR #22, opened 2018-09-07
Bugfix/pelagos 3307 gml namespace prefix missing

mickel1138:
For task PELAGOS-3307

### DIFF
--- a/src/Pelagos/Util/GmlUtil.php
+++ b/src/Pelagos/Util/GmlUtil.php
@@ -9,13 +9,21 @@ class GmlUtil
     /**
      * This function add namespace for validation to the given gml.
      *
-     * @param string $gml        Gml that needs namespace.
-     * @param array  $namespaces Array of attributes and values.
+     * @param string     $gml        Gml that needs namespace.
+     * @param array|null $namespaces Array of attributes and values, can be left null for std namespaces.
      *
      * @return string GML string with namespace.
      */
-    public static function addNamespace($gml, array $namespaces)
+    public static function addNamespace($gml, array $namespaces = null)
     {
+        if ($namespaces == null) {
+            $namespaces = array(
+                'xmlns:gml' => 'http://www.opengis.net/gml/3.2',
+                'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance',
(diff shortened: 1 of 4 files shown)


mickel1138:
test

praneethpr:
Please provide more information about on how to test this, there is not much information specified on the JIRA bug

mickel1138:
@praneethpr To test we have to check the log, @fightingtexasaggie should be able to help you with that.
Before this fix, warning should appears in the Postgres log file, after this they will be no more warning.

src/Pelagos/Util/GmlUtil.php
```diff
@@ -0,0 +1,54 @@
...
+     * This function add namespace for validation to the given gml.
+     *
+     * @param string     $gml        Gml that needs namespace.
+     * @param array|null $namespaces Array of attributes and values, can be left blank for std namespaces.
+     *
+     * @return string GML string with namespace.
+     */
+    public static function addNamespace($gml, array $namespaces = self::GML_NAMESPACES)
+    {
+        if (($gml != null) and (1 == preg_match('/gml/', $gml))) {
```
praneethpr: Type comparision

src/Pelagos/Util/GmlUtil.php
```diff
@@ -0,0 +1,54 @@
...
+                $strNameSpaces = '';
+                foreach ($namespaces as $key => $value) {
+                    $strNameSpaces .= ' ' . $key . '="' . $value . '"';
+                }
+                $regEx = '/^<gml:\S*/';
+                $gml = preg_replace($regEx, "$0$strNameSpaces", $gml);
+            }
+            return $gml;
+        } else {
+            return $gml;
```
praneethpr: If and else return the same variable, probably remove the else statement and add the return at the end of the function?

mickel1138: That's what I get for not looking at cut and paste code.

praneethpr: Haha :)

src/Pelagos/Util/GmlUtil.php
mickel1138: You could use !== for the first one, but null does have type. Also preg_match can return a boolean, that would cover this case also. But I don't think it would return true, only false.

PR #26, opened 2018-09-13
Rebuild data discovery - feature/PELAGOS-3306

snguyen1:
Notes:
+ The tree stills work (we may need to discuss whether we need to rebuild the tree function or not).
+ Geometry is not working (on hover and show spatial): This will be done in another subtask.


src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -138,45 +353,30 @@ function showDatasets(by,id) {
...
-            if (myGeoViz.getFilter()) {
-                $("#clearGeoFilterButton").button("enable");
-            }
+            //this trigger infinites scrolling
+            $(".viewport").has(".datasets").scroll(function(){
+                if ($(this).scrollTop() + $(this).innerHeight() >= $(this)[0].scrollHeight) {
+                    loadData(by, id);
+                }
+            });
+            // if (myGeoViz.getFilter()) {
```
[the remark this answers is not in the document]
snguyen1: This one is purposely left out so I can work on the other sub-task (fixing geometry) later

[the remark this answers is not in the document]
praneethpr: 👍

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -99,8 +106,213 @@ function resizeLeftRight() {
...
 function showDatasets(by,id) {
+    var filter = jQuery("#filter-applied").val();
+    geo_filter = "";
+    if (trees["tree"].geo_filter) {
+        geo_filter = trees["tree"].geo_filter;
+    }
+
+    //reset existing data to prepare for a new search
+    currentIndex = 0;
+    datasetList[0] = [];
```
praneethpr: Similarly, even here you can just reset the whole array `datasetList = []`

snguyen1: If I don't explicitly initialize/reset them as a new list/array in here it would break the page (seems to be JS thing)

praneethpr: Yeah, I tried it too.. I'm not really sure how the global variable is working. It should also throw the same undefined error, but its not. Seems weird.

src/Pelagos/Bundle/AppBundle/Controller/UI/DataDiscoveryController.php
```diff
@@ -3,6 +3,7 @@
 namespace Pelagos\Bundle\AppBundle\Controller\UI;
 
 use Elastica\ResultSet;
+use Symfony\Component\HttpFoundation\JsonResponse;
```
mickel1138: I would like to see an extra line here, between use statements.

src/Pelagos/Bundle/AppBundle/Controller/UI/DataDiscoveryController.php
```diff
@@ -40,22 +38,22 @@ public function defaultAction(Request $request)
             array(
                 'defaultFilter' => $request->query->get('filter'),
                 'pageName' => 'data-discovery',
-                'download' => false,
```
mickel1138: We use the dangling comma, we use it in other places, so we have to be consistent.

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -99,8 +106,213 @@ function resizeLeftRight() {
...
+        var row = $(activeTabSelector + " tr:eq(" + currentIndex + ")");
+        row.attr("udi", data["udi"]);
+        row.attr("datasetid", dataset["_id"]);
+
+        var rowContent = createRow(data, row);
+        row.html(rowContent);
+        row.show();
+        currentIndex++;
+        //add the data to the data array
+        datasetList[activeTabIndex][dataset["udi"]] = dataset;
```
praneethpr: dataset["udi"] seems to be undefined. I think it is data["udi"] just like LIne no.172 ?

snguyen1: good catch!!!!!!

src/Pelagos/Bundle/AppBundle/Resources/views/DataDiscovery/template-dataset-row.html.twig
```diff
@@ -0,0 +1,25 @@
+<table hidden>
```
mickel1138: the hidden tag is new in HTML 5. Using this tag may not give the desired result. When cloned, this tag will be copied too. You'd have to remove it, but the same would go when using a class.

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -138,45 +349,30 @@ function showDatasets(by,id) {
                             }
                         }
                     }
+                    var activeTabIndex = getActiveTabIndex();
+                    currentIndex = (datasetList[activeTabIndex].length === 0) ? 0 : datasetList[activeTabIndex].length - 1;
```
praneethpr: Is this check required? because currentIndex and datasetList values are reset in this same function

snguyen1: Yes because the indices are different for every tab. This code is in the activate function which will triggers when switching the tabs.

praneethpr: I think the use of global variables is not encouraged..  it will become confusing for us, when we look at the code sometime later

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -99,8 +106,213 @@ function resizeLeftRight() {
     $("#right").height(rh);
 }
 
+function formatBytes(bytes,decimals) {
+    if(bytes == 0) return '0 Bytes';
+    var k = 1024,
+        dm = decimals <= 0 ? 0 : decimals || 2,
+        sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'],
```
praneethpr: I think for consistency we decided to use double quotes in Javascript

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -99,8 +106,213 @@ function resizeLeftRight() {
...
+        i = Math.floor(Math.log(bytes) / Math.log(k));
+    return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
```
praneethpr: Same here

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -99,8 +106,213 @@ function resizeLeftRight() {
...
+
+    //this shows the details dataset on Show Details button
+    $(rowContent).find(".details_link").bind("click", function(){
+        var row = $(this).closest("tr");
+        if ($(row).has(".details:empty").length == 1) {
+            var datasetId = $(row).attr("datasetid");
+            $.ajax({
+                "url": Routing.generate("pelagos_app_ui_datadiscovery_showdetails", { "id": datasetId }),
+                "success": function(data) {
+                    $(row).find('.details').html(data);
```
praneethpr: Double quotes here, and the below statements

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
snguyen1: yup copy/paste code

src/Pelagos/Bundle/AppBundle/Resources/views/DataDiscovery/template-dataset-row.html.twig
fightingtexasaggie: Performance statistics using Firefox's bandwidth simulator, time in seconds. 1st column original, 2nd column this PR
:
Regular 3G
100 20
100 20
100 18

Good-3G
50  10
50  11
50  11

Regular 4G/LTE
21  4
20  5
20  5

DSL
38  8
37  8
38  8

WiFi
7   3
7   3
7   3

Unthrottled
6   3
6   3
6   3

~

PR #31, opened 2018-09-19
Hotfix/pelagos 3314 strip non ascii on 500

mickel1138:
Not very pretty, suggestions welcome.


src/Pelagos/Util/DOIutil.php
```diff
@@ -367,4 +375,56 @@ function ($matches) {
...
+
+        // Replace Middle Dot
+        $search[]  = chr(194) . chr(183);
+        $replace[] = '*';
+
+        // Apply Replacements
+        $string = str_replace($search, $replace, $string);
+
+        // Remove any non-ASCII Characters
+        $string = preg_replace("/[^\x01-\x7F]/", '', $string);
```
fightingtexasaggie: The brute squad!

PR #32, opened 2018-09-20
Hotfix/pelagos 3313 create doi mint create function in doi utility

praneethpr:


praneethpr:
It looks like we have conflicting paths, we can't merge this hotfix directly into develop which will undo changes that are in develop

PR #37, opened 2018-09-27
Feature/pelagos 3245 removed unused metadata variables

praneethpr:


snguyen1:
Note: Needs to run a migration to test this

PR #39, opened 2018-09-27
Added migration to remove metadata entity

praneethpr:


praneethpr:
@fightingtexasaggie Good find!

praneethpr:
@fightingtexasaggie It was the order of the migrations, which was making the udi-prefix to fail. I've fixed it

PR #46, opened 2018-10-02
Epic/pelagos 3217 task that need to be done right after integrated submission is completed

praneethpr:


praneethpr:
@fightingtexasaggie Yeah, it makes sense to remove them. I've made the changes

PR #50, opened 2018-10-04
add udi

snguyen1:


src/Pelagos/Bundle/AppBundle/Resources/views/DataDiscovery/template-dataset-row.html.twig
```diff
@@ -14,6 +14,7 @@
         </a>.
         <span id="container-dataset-research-group"><span id="dataset-research-group"></span>. </span>
         <span id="container-dataset-doi"><span id="label-dataset-doi">DOI: </span><a href="http://dx.doi.org/{{ 'placeholder-doi' }}" target="_blank" id="dataset-doi"></a>. </span>
+        <span id="container-dataset-udi"><span id="label-dataset-udi">UDI: <a href="https://data.gulfresearchinitiative.org/data/{{ 'placeholder-udi' }}" target="_blank" id="dataset-udi"></a>. </span></span>
```
praneethpr: We should probably not hardcode the hyper link and use path variable like the hyperlink for title?

snguyen1: that can be done, but I'm not sure if the "pelagos-symfony" will be in the generated url on production server

PR #51, opened 2018-10-08
bugfix/datadiscovery-fix-emptyavailable-remotelyhosted-dataset-filesize

snguyen1:
- Fixing the empty file size [Size : ] on dataset row
- replace jquery closest with jquery parents 
- Show zero result tabs (like pre-optimization behavior)
- Fixing multiple binding issue on Show Details click event


snguyen1:
I had another commit on: Not hiding tabs with zero results (since it's the behavior we had before the optimization)

PR #53, opened 2018-10-09
Feature/pelagos 3334 make select options sortable

mickel1138:


mickel1138:
Used: https://github.com/NikitaKA/jquery.dragOptions

mickel1138:
May not work on IE! (eek)

praneethpr:
The user-story mentions about all other metadata editor tools which i'm not aware off other than submission and review which uses this keyword section

mickel1138:
I changed this from the drag-drop to simple up-down buttons.

src/Pelagos/Bundle/AppBundle/Resources/public/js/dataset-submission.js
```diff
@@ -727,6 +727,18 @@ $(function() {
                 source.append(option);
                 source.append(sortOptions(source.find("option").detach()));
             }
+        } else if ($(event.currentTarget).text() == "up") {
```
praneethpr: Could we assign an id to the button element and use the identifier, instead of the value?

src/Pelagos/Bundle/AppBundle/Resources/public/js/dataset-submission.js
```diff
@@ -727,6 +727,18 @@ $(function() {
...
+            var selectedOption = target.find("option:selected");
+            var prevOption = selectedOption.prev("option");
+            if (prevOption.is("option")) {
+                selectedOption.detach().insertBefore(prevOption);
+            }
+        } else if ($(event.currentTarget).text() == "down") {
```
praneethpr: Use '===' :)

mickel1138: This was just copied from the lines above. Maybe we have to change it everywhere?
I'm hesitant to change it for a language like Javascript using jQuery.  $.text() always returns a string, is it really necessary to check to type too.  Triple equals are good if you can expect something else, and want to make sure.

mickel1138: I guess === is better.

PR #54, opened 2018-10-10
Feature/pelagos 3335 make remove copy selected keyword into input

mickel1138:
Double clicking a selected keyword, or clicking remove will copy the content into the input box to edit.


mickel1138:
@praneethpr yes, before double click would only remove it. You can use enter on the input box to add, move it to the list, and double click on any select list to move from one side to another.

snguyen1:
It works!!! simple, yet very efficient.

PR #57, opened 2018-10-12
Release/5.4.0

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/BackFillApprovedDateDifCommand.php
```diff
@@ -0,0 +1,103 @@
+<?php
```
praneethpr: BackFill Script won't be run in the 5.4.0 deployment, we have created a DMS task for it. Only after it is fixed and tested thoroughly we will run this script

src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -112,39 +358,42 @@ function showDatasets(by,id) {
...
-            }
-            );
+            });
             enableFilterButton();
             $("#clear-button").button("enable");
             $("#drawGeoFilterButton").button("enable");
-            if (myGeoViz.getFilter()) {
+            //this triggers infinite scrolling
+            $(".viewport").has(".datasets").scroll(function(){
+                if ($(this).scrollTop() + $(this).innerHeight() >= $(this)[0].scrollHeight) {
```
mickel1138: scroll doesn't fire at certain resolutions, I fixed it by subtracting a from scrollHeight like this:
`$(this)[0].scrollHeight-10)`

As discussed here: https://stackoverflow.com/questions/3898130/check-if-a-user-has-scrolled-to-the-bottom

mickel1138: Actually -1 seems to works just about right.

PR #60, opened 2018-10-18
Added status to the update DOI method to fix the auto-publish bug in …

praneethpr:
…Datacite EZ API

### DIFF
--- a/src/Pelagos/Util/DOIutil.php
+++ b/src/Pelagos/Util/DOIutil.php
@@ -218,6 +218,10 @@ class DOIutil
         $input .= 'datacite.publisher:' . $this->escapeSpecialCharacters($publisher) . "\n";
         $input .= "datacite.publicationyear:$publicationYear\n";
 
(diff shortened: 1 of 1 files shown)


src/Pelagos/Util/DOIutil.php
```diff
@@ -218,6 +218,10 @@ public function updateDOI(
         $input .= 'datacite.publisher:' . $this->escapeSpecialCharacters($publisher) . "\n";
         $input .= "datacite.publicationyear:$publicationYear\n";
 
+        //Adding status to the update
+        $doiMetadata = $this->getDOIMetadata($doi);
```
mickel1138: Shouldn't we get the status from dataset instead?

praneethpr: If we get the status from the dataset, it is redundant to the publish method and we need to validate the doi status that needs to be set based on the dataset status and restriction, which already exists in the DoiConsumer class

PR #61, opened 2018-10-19
Pelagos 3338 create comparison view for the different dataset versions

mickel1138:


src/Pelagos/Bundle/AppBundle/Resources/views/SideBySide/index.html.twig
```diff
@@ -0,0 +1,159 @@
...
+{% block body %}
+
+<div>
+<h1>UDI: <span class="udi-title"></span></h1>
+<table class="side-table">
+    <tr>
+        <td colspan="8" width="100%">
+            <label>UDI:</label>
+            <input name="udi" size="100">
+            <button id="get-versions-button" type="button">Get UDI</button>
```
praneethpr: Can be disabled if no input is entered

src/Pelagos/Bundle/AppBundle/Resources/public/js/sidebyside.js
```diff
@@ -0,0 +1,95 @@
...
+            .find(".submission-status")
+            .text($(this).find("option:selected").data("status"));
+        $(this).parents("div.left-version")
+            .find(".submission-modificationtimestamp")
+            .text($(this).find("option:selected").data("modificationtimestamp"));
+        $(this).parents("div.left-version")
+            .find(".submission-modifier")
+            .text($(this).find("option:selected").data("modifier"));
+        
+        $("#left").load("/pelagos-symfony/dev/mvde/sidebyside/getForm/" + udi + "/" + version, function() {
```
praneethpr: Your url is hard-coded, so it is throwing a no route found  "NotFoundHttp" Exception

mickel1138: Fixed it

src/Pelagos/Bundle/AppBundle/Resources/views/SideBySide/index.html.twig
mickel1138: This can be done in a future improvement.

src/Pelagos/Bundle/AppBundle/Resources/views/SideBySide/extent.html.twig
```diff
@@ -0,0 +1,106 @@
...
+                    {% endif %}
+                </td>
+            </tr>
+            <tr class="spatialExtras">
+                <td colspan="2">
+					<span class="helpSection">
+						{% image '@PelagosAppBundle/Resources/public/images/info32.png' %}
+						<img class="info" src="{{ asset_url }}">
+						{% endimage %}
+						<div class="tooltiptext">
```
praneethpr: None of the tooltip text seems to appear on the browser

src/Pelagos/Bundle/AppBundle/Controller/UI/SideBySideController.php
```diff
@@ -0,0 +1,206 @@
...
+     * @Method("POST")
+     *
+     * @return Response A Response instance.
+     */
+    public function getVersions(Request $request)
+    {
+        $udi = $request->request->get('udi');
+
+        try {
+            $datasetSubmissionHistory = $this->getDatasetSubmissionHistory($udi);
```
praneethpr: Should we not get drafts? probably check for status incomplete?

snguyen1:
The dropdowns' height looks a bit short and doesn't look good on my screen. the current value is 20px, I think we can set it to 25px in the css
```css
select {
    height:25px;
}
```

src/Pelagos/Bundle/AppBundle/Controller/UI/SideBySideController.php
mickel1138: We want to get all of them. Including drafts.

PR #68, opened 2018-10-23
Bugfix/pelagos 3344 throwing proper exceptions for dif api calls

snguyen1:
Use more proper exception
Add error message for undefined error

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
+++ b/src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
@@ -611,11 +611,11 @@ function updateDIF(form)
                 type: "PATCH",
                 datatype: "json",
                 data: formData
-            }).success(function(json, textStatus, jqXHR) {
+            }).done(function(json, textStatus, jqXHR) {
                 if (jqXHR.status === 204) {
                     response.status = "success";
                 }
-            }).error(function (json, text, jqXHR) {
+            }).fail(function (json, text, jqXHR) {
                 var errorMessage = JSON.parse(json.responseText);
                 response.status = "error";
                 response.message = errorMessage.message;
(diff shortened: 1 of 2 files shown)


src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
```diff
@@ -660,6 +660,10 @@ function updateDIF(form)
                     "<p>The application with DIF ID: " + udi + " failed to complete action!" +
                     "<br></p></div>";
             }
+        } else {
+            var title = "Unable to process DIF form";
+            var message = "<div><p>There was an error processing your request. Your session might have expired.<br>" +
+                "If the problem still persists after you re-login, please contact the administrator</p></div>";
```
mickel1138: needs period

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
```diff
@@ -611,11 +611,11 @@ function updateDIF(form)
                 type: "PATCH",
                 datatype: "json",
                 data: formData
-            }).success(function(json, textStatus, jqXHR) {
+            }).done(function(json, textStatus, jqXHR) {
                 if (jqXHR.status === 204) {
                     response.status = "success";
                 }
-            }).error(function (json, text, jqXHR) {
+            }).fail(function (json, text, jqXHR) {
```
praneethpr: is there a reason to change error to fail? because the response.message or status attribute is empty when it is used

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
```diff
@@ -660,6 +660,10 @@ function updateDIF(form)
                     "<p>The application with DIF ID: " + udi + " failed to complete action!" +
                     "<br></p></div>";
             }
+        } else {
```
praneethpr: So irrespective of any error, it always comes to this block.

src/Pelagos/Bundle/AppBundle/Resources/public/js/dif.js
snguyen1: .error is a deprecated according to jquery doc

praneethpr: I think the response.status and message is not supported by the .fail() jquery, and we need to use `.fail(function( jqxhr, textStatus, error ) {
    var err = textStatus + ", " + error;
    console.log( "Request Failed: " + err );
});`
just an example. probably you can look into it more.

PR #72, opened 2018-10-30
feature/PELAGOS-3351-add-affiliation-to-emails-sent-from-griidc-staff 

snguyen1:
Test this in 3 cases:
- Make a submission as a GRIIDC staff only
- Make a submission as a Researcher from a research group only
- Make a submission as a member of both GRIIDC and a Researcher Group

### DIFF
--- a/src/Pelagos/Entity/Person.php
+++ b/src/Pelagos/Entity/Person.php
@@ -766,6 +766,20 @@ class Person extends Entity
     }
 
     /**
+     * Get a list of the names of all Data Repositories this person is associated with.
+     *
+     * @return array
+     */
+    public function getDataRepositoryNames()
+    {
+        $dataRepositoryNames = array();
+        foreach ($this->personDataRepositories as $personDataRepository) {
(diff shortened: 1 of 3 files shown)


src/Pelagos/Entity/Person.php
```diff
@@ -765,6 +765,20 @@ public function getDataRepositories()
...
+    /**
+     * Get a list of the names of all Data Repositories this person is associated with.
+     *
+     * @return array
+     */
+    public function getDataRepositoryNames()
+    {
+        $dataRepositoryNames = array();
+        foreach ($this->personDataRepositories as $personDataRepository) {
+            $dataRepositoryNames[] = $personDataRepository->getDataRepository()->getName();
```
praneethpr: We have only GRIIDC as the data repository, shouldn't this just be a string instead of an array

praneethpr: If we add another data repository, I think we should re-architect the structure of entities. I think returning an array is confusing now.

mickel1138: I'm ok with this. The twig is a little confusing, but it's fine.

PR #83, opened 2018-11-07
Feature/pelagos 3372 delete splash twig and remove authentication

praneethpr:


snguyen1:
Is there other subtask for this? The popup loads a new home page when I click download. Not sure what the behavior will be in prod mode

mickel1138:
I have not been able to fully test this, but I don't this this will work. At least the httpAction on the download controller has a authentication check, that will throw an error.

mickel1138:
@praneethpr also, in anonymous, getUser will just get "guest" or "anonymous", I forgot. We may have to tweak that a bit.

PR #84, opened 2018-11-07
Feature/pelagos 3373 add attribute accepteddate in dataset entity

praneethpr:


src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -2373,6 +2373,9 @@ protected function updateDatasetStatus()
     {
         if ($this->getDataset() instanceof Dataset) {
             $this->getDataset()->setDatasetStatus($this->getDatasetStatus());
+            if ($this->getDatasetStatus() === Dataset::DATASET_STATUS_ACCEPTED) {
```
fightingtexasaggie: This could be a re-submission, and an update being made via a new dataset review. This should only set the accepted date if the newest submission is being accepted, I think.

praneethpr: Good catch!

praneethpr: Fixed it 👍

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -1194,6 +1194,7 @@ public function reviewEvent(Person $reviewer, $eventName)
                 //Setting the status to in-review.
                 $this->status = self::STATUS_COMPLETE;
                 $this->setDatasetStatus(Dataset::DATASET_STATUS_ACCEPTED);
+                $this->getDataset()->setAcceptedDate(new \DateTime('now', new \DateTimeZone('UTC')));
```
mickel1138: I feel these three calls should move Dataset Entity, and just have one function name acceptDataset that can check if a dataset is in state to be acceptable, and then set the date, and status of itself. Dataset status and setAcceptedDate should be private functions. However, then there should also be function called rejectDataset and such. Maybe an improvement?

praneethpr: Yeah, it think you are right. Previously we had the status tied up to the DatasetSubmission entity, now we've moved it to the dataset. Should we add it in the issues in github? to keep track of it.

PR #85, opened 2018-11-08
 bugfix/PELAGOS-3370-catch-gml-conversion-errors

snguyen1:
- On empty GML (gmltoWkt): Return false on promise resolve to abort .then
- Fix undefined message console log on Dif ajax setup


mickel1138:
I feel this bug is not all Javascript, but more the API function not returning a better error message?

PR #89, opened 2018-11-14
Fix deprecated 3.4

praneethpr:


mickel1138:
Checked out the branch, everything seems fine.

praneethpr:
I found a bug in data-discovery, the last tab wasn't getting populated because it didn't have the Dataset Submission object (approved DIFs tab) and javascript was throwing "undefined" error. Previously elastic used to return an empty array for this, so we wrote our logic based on that. I've flipped the logic to check if there is a dataset submission object before we access other attributes from that.

fightingtexasaggie:
After completion, please update https://griidc.atlassian.net/browse/ACTION-874

fightingtexasaggie:
Everything still seems to be working, except when I issue a fos:elastica:reset command I get the following error:

In Http.php line 181:
                                                                                                                                                        
  No handler for type [text] declared on field [year] [reason: Failed to parse mapping [dataset]: No handler for type [text] declared on field [year]]

praneethpr:
> Everything still seems to be working, except when I issue a fos:elastica:reset command I get the following error:
> 
> In Http.php line 181:
> 
> No handler for type [text] declared on field [year] [reason: Failed to parse mapping [dataset]: No handler for type [text] declared on field [year]]

Did you test it on griidc-dev2 or griidc-dev?

PR #91, opened 2018-11-15
Replaced auditReader with raw sql and used dataset submission audit

praneethpr:


praneethpr:
Test Dataset (really old and accepted twice)
R1.x135.118:0004
https://griidc.atlassian.net/browse/MDREV-7499

PR #92, opened 2018-11-19
Feature/pelagos 3387 add current sequence

mickel1138:
Added dataset status, including overal status, and status per submission.
And count of revisions.

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Controller/UI/SideBySideController.php
+++ b/src/Pelagos/Bundle/AppBundle/Controller/UI/SideBySideController.php
@@ -98,8 +98,13 @@ class SideBySideController extends UIController
             $data['modifier'] = $submission->getModifier()->getLastName() .
                 ', ' . $submission->getModifier()->getFirstName();
             $data['modificationtimestamp'] = $submission->getModificationTimeStamp()->format('c');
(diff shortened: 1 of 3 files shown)


src/Pelagos/Bundle/AppBundle/Controller/UI/SideBySideController.php
```diff
@@ -98,8 +98,13 @@ public function getVersions(Request $request)
             $data['modifier'] = $submission->getModifier()->getLastName() .
                 ', ' . $submission->getModifier()->getFirstName();
             $data['modificationtimestamp'] = $submission->getModificationTimeStamp()->format('c');
+            $data['datasetstatus'] = Dataset::DATASET_STATUSES[$submission->getDatasetStatus()];
```
praneethpr: I don't think displaying the submission's dataset status was part of this user-story

praneethpr: Probably we can do it another user-story? or we can ask the PO to add in this story

mickel1138: Ok, as per PO, removed for now.

PR #97, opened 2018-11-26
Changed style to look like production map style (orange).

mickel1138:


src/Pelagos/Bundle/AppBundle/Resources/public/js/gMap.js
```diff
@@ -18,39 +18,60 @@
...
 
         var feature = format.readFeature(wkt, {
             dataProjection: "EPSG:4326",
             featureProjection: "EPSG:3857"
         });
-            
+
+        var style = new ol.style.Style({
+            fill: new ol.style.Fill({
+              color: 'rgba(255, 255, 255, 0)'
```
praneethpr: Use double quotes

fightingtexasaggie: eslint complains about single quotes used in this section, otherwise looks good.

PR #99, opened 2018-11-27
Feature/pelagos 3394 add json ld to dataland

mickel1138:
Make sure the dataset is approved an available. Then simply "view source" and look for the JSON-LD.


fightingtexasaggie:
Tested and verified additional text is no longer being injected above the desired added header. Renders well in Firefox, Chrome, and IE Edge.

PR #101, opened 2018-11-28
Changed the logic to get the spatial extent from the datasetSubmissio…

praneethpr:
…n object


praneethpr:
@snguyen1 Thanks Son! I added it to the user story, so that we can do it when we deploy it

PR #102, opened 2018-11-29
Feature/pelagos 3369 manifest generator

fightingtexasaggie:


fightingtexasaggie:
Sample output data (and input) is on Jira the Jira ticket. Format looks a little different than William's example, but this might (or might not) be OK.

praneethpr:
@fightingtexasaggie : The destination branch needs to be changed to develop

fightingtexasaggie:
William is OK with the format, as he noted in https://griidc.atlassian.net/browse/PELAGOS-3348. Please review now.

share/python/directory_tree_node.py
```diff
@@ -0,0 +1,215 @@
+from __future__ import print_function
+import sys
+import os
+import platform
+
+'''
+@author: jvh
```
fightingtexasaggie: We generally don't put author in code comments. This is provided by github.

PR #107, opened 2018-12-06
Added the timestamp download report and refactored to reduce redundancy

praneethpr:


src/Pelagos/Bundle/AppBundle/Controller/UI/DatasetDownloadReportController.php
```diff
@@ -200,4 +180,204 @@ private function formatSizeUnits($fileSizeBytes)
 
         return $fileSizeBytes;
     }
+
+    /**
+     * This defaultAction generates the form to select the date range for the report.
```
fightingtexasaggie: I think this is from the other method.

src/Pelagos/Bundle/AppBundle/Controller/UI/DatasetDownloadReportController.php
```diff
@@ -200,4 +180,204 @@ private function formatSizeUnits($fileSizeBytes)
...
+     * @return float
+     */
+    private function getFileSize(Dataset $dataset): float
+    {
+        $fileSize = 0;
+        $datasetSubmission = $dataset->getDatasetSubmission();
+        if ($datasetSubmission instanceof DatasetSubmission) {
+            $fileSize = $this->formatSizeUnits($datasetSubmission->getDatasetFileSize());
+        }
+
```
fightingtexasaggie: If datasetFileSize is null, this will still actually return null despite initializing it to fileSize = 0 because the call to getDatasetFileSize will return null.

PR #108, opened 2018-12-06
"Added" spatial extent field, make it un-hidden.

mickel1138:


src/Pelagos/Bundle/AppBundle/Controller/UI/DatasetReviewController.php
```diff
@@ -257,6 +258,21 @@ protected function makeSubmissionForm($udi, Dataset $dataset = null, DatasetSubm
...
+                'input-xml' => true,
+                'output-xml' => true,
+                'indent' => true,
+                'indent-spaces' => 4,
+                'wrap' => 0,
+            ),
+            'utf8'
+        );
+
+        $datasetSubmission->setSpatialExtent($gml);
```
praneethpr: I don't think this is setter is needed.

praneethpr: Okay, I see it the difference now. The gml looks very much readable on the Dataset-review tool with the formatting. I think the tidy_parse_string should be in the Entity when the spatial extent is set.

PR #109, opened 2018-12-06
Added file name, size and hash to review.

mickel1138:
PELAGOS-3412 will modify the same file, so watch out for merge conflicts.


praneethpr:
@mickel1138 Do we need to add on-change javascript? Because when the file gets updated, the information still points to the previous informaiton

mickel1138:
@fightingtexasaggie this is the correct behaviour, if there is not hash, the submission can not be reviewed.

PR #114, opened 2018-12-11
Feature/pelagos 3420 update template

fightingtexasaggie:
I also used this data use policy on the external splash screen, but kept the line in place "This particular dataset is not hosted directly by GRIIDC, so additional terms and conditions may be imposed by the hosting entity." This likely needs to be approved upstream, but the external splash screen was a mess.

To faciliatate testing, Y1.x122.045:0001 is remotely hosted.

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Resources/views/Download/download-external-resource-splash-screen.html.twig
+++ b/src/Pelagos/Bundle/AppBundle/Resources/views/Download/download-external-resource-splash-screen.html.twig
@@ -14,16 +14,20 @@
 
         <div style="border: 1px solid #aaa; padding: 10px; margin-top: 20px; border-radius: 4px;">
             <p style="margin-top:0">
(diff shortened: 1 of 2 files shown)


src/Pelagos/Bundle/AppBundle/Resources/views/Download/download-external-resource-splash-screen.html.twig
```diff
@@ -12,48 +12,49 @@
...
+                This particular dataset is not hosted directly by GRIIDC, so additional terms and conditions may be
+                imposed by the hosting entity.
+            </p>
+            <div style="text-align:center;">
+                <p style="margin-bottom:0">
+                    <button class="close_button" style="font-weight:bold;margin:10px 10px 10px 0px;">Cancel</button>
+                </p>
+            </div>
+        </div>
+        &nbsp;
```
mickel1138: &nbsp; ? Use the style! (padding-bottom)

PR #115, opened 2018-12-12
Feature/pelagos 3421 adj tmplts mv citation suggestion

fightingtexasaggie:


src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/details.html.twig
```diff
@@ -1,3 +1,8 @@
+<h2>Suggested Citation:</h2>
+<div>
```
mickel1138: Should it be on the way top of this section?

PR #118, opened 2018-12-12
Added file info to dataland template.

fightingtexasaggie:
I confirmed the custom twig extension that we use to format bytes into human-readable form is indeed based on powers of 10, so that simplified things.

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/details.html.twig
+++ b/src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/details.html.twig
@@ -30,6 +30,14 @@
         </div>
     {% endif %}
 
+    {% if dataset.datasetSubmission.datasetFilename|default %}
+        <h2>Filename:</h2>
+        <div>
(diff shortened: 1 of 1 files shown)


src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/details.html.twig
```diff
@@ -30,6 +30,14 @@
         </div>
     {% endif %}
 
+    {% if dataset.datasetSubmission.datasetFilename|default %}
+        <h2>Filename:</h2>
+        <div>
+            {{ dataset.datasetSubmission.datasetFilename|default }} ({{ dataset.datasetSubmission.datasetFileSize|formatBytes }})
```
praneethpr: Should file size be under a separate heading?

fightingtexasaggie:
I'll ask the P.O. Maybe I went too far in cleaning it up.

mickel1138:
I don't mind the file size on the same line, but it's easy to miss. Either make it bold, or a different font, or like @praneethpr said, different heading might be best.

PR #119, opened 2018-12-17
Only suggest citation for accepted datasets.

fightingtexasaggie:


fightingtexasaggie:
To test, from data discovery, choose dataland page from accepted tab, and other from the in-review tab.

PR #120, opened 2018-12-18
Added erddap url attribute to the entity and twig template

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/DatasetReview/distribution-info.html.twig
```diff
@@ -33,6 +33,25 @@
...
+            <td>
+				<span class="helpSection">
+					{% image '@PelagosAppBundle/Resources/public/images/info32.png' %}
+                        <img class="info" src="{{ asset_url }}">
+                    {% endimage %}
+                    <div class="tooltiptext">
+						Please provide the ERDDAP Url.
+					</div>
+				</span>
+                {{ form_row(form.erddapUrl) }}
```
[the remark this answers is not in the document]
praneethpr: I've fixed it.

snguyen1:
a strange use case causes duplicate error messages.
Follow these steps to reproduce:
- Enter a good url (eg: https://data.gulfresearchinitiative.org/)
- Enter some random character to make it a 404 site (https://data.gulfresearchinitiative.orgeee)
- Remove focus on the box
- Now remove the colon (:) out of the url
- Remove focus again

src/Pelagos/Bundle/AppBundle/Form/DatasetSubmissionType.php
```diff
@@ -262,6 +262,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'delete_empty' => true,
                 'required' => true,
             ))
+            ->add('erddapUrl', Type\TextType::class, array(
+                'label' => 'ERDDAP Url',
+                'required' => false,
+                'attr' => array('data-rule-url' => true),
```
mickel1138: What does this do?

praneethpr: It is a jquery validate plugin for html 5 data attributes

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -1004,6 +1013,7 @@ public function __construct(Entity $entity, PersonDatasetSubmissionDatasetContac
             $this->setTemporalExtentNilReasonType($entity->getTemporalExtentNilReasonType());
             $this->setDistributionFormatName($entity->getDistributionFormatName());
             $this->setFileDecompressionTechnique($entity->getFileDecompressionTechnique());
+            $this->setErddapUrl($entity->getErddapUrl());
```
mickel1138: It is debatable if this is needed, because the URL could get invalidated after another submissions. Maybe a questions for the PO?

praneethpr: The only way to set this field is through Dataset-review tool, and if we don't add it in here, the data from a previous submission would be lost. For example: If dataset in accepted status, with an existing erddapUrl is opened in dataset-review tool, the new object would have an empty erddapUrl attribute.

src/Pelagos/Bundle/AppBundle/Form/DatasetSubmissionType.php
mickel1138: I guess I should have known that.

src/Pelagos/Entity/DatasetSubmission.php
mickel1138: Right again!

src/Pelagos/Util/UrlValidation.php
```diff
@@ -28,7 +28,8 @@ public function validateUrl($url)
 
         //check to see if it worked.
         if (200 !== $httpCode) {
-            throw new \Exception("url failed with:$httpCode($output)", $httpCode);
+
+            return "Could not get URL, returned HTTP code:$httpCode";
```
snguyen1: I don't think many people know about http code. In my opinion, we can change this to something like "URL is not available (httpCode)".

praneethpr: We'll run it by William and Rosalie about the error message

PR #121, opened 2018-12-19
Bugfix/pelagos 3431 remove foreign key constraint

praneethpr:


src/Pelagos/Bundle/AppBundle/Controller/UI/DatasetReviewController.php
```diff
@@ -380,7 +380,7 @@ private function createNewDatasetSubmission(DatasetSubmission $datasetSubmission
         $datasetSubmission = new DatasetSubmission($datasetSubmission);
         $reviewedBy = $this->getUser()->getPerson();
         $reviewStartTimeStamp = new \DateTime('now', new \DateTimeZone('UTC'));
-        $datasetSubmissionReview = new DatasetSubmissionReview($datasetSubmission, $reviewedBy, $reviewStartTimeStamp);
+        $datasetSubmissionReview = new DatasetSubmissionReview($reviewedBy, $reviewStartTimeStamp);
```
mickel1138: I thought there would be more places, but this is the only one.

app/DoctrineMigrations/Version20181219204218.php
```diff
@@ -0,0 +1,46 @@
...
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE dataset_submission_review DROP CONSTRAINT fk_3fa5c62f8488ba54');
+        $this->addSql('DROP INDEX uniq_3fa5c62f8488ba54');
+        $this->addSql('ALTER TABLE dataset_submission_review DROP dataset_submission_id');
```
fightingtexasaggie: Considering adding SQL to populate the new field before loosing reference to the old one.

praneethpr: I've added that update statement to the migration, thanks mike

src/Pelagos/Bundle/AppBundle/Controller/UI/DatasetReviewController.php
praneethpr: Need to set DatasetSubmissionReview object to DatasetSubmission, and it needs to updated after the DatasetSubmissionReview is created

src/Pelagos/Bundle/AppBundle/Controller/UI/DatasetReviewController.php
```diff
@@ -379,8 +379,6 @@ private function createNewDatasetSubmission(DatasetSubmission $datasetSubmission
         // The latest submission is complete, so create new one based on it.
         $datasetSubmission = new DatasetSubmission($datasetSubmission);
         $reviewedBy = $this->getUser()->getPerson();
-        $reviewStartTimeStamp = new \DateTime('now', new \DateTimeZone('UTC'));
-        $datasetSubmissionReview = new DatasetSubmissionReview($datasetSubmission, $reviewedBy, $reviewStartTimeStamp);
```
mickel1138: Could have left the datasetSubmission in, and then called the setting of the review on it with this.

fightingtexasaggie:
This is now working.

PR #122, opened 2018-12-20
Only produce sitemap for available datasets.

mickel1138:


src/Pelagos/Bundle/AppBundle/Controller/DefaultController.php
```diff
@@ -45,7 +46,9 @@ public function showSiteMapXmlAction()
         $response = new StreamedResponse(function () use ($container) {
             $datasets = $container->get('pelagos.entity.handler')->getBy(
                 Dataset::class,
-                array(),
+                array(
+                    'availabilityStatus' => DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE
```
praneethpr: I guess we should include AVAILABILITY_STATUS_PUBLICLY_AVAILABLE_REMOTELY_HOSTED?

fightingtexasaggie: I'm asking William now.

PR #124, opened 2019-01-03
Added the version 2 for gomri report using new parameters to build qu…

praneethpr:
…eries

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Controller/UI/GomriReportController.php
+++ b/src/Pelagos/Bundle/AppBundle/Controller/UI/GomriReportController.php
@@ -59,7 +59,8 @@ class GomriReportController extends ReportController
             '.csv';
         // Add header to CSV.
         return $this->writeCsvResponse(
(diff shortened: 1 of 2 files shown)


src/Pelagos/Bundle/AppBundle/Controller/UI/GomriReportController.php
```diff
@@ -37,6 +37,24 @@ public function defaultAction()
...
+     * @return Response A Response instance
+     */
+    public function versionTwoReportAction()
+    {
+        if (!$this->isGranted('ROLE_DATA_REPOSITORY_MANAGER')) {
+            return $this->render('PelagosAppBundle:template:AdminOnly.html.twig');
+        }
+        // Add header to CSV.
         return $this->writeCsvResponse(
             $this->getData()
```
snguyen1: You may want to consider setting custom filename for these two reports to avoid future confusion when downloading and comparing two versions. The custom filename is the second param of writeCsvResponse (see Report Controller)

PR #125, opened 2019-01-07
feature/PELAGOS-3437-create-remotely-hosted-datasets-listing-page 

snguyen1:


src/Pelagos/Bundle/AppBundle/Resources/views/List/RemotelyHostedDatasets.html.twig
```diff
@@ -0,0 +1,79 @@
...
+] %}
+
+{% block body %}
+    <h2> Remotely Hosted Datasets</h2>
+    <table
+            id="remotelyHostedDatasetsTable"
+            class="stripe"
+            width="100%"
+            data-ajax='{
+                "url": "{{ path("pelagos_api_datasets_get_collection", {"availabilityStatus": "5,7", "datasetSubmission": "*", "_properties": properties | join(',')}) }}",
```
praneethpr: Should be based on TRANSFER_STATUS_REMOTELY_HOSTED, this gets back only accepted and restricted ones. It doesn't get the datasets which are pending for approval.

PR #126, opened 2019-01-10
Added datasetSubmission Delete API before it does the dataset delete

praneethpr:


mickel1138:
Just FYI. Chrome told me webkitURL is depricated. (line 98)

praneethpr:
Made changes to remove each Dataset Submission object which has bidirectional associations mapped to Dataset Submission entity, before deleting the Dataset. It should work now!

PR #129, opened 2019-01-15
Added needs review case, set it to pending approval and added updateA…

praneethpr:
…vailabilitystatus after Dataset is processed in the consumer


praneethpr:
> Are we still asking about a possible operational task to back-populate this state for those in-review datasets that have a URL?

I checked and there are only 6 such datasets in that state as of now, which could be backfilled.

praneethpr:
> This only works for the new dataset?

You mean, after checking out this branch, and then submitting a dataset? Yes. Not for the ones which are already in that state

PR #136, opened 2019-01-22
feature/PELAGOS-3451-remove-doi-from-unavailable-and-expected-soon-tab-on-data-discovery

snguyen1:


praneethpr:
> This still leaves the DOI in the html sent to the client, albeit hidden. I'd suggest perhaps adding some logic to the view to remove it from what is generated altogether.

I agree, we shouldn't be sending those details to the client and then hiding it. But it looks it has been implemented similarly for other fields, I guess we've to rewrite data-discovery

mickel1138:
> > This still leaves the DOI in the html sent to the client, albeit hidden. I'd suggest perhaps adding some logic to the view to remove it from what is generated altogether.
> 
> I agree, we shouldn't be sending those details to the client and then hiding it. But it looks it has been implemented similarly for other fields, I guess we've to rewrite data-discovery

I'm OK with this if the rewrite is too involving. As we have used this elsewhere.

PR #144, opened 2019-01-23
javascript disables hyperlink on pending-review tab

fightingtexasaggie:
Tested on Firefox/Linux, Chrome/Linux, Edge/Windows-10, Chrome/Windows-10, Firefox/Windows-10 (all current versions)

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
+++ b/src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
@@ -286,10 +286,9 @@ function createRow(data, row)
 
     if (!data["doi"]["doi"]) {
         $(rowContent).find("#container-dataset-doi").hide();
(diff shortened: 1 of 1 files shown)


src/Pelagos/Bundle/AppBundle/Resources/public/js/data-discovery.js
```diff
@@ -286,6 +286,10 @@ function createRow(data, row)
 
     if (!data["doi"]["doi"]) {
         $(rowContent).find("#container-dataset-doi").hide();
+    } else if (activeTabIndex == 2) {
+        // Remove the hyperlink, but keep text for this tab.
+        $(rowContent).find("#dataset-doi").attr("href", $(rowContent).find("#dataset-doi").attr("href")).removeAttr("href");
```
mickel1138: Why not simply removeAttr? Or you can use replaceWith.
(And you won't need line 292).

PR #149, opened 2019-01-31
Feature/pelagos 3467 modify remotely hosted view to properly show test results

fightingtexasaggie:
To test, generate some test data (404, 500, etc):

UPDATE dataset_submission SET dataset_file_url_last_checked_date = now(), dataset_file_url_status_code = '404' WHERE id = (SELECT dataset_submission_id FROM dataset WHERE udi = 'R4.x257.000:0015');

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Resources/public/css/remotelyhosted-datasets.css
+++ b/src/Pelagos/Bundle/AppBundle/Resources/public/css/remotelyhosted-datasets.css
@@ -19,6 +19,10 @@
     box-shadow: 0px 0px 8px rgba(0, 0, 0, 0.3);
 }
 
+.red {
+    background-color: #ff9da7 !important;
+}
+
 fieldset {
     border: 1px dotted #AAA;
 }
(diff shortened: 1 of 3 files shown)


src/Pelagos/Bundle/AppBundle/Resources/public/css/remotelyhosted-datasets.css
```diff
@@ -19,6 +19,10 @@
     box-shadow: 0px 0px 8px rgba(0, 0, 0, 0.3);
 }
 
+.red {
```
mickel1138: A better classname than just "red" may be desired.

mickel1138:
Works, but I feel the red washed out the text a bit. 
Also, when you select/click it, the red highlighting is replaced by normal select highlighting.

praneethpr:
I think this is more of a discussion with the PO, Instead of error codes showing up, status of OK or NOT OK would be much better for understanding it and I dont know whether it is a good practice to just show the status based on color

PR #150, opened 2019-01-31
Feature/pelagos 3470 add cold storage attributes to the entity

fightingtexasaggie:


src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -602,7 +602,25 @@ class DatasetSubmission extends Entity
...
 
-     /**
+    /**
+     * The dataset file cold storage archive size.
+     *
+     * @var integer
+     *
+     * @ORM\Column(type="bigint", nullable=true)
+     */
+    protected $datasetFileColdStorageArchiveSize;
```
praneethpr: Migration needed

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -1689,6 +1707,50 @@ public function getDatasetFileSha256Hash()
...
+    }
+
+    /**
+     * Set the size of the single archive file to be stored in cold storage.
+     *
+     * @param integer $datasetFileColdStorageArchiveSize The archive size, in bytes.
+     *
+     * @return void
+     */
+    public function setDatasetFileColdStorageArchiveSize($datasetFileColdStorageArchiveSize)
```
praneethpr: Should it be initialized to null? because not all datasetSubmission objects would have this value and it would throw an error (must be an instance of integer and null given)

src/Pelagos/Entity/DatasetSubmission.php
fightingtexasaggie: Oops. I've now added.

src/Pelagos/Entity/DatasetSubmission.php
fightingtexasaggie: If copying from another dataset submission, it may or may not be null. I changed the docblock to allow for nulls.

tests/unit/Pelagos/Entity/DatasetSubmissionTest.php
```diff
@@ -323,6 +323,25 @@ public function testDatasetFileAttributes()
         );
     }
 
+    /**
+     * Test dataset cold-storage related file attributes.
+     *
+     * @return void
+     */
+    public function testDatasetFileColdStorageAttributes()
```
snguyen1: Can we change this to test function name to something like the existing ones? Like... testCanSetAndGetDatasetFileColdStorageAttributes (sorry for being nitpicky)

fightingtexasaggie: No problem at all. I'll change.

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -1689,6 +1717,73 @@ public function getDatasetFileSha256Hash()
...
+     * Get the size of the single archive file to be stored in cold storage, in bytes.
+     *
+     * @return integer
+     */
+    public function getDatasetFileColdStorageArchiveSize()
+    {
+        return $this->datasetFileColdStorageArchiveSize;
+    }
+
+
```
mickel1138: one too many lines

PR #152, opened 2019-02-01
Feature/pelagos 3466 write pelagos command that performs these checks

praneethpr:


src/Pelagos/Util/UrlValidation.php
```diff
@@ -20,8 +20,12 @@ class UrlValidation
     public function validateUrl($url)
```
mickel1138: Would we like to make follow location optional, by default true?

src/Pelagos/Util/UrlValidation.php
```diff
@@ -20,8 +20,12 @@ class UrlValidation
...
     {
         $ch = curl_init();
+        $cookies = tempnam('/tmp','cookie.txt');
+        curl_setopt($ch, CURLOPT_COOKIEJAR, $cookies);
+        curl_setopt($ch, CURLOPT_COOKIEFILE, $cookies);
```
mickel1138: When I tried this only cookiejar is needed.

src/Pelagos/Util/UrlValidation.php
```diff
@@ -20,8 +20,12 @@ class UrlValidation
     public function validateUrl($url)
     {
         $ch = curl_init();
+        $cookies = tempnam('/tmp','cookie.txt');
+        curl_setopt($ch, CURLOPT_COOKIEJAR, $cookies);
```
mickel1138: Instead of giving it a cookie file, you can use '-' which means output to standard out. (Even if you don't use it.)

mickel1138: Or even better use: 'php://memory' as file name.

src/Pelagos/Util/UrlValidation.php
```diff
@@ -20,8 +20,12 @@ class UrlValidation
...
+        curl_setopt($ch, CURLOPT_COOKIEFILE, $cookies);
         curl_setopt($ch, CURLOPT_URL, $url);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
         $output = curl_exec($ch);
```
mickel1138: $output never used. Also look into CURLOPT_NOBODY which does a HEAD instead of GET.

src/Pelagos/Bundle/AppBundle/Command/ValidateRemotelyHostedLinksCommand.php
```diff
@@ -0,0 +1,124 @@
...
+                    $httpResponse = $urlValidationService->validateUrl($datasetSubmission->getDatasetFileUri());
+                    if ($httpResponse === true) {
+                        $httpCode = 200;
+                    } else {
+                        $httpCode = (trim(str_replace('Could not get URL, returned HTTP code', '', $httpResponse)));
+                        array_push($errorUdi, $dataset->getUdi());
+                    }
+                    $datasetSubmission->setDatasetFileUrlStatusCode($httpCode);
+                } catch (\Exception $e) {
+                    $output->writeln('Unable to process curl command, Error response: ' . $e->getMessage());
```
mickel1138: Should we add a return here? Or maybe add this message to the array?

src/Pelagos/Bundle/AppBundle/Command/ValidateRemotelyHostedLinksCommand.php
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+            $datasetSubmission->setDatasetFileUrlLastCheckedDate(new \DateTime('now', new \DateTimeZone('UTC')));
```
mickel1138: Should you do this inside the if statement, if $datasetSubmission is not a instance of DatasetSubmission, you couldn't even set it. You can however do this after the try-catch, to say that you checked it, even if curl fails.

src/Pelagos/Bundle/AppBundle/Resources/views/Email/data-repository-managers.error-remotely-hosted.email.twig
```diff
@@ -0,0 +1,14 @@
...
+
+{% block body_html %}
+    <h3>
+        Below are the list of Dataset Udi's which have failed.
+    </h3>
+
+    <br>
+
+    {% for udi in listOfUdi %}
+        <li>{{ udi }}</li>
```
mickel1138: Do we also want to give them a reason, or error code? Or maybe even the link?

src/Pelagos/Bundle/AppBundle/Command/ValidateRemotelyHostedLinksCommand.php
```diff
@@ -0,0 +1,124 @@
...
+            array(
+                'availabilityStatus' =>
+                    array(
+                        DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE_REMOTELY_HOSTED,
+                    )
+            )
+        );
+        $urlValidationService = new UrlValidation();
+        $errorUdi = array();
+
```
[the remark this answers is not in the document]
praneethpr: 👍

src/Pelagos/Util/UrlValidation.php
praneethpr: Probably we can change it to optional, when there is a use case of not following redirects?

src/Pelagos/Util/UrlValidation.php
praneethpr: Done. Changed it

src/Pelagos/Util/UrlValidation.php
praneethpr: Used php:://memory. 👍

src/Pelagos/Bundle/AppBundle/Command/ValidateRemotelyHostedLinksCommand.php
praneethpr: Added a continue so that it doesn't set for the current object and proceeds to the next iteration

src/Pelagos/Bundle/AppBundle/Command/ValidateRemotelyHostedLinksCommand.php
praneethpr: Yeah, moved it inside the if loop. Good catch :)

src/Pelagos/Bundle/AppBundle/Resources/views/Email/data-repository-managers.error-remotely-hosted.email.twig
praneethpr: We can push it to the PO, and if they want changes to the email sent, we can add in more stuff

src/Pelagos/Util/UrlValidation.php
mickel1138: Can't think of any really.

src/Pelagos/Util/UrlValidation.php
praneethpr: Changed it.

src/Pelagos/Bundle/AppBundle/Command/ValidateRemotelyHostedLinksCommand.php
praneethpr: Removed try-catch block

PR #154, opened 2019-02-05
Added creator to be tba when authors is null in the Doi Consumer

praneethpr:


src/Pelagos/Bundle/AppBundle/Rabbit/Consumer/DoiConsumer.php
```diff
@@ -229,10 +229,11 @@ protected function updateDoi(Dataset $dataset, array $loggingContext)
                 $doiUrl = 'https://data.gulfresearchinitiative.org/tombstone/' . $dataset->getUdi();
             }
 
+            $creator = ($dataset->getAuthors()) ? $dataset->getAuthors() : 'tba';
```
mickel1138: The code is '(:tba)'

praneethpr: Changed it

PR #155, opened 2019-02-06
Feature/pelagos 3471 add cold storage attributes to dataset review form

snguyen1:


praneethpr:
> I wish github would let me delete my review and start fresh.

You can hit the dismiss review

src/Pelagos/Bundle/AppBundle/Resources/views/DatasetReview/distribution-info.html.twig
```diff
@@ -52,6 +52,28 @@
...
+    <table class="tab-table">
+        <tr>
+            <td>
+                {{ form_label(form.isDatasetFileInColdStorage) }}
+                {{ form_widget(form.isDatasetFileInColdStorage, { 'attr': {'class': 'coldstorage-checkbox'} }) }}
+            </td>
+        </tr>
+        <tr class="row-coldstorage-filesize">
+            <td>
+                {{ form_row(form.datasetFileColdStorageArchiveSize) }}
```
praneethpr: Should have units beside it? Easier to understand. They wanted it in bytes, so having scroll wouldn't be that helpful i guess.

src/Pelagos/Bundle/AppBundle/Resources/public/js/datasetReview.js
```diff
@@ -656,8 +656,28 @@ $(document).ready(function(){
...
+function checkColdStorageCheckBoxState() {
+    if ($(".coldstorage-checkbox").is(":checked")) {
+        $(".row-coldstorage-filesize").show();
+        $(".row-coldstorage-sha256hash").show();
+        $("#datasetFileColdStorageArchiveSize").attr("required", "required");
+        $("#datasetFileColdStorageArchiveSha256Hash").attr("required", "required");
+    } else {
+        $(".row-coldstorage-filesize").hide();
+        $(".row-coldstorage-sha256hash").hide();
+        $("#datasetFileColdStorageArchiveSize").removeAttr("required");
```
praneethpr: Clear out the value (val("")), or else if you had marked it previously and unmarking doesn't clear the field, It would still persist the value when you submit the form

src/Pelagos/Bundle/AppBundle/Resources/public/js/datasetReview.js
```diff
@@ -656,8 +656,28 @@ $(document).ready(function(){
...
+    $(".coldstorage-checkbox").change(function (){
+        checkColdStorageCheckBoxState();
+    });
 });
 
+function checkColdStorageCheckBoxState() {
+    if ($(".coldstorage-checkbox").is(":checked")) {
+        $(".row-coldstorage-filesize").show();
+        $(".row-coldstorage-sha256hash").show();
+        $("#datasetFileColdStorageArchiveSize").attr("required", "required");
```
praneethpr: I think it's better to clear out the values here too.

src/Pelagos/Bundle/AppBundle/Resources/views/DatasetReview/distribution-info.html.twig
snguyen1: I put it in but somehow deleted it. See the fix in the DatasetSubmissionType.php (form)

src/Pelagos/Bundle/AppBundle/Resources/public/js/datasetReview.js
snguyen1: There is a use case where user loads the review that has existing values, if they accidentally toggle the switch, the values would be cleared (user may panic because they can't undo it).
If they leave the value, the form will know how to clear the value using on the PRE_SUBMIT event.

src/Pelagos/Bundle/AppBundle/Resources/views/DatasetReview/distribution-info.html.twig
praneethpr: 👍

src/Pelagos/Bundle/AppBundle/Resources/public/js/datasetReview.js
snguyen1: Update: I add a dialog box to prevent user from disabling the fields and empty the values by accident

praneethpr: 👍

src/Pelagos/Bundle/AppBundle/Form/DatasetSubmissionType.php
```diff
@@ -283,6 +299,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'label' => 'Request Revisions',
                 'attr'  => array('class' => 'submitButton'),
             ));
+
+        $builder->addEventListener(FormEvents::POST_SET_DATA, function (FormEvent $event) {
+            $data = $event->getData();
+            $form = $event->getForm();
+            if ($data->isDatasetFileInColdStorage() === true) {
```
praneethpr: I think this function is an overkill? You have client side and server side validation in place

snguyen1: Yes, It's kinda overkill but this way works fine and it also guarantees that everything is right from the backend IMO

PR #159, opened 2019-02-11
Feature/pelagos 3479 switch to symfony encore

mickel1138:


package.json
```diff
@@ -0,0 +1,11 @@
+{
```
praneethpr: Do you also want to update the bin/deploy script to run the yarn install command

mickel1138: yes, that's a good idea. And it would also need to run encore. But maybe we should wait until it'll done? So we optimize it.

src/Pelagos/Bundle/AppBundle/Resources/views/template/UI.html.twig
```diff
@@ -1,6 +1,10 @@
 {% extends "PelagosAppBundle::layout.html.twig" %}
 
 {% block head %}
+    
+    <link rel="stylesheet" href="{{ asset('/build/layout.css') }}">
+    <script src="{{ asset('/build/layout.js') }}"></script>
+
     {{
         add_library (
```
praneethpr: Instead of making changes to Twig extensions, shouldn't we just go ahead and remove this?

mickel1138: Yes, eventually we will do that. But since I consider this pull request just the "framework", we can make all those changes in each "application" template. In this case I could have deleted this section, because I already modified the file, and this function now does nothing. I'm going to change this file.

mickel1138:
To make this work follow these steps:
composer install
yarn install
yarn run encore dev
bin/console server:run {port#}

Then get to the listening server with the usual route like: "/pelagos-symfony/dev/mvde"

PR #163, opened 2019-02-13
feature/PELAGOS-3473a-creating-drupal-header-templates-for-pelagos 

snguyen1:


assets/css/template.css
```diff
@@ -0,0 +1,537 @@
...
+    border-image: initial;
+    background-color: rgb(63, 98, 106);
+    white-space: nowrap;
+    text-decoration: none !important;
+    -webkit-box-shadow: 0 0 5px rgba(0, 0, 0, 0.5);
+    box-shadow: 0 0 5px rgba(0, 0, 0, 0.5);
+    border-radius: 0 0 5px 5px;
+    top: 0px;
+    position: relative !important;
+}
```
mickel1138: need new-line

assets/css/template.css
```diff
@@ -0,0 +1,537 @@
...
+
+.atlassian-btns li a {
+    color: white !important;
+    font-size: 9pt;
+    font-family: sans-serif;
+    font-weight: normal;
+    padding: 0 5px 0 5px;
+    border-width: 0px 1px 1px;
+    border-top-style: initial;
```
mickel1138: If all these are set the same, you can use border for all the sides (left, right, bottom etc...)

src/Pelagos/Bundle/AppBundle/Resources/views/layout-head.html.twig
```diff
@@ -0,0 +1,41 @@
+{% block head %}
+<head>
```
mickel1138: I think the head tag should be outside the block, because if you want to overwrite the head block, the tags should still be there.

PR #164, opened 2019-02-13
Removed drupal functions to include metadata in the head tag

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/index.html.twig
```diff
@@ -42,6 +39,13 @@
 
     {{ add_js(path('fos_js_routing_js', { callback: 'fos.Router.setData' })) }}
 
+    {% if dataset.availabilityStatus == 10 %}
```
mickel1138: Rather than using 10, use the contant like: constant('Pelagos\\Entity\\DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE')

src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/index.html.twig
```diff
@@ -42,6 +39,13 @@
...
+        {% include 'PelagosAppBundle:Dataland:metadata.markup.twig' %}
+    {% else %}
+        <meta name="robots" content="noindex" />
+    {% endif %}
+
+
```
mickel1138: extra line

src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/index.html.twig
mickel1138: Maybe put this in the place where {{parent()}} is, and put parent right under it, that way the META data is the first thing in the head. Technically doesn't matter. What do you think?

src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/index.html.twig
```diff
@@ -6,9 +6,6 @@
     {{
```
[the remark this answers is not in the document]
praneethpr: Are we going to use only Comment and Approve changes?  No more Request Changes ? :D

src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/index.html.twig
praneethpr: 👍

src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/index.html.twig
praneethpr: Removed it

src/Pelagos/Bundle/AppBundle/Resources/views/Dataland/index.html.twig
praneethpr: I checked how it works now, it looks like its placed above, so i moved it 👍

PR #165, opened 2019-02-15
Added custom jquery method to validate spatial extent on dataset subm…

praneethpr:
…ission

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Resources/public/js/dataset-submission.js
+++ b/src/Pelagos/Bundle/AppBundle/Resources/public/js/dataset-submission.js
@@ -137,15 +137,34 @@ $(function() {
         return this.optional(element) || ((Date.parse(value)) && regPattern.test(value));
     });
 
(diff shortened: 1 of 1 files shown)


praneethpr:
To test:
You can use the empty gml XML file from the bug (DMS-543) to import and then try to submit.

praneethpr:
> After this morning's discussion with the PO, are we not doing this?

I think we need to discuss about it

src/Pelagos/Bundle/AppBundle/Resources/public/js/dataset-submission.js
```diff
@@ -137,15 +137,34 @@ $(function() {
...
+        var isSuccess = true;
+        if ($("#spatialExtentDescription").val() === "") {
+            isSuccess = false;
+            $.ajax({
+                url: Routing.generate("pelagos_app_gml_towkt"),
+                type: "POST",
+                async: false,
+                data: {gml: value},
+                success: function () {
+                    isSuccess = "true";
```
snguyen1: text or boolean?

PR #166, opened 2019-02-15
bugfix/pelagos 3495 write command to handle submission with bad gml

snguyen1:
Command:  bin/console onetime:handle-bad-gml-submission R4.x268.000:0046
To test this:
1) bin/loaddump ...
2) On dev only: run sql command: **UPDATE dataset_submission SET dataset_file_uri = ‘file:///var/tmp/hello.txt’ where dataset_submission.id = 12034** (to point to a file on dev)
3) rebuild rabbit  (?)
3) run command: bin/console onetime:handle-bad-gml-submission R4.x268.000:0046
4) View R4.x268.000:0046 on dataset review

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
+++ b/src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
@@ -6,12 +6,7 @@ use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-
-use Pelagos\Entity\DataCenter;
-use Pelagos\Entity\Dataset;
 use Pelagos\Entity\DatasetSubmission;
-use Pelagos\Entity\DistributionPoint;
-use Pelagos\Entity\Person;
 
 /**
  * Back fill all the submitted metadata xml to dataset submission.
--- a/src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
+++ b/src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
@@ -77,7 +77,7 @@ class HandleSubmissionWithBadGMLCommand extends ContainerAwareCommand
         $entityManager->flush();
 
         $output->writeln('Success: submission ID:' . $datasetSubmission->getId() . ' - Dataset udi: ' . $udi);
-        
+
         return 0;
     }
 }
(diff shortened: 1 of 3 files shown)


src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
```diff
@@ -0,0 +1,83 @@
...
+        $dataset = $datasets[0];
+
+        $datasetSubmission = $dataset->getDatasetSubmission();
+        if (!($datasetSubmission instanceof DatasetSubmission)) {
+            throw new \Exception('Could not find Dataset Submission.');
+        }
+
+        $datasetSubmission->setSpatialExtent(null);
+
+        $datasetSubmission->setDatasetFileTransferStatus(DatasetSubmission::TRANSFER_STATUS_NONE);
```
praneethpr: If you set this to NONE, I dont think we can open it in review tool

src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
```diff
@@ -0,0 +1,84 @@
...
```
praneethpr: I don't think this is necessary, I guess it is already in this status

src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
```diff
@@ -0,0 +1,84 @@
...
+
+        //clear distribution points if there is one that has no Data Center
+        if (count($datasetSubmission->getDistributionPoints()) === 1 ) {
```
praneethpr: Can use clear() method to remove all elements in the collection
$datasetSubmission->getDistributionPoints()->clear()

src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
```diff
@@ -0,0 +1,84 @@
...
+
+        //re-trigger dataset submission producer
+        $this->getContainer()->get('old_sound_rabbit_mq.dataset_submission_producer')->publish(
+            $datasetSubmission->getId(),
+            'dataset.' . $datasetSubmission->getDatasetFileTransferType()
+        );
+
+        $output->writeln('Success: submission ID:' . $datasetSubmission->getId() . ' - Dataset udi: ' . $udi);
+
+        return 0;
```
praneethpr: Can remove this, Commands don't have to return anything. I know we've written few in the past like this.

src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
snguyen1: I have to clear it manually because if we need to do this again for a remotelyhosted dataset, this would not clear the data. But if we only use it once, either way works

src/Pelagos/Bundle/AppBundle/Command/HandleSubmissionWithBadGMLCommand.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Script to nullify spatial extents and remove distribution point tied to the dataset submission of a dataset
+     *
+     * @param InputInterface  $input  An InputInterface instance.
+     * @param OutputInterface $output An OutputInterface instance.
+     *
+     * @throws \Exception When dataset not found.
+     * @throws \Exception When datasetSubmission not found.
+     *
+     * @return integer Return 0 on success, or an error code otherwise.
```
praneethpr: Do you wanna change the annotation to void?

PR #169, opened 2019-02-18
Feature/pelagos 3492 bulk coldstorage marking cmd

fightingtexasaggie:
See testing procedure at: https://griidc.atlassian.net/browse/PELAGOS-3492


src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,181 @@
...
+     * @throws \ReflectionException If reflection class fails.
+     * @throws \Exception If stubfile not readable/accessible.
+     * @throws \Exception If infofile not readable/accessible.
+     *
+     * @return integer
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $entityManager = $this->getContainer()->get('doctrine.orm.entity_manager');
+        $entityEventDispatcher = $this->getContainer()->get('pelagos.event.entity_event_dispatcher');
```
snguyen1: $entityEventDispatcher was never used.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,181 @@
+<?php
+
+namespace Pelagos\Bundle\AppBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Pelagos\Entity\Dataset;
```
snguyen1: Dataset is never used.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,181 @@
...
+ */
+class ColdStorageFlagCommand extends ContainerAwareCommand
+{
+
+    /**
+     * Doctrine EntityManager, as provided in container.
+     *
+     * @var mixed entityManager
+     */
+    protected $entityManager;
```
praneethpr: Not used, can be removed

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,181 @@
...
+                        'dataset.SFTP'
+                    );
+                }
+            } else {
+                throw new \Exception("Could not open $stubFileName, expected to be at same location as $infoFileName.");
+            }
+        } else {
+            throw new \Exception("Error: Could not open $infoFileName.");
+        }
+        return 0;
```
praneethpr: Can remove this, Commands don't have to return anything. I know we've written few in the past like this.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,181 @@
...
+                    // Create a new submission using latest via constructor.
+                    $newDatasetSubmission = new DatasetSubmission($datasetSubmission);
+                    if ($entityManager->contains($newDatasetSubmission)) {
+                        throw new \Exception('Attempted to create a dataset submission that is already tracked');
+                    }
+                    $output->writeln('original submission ID is: ' . $dataset->getDatasetSubmission()->getId());
+                    $output->writeln('new submission ID is initially: ' . $newDatasetSubmission->getId());
+
+                    // This method sets submission status to: DatasetSubmission::STATUS_IN_REVIEW (3).
+                    $newDatasetSubmission->setDatasetSubmissionReviewStatus();
```
praneethpr: I don't think this should be done as part of this command, if you are setting this, then we should create a DatasetSubmissionReview object as well.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,181 @@
...
+
+                    if ($newDatasetSubmissionId !== null) {
+                        $metadata->setIdGenerator($newDatasetSubmissionIdGenerator);
+                    }
+
+                    $dataset->setDatasetSubmission($newDatasetSubmission);
+                    $entityManager->persist($dataset);
+                    $entityManager->flush($dataset);
+
+                    // This is an ugly hack, having to re-persist, but there seems not to be a cleaner way...
```
praneethpr: There is no setter for status in DatasetSubmission Entity, so i think we should probably not change it using Reflection class. I think lot of other attributes are tied to it, like submissionTimeStamp and submitter which also need to be set or else we would have lot of discrepancies between these datasets and others.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,179 @@
...
+                    // Reset original status.
+                    try {
+                        $newDatasetSubmissionReflection = new \ReflectionClass($newDatasetSubmission);
+                        $statusReflection = $newDatasetSubmissionReflection->getProperty('status');
+                        $statusReflection->setACcessible(true);
```
snguyen1: A**c**cessible?

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
fightingtexasaggie: This is a mistake I meant to remove earlier...good catch.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
fightingtexasaggie: I'm using reflections here just for changing it back to the value it was originally in the original dataset submission. The issue is that I cannot call dataset->setDatasetSubmission() on the new submission if the value is something other than DatasetSubmission :STATUS_COMPLETE or DatasetSubmission::STATUS_IN_REVIEW otherwise it'll be specifically nulled in the Dataset->setDatasetSubmission().

The constructor in DatasetSubmission.php handles submissionTimeStamp and submitter. modificationTimeStamp and modifier will get stomped, but everything that touches the entity stomps these anyway.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
fightingtexasaggie: Luckily, method names are case insensitive.  Will fix as it looks silly.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
fightingtexasaggie: removed.

src/Pelagos/Bundle/AppBundle/Command/ColdStorageFlagCommand.php
fightingtexasaggie: I could have also used reflections to bypass dataset's setter. This might be cleaner.

fightingtexasaggie:
When testing, make sure and do a composer install. I just merged in a fresh develop branch. If you forget, you may see the error:
 _Content-Type header [application/x-www-form-urlencoded] is not supported_

PR #170, opened 2019-02-18
Added symfony command to set primary point of contact for datasets

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/SetPrimaryPointOfContactCommand.php
```diff
@@ -0,0 +1,76 @@
...
+            $udi = $dataset->getUdi();
+            $output->writeln('Checking primary contact for udi: ' . $udi);
+
+            if ($datasetSubmission instanceof DatasetSubmission) {
+                $datasetContacts = $datasetSubmission->getDatasetContacts();
+                $primaryContact = array();
+
+                foreach ($datasetContacts as $datasetContact) {
+                    if ($datasetContact->isPrimaryContact()) {
+                        continue 2;
```
mickel1138: What's this do?

praneethpr: if primary contact is already set for a dataset, it exits the current iteration and moves forward with the next iteration

src/Pelagos/Bundle/AppBundle/Command/SetPrimaryPointOfContactCommand.php
```diff
@@ -0,0 +1,76 @@
...
+     * @param OutputInterface $output An OutputInterface instance.
+     *
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $entityManager = $this->getContainer()->get('doctrine.orm.entity_manager');
+
+        $datasets = $entityManager->getRepository(Dataset::class)
+            ->findBy(array('datasetStatus' => Dataset::DATASET_STATUS_ACCEPTED));
```
snguyen1: only the Accepted??? 
What about InReview?

praneethpr: This data discrepancy happened during the release of 5.0.0, so all the datasets which were in accepted status got migrated without a primary contact. And checking for InReview status and setting an attribute without creating a new dataset submission object might bring new complications, and there should be only 2-3 datasets in that InReview status which don't have a primary contact.

PR #172, opened 2019-02-20
feature/PELAGOS-3502-toggle-show-differences-in-side-by-side-tool 

snguyen1:


src/Pelagos/Bundle/AppBundle/Resources/public/js/sidebyside.js
```diff
@@ -112,7 +112,14 @@ $(document).ready(function()
     });
 
     $("#show-diff-button").button("option", "disabled", true).click(function (){
-        showDifferences();
+        if ($(this).hasClass(".show-diffs")) { //hide
```
mickel1138: "." not needed.

PR #173, opened 2019-02-21
feature/PELAGOS-3504-change-erdap-button

snguyen1:
Remember to add erddap link to the dataset submission before opening in dataland


mickel1138:
Looks fine, but still think CSS would have been easier.

PR #174, opened 2019-02-21
Feature/pelagos 3501 fix templates to use ui template if they are ui

mickel1138:
This fixed all the templates, and add_library is removed. But it's still using add_css and add_js, but we can replace those too if we like, or wait.

Also not every template uses the new blocks, should we do that now also?


src/Pelagos/Bundle/AppBundle/Resources/views/MdApp/upload-complete.html.twig
```diff
@@ -1,17 +1,8 @@
-{% extends "PelagosAppBundle::layout.html.twig" %}
+{% extends "PelagosAppBundle:template:UI.html.twig" %}
 
 {% block head %}
     {{ parent() }}
```
fightingtexasaggie: I don't think this view is even used anymore.

PR #178, opened 2019-02-25
Remove unused bundles

mickel1138:


fightingtexasaggie:
I'm currently getting a CSRF error on the login screen since the token is still being set on the login, and we've disabled CSRF.

PR #180, opened 2019-02-26
Added try catch block to handle Invalid gml exception

praneethpr:


src/Pelagos/Event/DatasetIndexSubscriber.php
```diff
@@ -52,9 +54,14 @@ public function populateCalculatedFields(TransformEvent $event)
...
-            $dataset->getSpatialExtentGeometry()
-        );
+        try {
+            $wkt = $this->geometryUtil->convertGmlToWkt(
+                $dataset->getSpatialExtentGeometry()
+            );
+        } catch (InvalidGmlException $exception) {
+            $wkt = null;
+        }
+
```
fightingtexasaggie: This change worked perfectly in testing.

PR #182, opened 2019-02-28
feature/pelagos 3515 fix css on dataset discovery and on resize pelagos

snguyen1:
You may want to merge it with data-discovery fix branch to see how it works

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Resources/views/layout.html.twig
+++ b/src/Pelagos/Bundle/AppBundle/Resources/views/layout.html.twig
@@ -95,20 +95,51 @@
                 </div>
             </div>
             {% include "PelagosAppBundle::layout-menu.html.twig" %}
-            <div id="pelagos-container">
-                <div id="pelagos-container-row">
-                    <div id="pelagos-container-cell">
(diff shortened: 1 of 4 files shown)


src/Pelagos/Bundle/AppBundle/Resources/views/layout.html.twig
```diff
@@ -95,20 +95,51 @@
...
-                                <div id="pelagos-content">
-                                    <h1 style="margin: 0 auto;">{% block pagetitle %}{% endblock %}</h1>
-                                    {% block body %}{% endblock %}
+            <div id="main-wrapper"><div id="main">
+
+                    <div id="content-wrapper">
+                        <div id="content-main" class="column">
+                            <div class="section">
+                                <div id="outer-container">
+                                    <div>
```
mickel1138: What's this stray div doing here?

snguyen1: Just wanted to show you the original drupal's div. we only need one extra div

mickel1138: It works tho

PR #185, opened 2019-03-01
Feature/pelagos 3518 add filename to entity

fightingtexasaggie:
using Pelagos-3517 as a mini-epic


src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -3,7 +3,7 @@
 namespace Pelagos\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
-
+use Pelagos\Util\PelagosUtilities;
```
praneethpr: Do you want to move this use statement below, for better grouping?

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -1129,16 +1139,13 @@ public function validate(ExecutionContextInterface $context)
...
-        if (null !== $this->datasetFileColdStorageArchiveSize && null === $this->datasetFileColdStorageArchiveSha256Hash) {
-                $context->buildViolation($coldStorageViolationMsg)
-                    ->atPath('datasetFileColdStorageArchiveSha256Hash')
-                    ->addViolation();
-        }
-
-        if (null === $this->datasetFileColdStorageArchiveSize && null !== $this->datasetFileColdStorageArchiveSha256Hash) {
+        $coldStorageViolationMsg = 'You must provide file size, sha256 hash, and original '
+        . 'filename for Cold Storage Information.';
+        if (false === (PelagosUtilities::nullOrNone(array($this->this->datasetFileColdStorageArchiveSize,
```
praneethpr: $this->this , Extra this keyword

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -1129,16 +1139,13 @@ public function validate(ExecutionContextInterface $context)
...
+            $this->datasetFileColdStorageArchiveSha256Hash,
+            $this->datasetFileColdStorageOriginalFilename)))) {
```
praneethpr: How would it get set in the first place? I'm not sure whether I totally understand

src/Pelagos/Entity/DatasetSubmission.php
fightingtexasaggie: yeah, we normally alphabetize them.

src/Pelagos/Entity/DatasetSubmission.php
fightingtexasaggie: It would be set normally from either the review form, or via the bulk-loader command. This just enforced the "all 3 or none at all" rule for the form.

src/Pelagos/Entity/DatasetSubmission.php
mickel1138: This is a lot of code that could be done with a simple if statement. Also, a whole new class and utility for a function that will likely never get used again? 

`$test1 = array(null, 1, null);
$test2 = array(null, null, null);
$test3 = array(1, 1, 1);

var_dump(!array_filter($test1) or !in_array(null,$test1));
var_dump(!array_filter($test2) or !in_array(null,$test2));
var_dump(!array_filter($test3) or !in_array(null,$test3));`

src/Pelagos/Entity/DatasetSubmission.php
fightingtexasaggie: Removed this as it is no longer needed with a public method that sets all 3 attributes and once and does not allow nulls, in place of the 3 separate public setters.

app/DoctrineMigrations/Version20190301182150.php
```diff
@@ -0,0 +1,30 @@
+<?php declare(strict_types=1);
```
praneethpr: Are we going to declare this for all new php files? Now, that we have moved to PHP7

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -1764,14 +1760,55 @@ public function getDatasetFileColdStorageArchiveSize()
...
+    /**
+     * Set the cold storage attributes, as one action, no nulls allowed.
+     *
+     * @param integer $filesize The original file size, in bytes, to be preserved.
+     * @param string  $hash     The original file sha256 hash to be preserved.
+     * @param string  $filename The original file name to be preserved.
+     *
+     * @return void
+     */
+    public function setDatasetFileColdStorageAttributes(int $filesize, string $hash, string $filename)
```
praneethpr: I'm not sure whether this is a good practice of grouping setters, we haven't done it in the past (for ex: spatial extent)

app/DoctrineMigrations/Version20190301182150.php
fightingtexasaggie: Personally I think it would be a good direction, especially for all new code.

PR #186, opened 2019-03-05
Feature/pelagos 3506 add additional admin menu when role role data repository manager

mickel1138:


src/Pelagos/Bundle/AppBundle/Resources/views/Menus/main.html.twig
```diff
@@ -47,10 +46,13 @@
                             <li class="last odd "><a href="https://data.gulfresearchinitiative.org/RB-data-compliance">Data Compliance Information</a></li>
                         </ul>
                     </li>
-                    <li class="last odd "><a href="http://research.gulfresearchinitiative.org/" title="GoMRI Research Information" name="GoMRI Research" target="_top" >Research</a></li>
+                    {% if is_granted('ROLE_DATA_REPOSITORY_MANAGER') %}
+                        {% include "PelagosAppBundle:Menus:admin.html.twig" %}
+                    {% else %}
+                        <li class="last odd"><a href="http://research.gulfresearchinitiative.org/" title="GoMRI Research Information" name="GoMRI Research" target="_top" >Research</a></li>
```
praneethpr: I think this should be moved out the if-else, so that everyone can see this on the menu

praneethpr:
Note: We also need to fix the css for "admin only" error message template

PR #189, opened 2019-03-05
feature/PELAGOS-3520-add-original-filename-handling-to-dataset-flag-coldstorage-pelagos-command

fightingtexasaggie:


fightingtexasaggie:
Make sure 3518 is merged in before testing.

bin/console dataset:flag-coldstorage /home/users/mwilliamson/COLD_STORAGE/manifests/FL.x704.000.0002-coldinfo.txt

examine result in side-by-side tool.

PR #191, opened 2019-03-06
Feature/pelagos 3477 rewrite data discovery

mickel1138:
I think this is working like before.


praneethpr:
> So far it looks pretty identical, but I have noticed that several times a selection from the tree results in a blank listing in the datasets section until I click it a second time. It's fairly easy to reproduce with a few tries, but not consistent.

I don't think it is related to this pull request, I'm able to reproduce this on Griidc-test as well. Its really a good find Mike W. 👍

src/Pelagos/Bundle/AppBundle/Resources/public/js/tree.js
```diff
@@ -167,11 +160,12 @@ function updateTree(tree) {
     });
 
     $("#" + tree.name).bind("loaded.jstree", function(event, data) {
+        console.log('loaded');
         if (typeof tree.onload !== 'undefined') {
```
praneethpr: Works when any either one of the if tree.onload block is removed. But I am not sure that I understand the whole function

mickel1138: Removed the second eval for now, this was broken in Production, so I don't know what it should do.

PR #194, opened 2019-03-07
Feature/pelagos 3527 conditionally add erddap info in md generator

fightingtexasaggie:
Awaiting William's codeListValue usage decision - don't merge until we know whether to use the numeric value or the text twice.

I did a lot of very loosly-related GCS-related cleanup, but if y'all would rather this be in a separate re-factor branch I can remove them as I did it in separate commits.


src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -2666,17 +2666,32 @@ public function getDistributionPoints()
...
 
+    /**
+     * Getter for the erddap url protocol.
+     *
+     * @return string|null
+     */
+    public function getErddapUrlProtocol()
+    {
+        if ($this->erddapUrl !== null) {
+            preg_match('/^((.*?):).*$/', $this->erddapUrl, $matches);
```
praneethpr: I think these changes are overwriting the refactor branch changes

PR #197, opened 2019-03-08
update-md-generator-with-new-elements

fightingtexasaggie:


fightingtexasaggie:
It's hardcoded because this is a fixed resource. After discussion with William, I added the codespace artifact as well. There's really nothing to calculate or pull from, this just goes along with the "data landing page" section.

PR #198, opened 2019-03-08
Parse distributionUrl protocol instead of hardcode.

fightingtexasaggie:
In another ticket it was discovered we were hardcoding 'https' instead of parsing from the distributionUrl. PO suggested it would be better to parse instead of hardcode.

### DIFF
--- a/src/Pelagos/Entity/DistributionPoint.php
+++ b/src/Pelagos/Entity/DistributionPoint.php
@@ -34,12 +34,12 @@ class DistributionPoint extends Entity
         ],
         'custodian' => [
             'name' => 'Custodian',
-            'description' => 'The individual/organization that has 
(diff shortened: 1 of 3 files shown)


src/Pelagos/Entity/DistributionPoint.php
```diff
@@ -183,6 +183,21 @@ public function getDistributionUrl()
...
 
+    /**
+     * Getter for distribution url protocol.
+     *
+     * @return string
+     */
+    public function getDistributionUrlProtocol()
+    {
+        if ($this->distributionUrl !== null) {
+            preg_match('/^((.*?):).*$/', $this->distributionUrl, $matches);
```
praneethpr: is the 2nd capturing group required?

PR #199, opened 2019-03-11
Feature/pelagos 3523 basic search form and button

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/Search/results.html.twig
```diff
@@ -0,0 +1,60 @@
...
+                    {% for result in results %}
+                        <div class="hr-line-dashed"></div>
+                        <div class="search-result">
+                            <h3><a class="search-results"
+                                   href="{{ path('pelagos_app_ui_dataland_default', {'udi' : result.udi } ) }}">
+                                    {{ result.datasetSubmission.authors | default }}
+                                    {{ result.year |default }}
+                                    {{ result.title }}
+                                    {{ result.researchGroup.name }}.
+                                    DOI: {{ result.doi.doi }}
```
fightingtexasaggie: The doi might possibly be null, so a |default could help.

praneethpr:
> You may want to look at the util/DatasetIndex class starting at line 80. This is how we handled the search by UDI and search by DOI, otherwise any UDI returns a ton of unrelated datasets. By default the UDI is split into 4 parts and each is search for separately. We regexp matched an UDI and a DOI, and made those exact-phrase only.

Udi search wasn't a part of this ticket, i just added that for testing. Removed it

PR #202, opened 2019-03-14
Set publication year to dif approved date if no reference date exists

praneethpr:


src/Pelagos/Bundle/AppBundle/Rabbit/Consumer/DoiConsumer.php
```diff
@@ -251,13 +251,20 @@ protected function updateDoi(Dataset $dataset, array $loggingContext)
             }
 
             $creator = ($dataset->getAuthors()) ? $dataset->getAuthors() : '(:tba)';
+
+            // PublicationYear field can not be null, as it is a required field when the DOI is published
+            $pubYear = $dataset->getReferenceDateYear();
+            if (empty($pubYear) and
+                $dataset->getDif()->getApprovedDate() instanceof \Datetime) {
```
fightingtexasaggie: In the event approvedDate is null, what are we going to do? Would it be better to throw an exception and log something than attempt to go ahead with an empty pub year?

praneethpr: When the updateDoi is called in the consumer, according to our workflow, DatasetSubmission object for a dataset should be submitted and which would have a referenceDateYear field not null. So using the DIF approved Date is a fallback if that fails and we create a Doi only when a DIF is approved. I think throwing an exception/logging is an overkill.

fightingtexasaggie: :+1:

PR #203, opened 2019-03-18
Feature/pelagos 3538 fix css and search results inclusive

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/Search/default.html.twig
```diff
@@ -3,19 +3,44 @@
 {% block head %}
     {{ add_css (
```
mickel1138: By overwriting the head, you basically remove all the head stuff from the template.

Just use stylesheet and javascipts blocks to add stuff, prefix each block with {{ parent() }}

mickel1138: I forgot, this is not the new framework yet.

src/Pelagos/Bundle/AppBundle/Resources/views/Search/default.html.twig
```diff
@@ -3,19 +3,44 @@
...
         ]
     ) }}
     {% stylesheets
         '@PelagosAppBundle/Resources/public/css/search.css' %}
         {{ add_css(asset_url) }}
     {% endstylesheets %}
+
+    {{ add_js (
+        [
+            '//code.jquery.com/jquery-2.1.3.min.js',
```
mickel1138: You shouldn't have to include jQuery right?

praneethpr: The pagination plugin requires at least that version of jQuery, I assumed that we have an older version of jQuery

mickel1138: I removed this line, and it seems to still work...

praneethpr: Yes, i tried it too. Will remove it 👍

mickel1138:
On subsequent searches, the paginator links do not get updated.

mickel1138:
also the search bar doesn't update, so I think that's were the issue is.

praneethpr:
> On subsequent searches, the paginator links do not get updated.

Good find. 👍

PR #205, opened 2019-03-20
Feature/pelagos 3505 add login attempts

mickel1138:
For easy testing, change lockoutTimeSeconds and maxAttempts to 60 and 10.


app/DoctrineMigrations/Version20190319155515.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->addSql('ALTER TABLE account ALTER user_id TYPE citext');
+        $this->addSql('ALTER TABLE account ALTER user_id DROP DEFAULT');
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
```
fightingtexasaggie: We usually remove this auto-generated line from down migrations don't we?

fightingtexasaggie: here too

mickel1138: It's been so long since i've done one. I can remove it?

praneethpr: Yes, i think you should remove it

src/Pelagos/Bundle/AppBundle/Resources/views/Security/login.html.twig
```diff
@@ -22,14 +22,14 @@
 
     <script type="text/javascript">
         $(document).ready(function(){
-            $(".pelagosNoty").pelagosNoty({timeout: 6000, showOnTop:false});
+            $(".pelagosNoty").pelagosNoty({timeout: 10000, showOnTop:false});
```
praneethpr: I think we should use something other than noty, so that the error message stays, even when you click on it.

src/Pelagos/Entity/Account.php
```diff
@@ -317,6 +329,71 @@ public function getPasswordEntity()
...
+     * Whether or not this account is locked out.
+     *
+     * @return boolean
+     */
+    public function isLockedOut()
+    {
+        $lockoutTimeSeconds = 600;
+        $maxAttempts = 100;
+
+        $tooManyAttemps = false;
```
praneethpr: Typo "Attempts"

src/Pelagos/Entity/Account.php
```diff
@@ -317,6 +329,71 @@ public function getPasswordEntity()
...
+        $timeHasPasssed = false;
```
praneethpr: Typo "Passed"

src/Pelagos/Entity/Account.php
```diff
@@ -192,7 +192,7 @@ public function __construct(Person $person = null, $userId = null, Password $pas
      */
     public function getId()
     {
-        return $this->getPerson();
+        return $this->getPerson()->getId();
```
praneethpr: We should check if the object exists before we call a method on it

src/Pelagos/Entity/Account.php
```diff
@@ -317,6 +329,71 @@ public function getPasswordEntity()
         return $this->password;
     }
 
+    /**
+     * Returns the LoginAttempts Collection for this Account.
+     *
+     * @return LoginAttempts The LoginAttempts entity attached to this Account.
```
praneethpr: return value is Collection of loginAttempts, we usually specify it as Collection

src/Pelagos/Bundle/AppBundle/Resources/views/Security/login.html.twig
mickel1138: Bootstrap Alert? :-D

src/Pelagos/Entity/Account.php
mickel1138: That's right,  but now I don't remember why I put getId() here...

mickel1138:
> I'm unable to login with my credentials. Let me know if there is something that i'm missing to do before i test

@praneethpr the password is the actual password in the database, not what's in LDAP. So it's the same one as production.

src/Pelagos/Bundle/AppBundle/Resources/views/Security/login.html.twig
fightingtexasaggie: One of the advantages of the Noty is we don't get into back-button navigation user confusion.

src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
```diff
@@ -114,12 +144,26 @@ public function getUser($credentials, UserProviderInterface $userProvider)
...
+     * @throws AuthenticationException When account is locked out.
+     * @throws AuthenticationException When this is a bad password.
+     *
      * @return bool True if the credentials are valid.
      */
     public function checkCredentials($credentials, UserInterface $user)
     {
+        // Here check to see if $user is locked out?
+        if ($user->isLockedOut()) {
+            throw new AuthenticationException('Too many login attempts');
```
fightingtexasaggie: Perhaps it would be better to lie here and just say "invalid credentials" like below. If we don't, we indirectly leak whether or not this account exists. I attempted to log in with 'rjones' (not a user) more than the prescribed number of times and never got this message.  It's just something to be aware of.

PR #206, opened 2019-03-25
Looping around with different URL if return is not biblio.

mickel1138:


src/Pelagos/Util/PubLinkUtil.php
```diff
@@ -34,37 +34,57 @@ public function __construct(EntityHandler $entityHandler)
...
         curl_close($ch);
         $status = $curlInfo['http_code'];
+        $contentType = $curlInfo['content_type'];
 
-        if ($status == 200) {
-            $citation = new PublicationCitation($curlResponse, $style, $locale);
-        } else {
-            $citation = null;
+        if ($status == 200 and !preg_match('/text\/bibliography/', $contentType) and !$useCrossRef) {
+            return $this->curlCitation($doi, $style, $locale, true);
```
praneethpr: Recursive call

src/Pelagos/Util/PubLinkUtil.php
```diff
@@ -34,37 +34,57 @@ public function __construct(EntityHandler $entityHandler)
...
-            204 => 'The request was OK but there was no metadata available.',
-            404 => 'The DOI requested doesn\'t exist.',
-            406 => 'Can\'t serve any requested content type.',
-        );
+        $curlCitation = $this->curlCitation($doi);
+        $curlResponse = $curlCitation['curlResponse'];
+        $status = $curlCitation['status'];
+        $errorText = $curlCitation['errorText'];
 
+        $citation = new PublicationCitation($curlResponse, $style, $locale);
```
praneethpr: Irrespective of the status, we are creating this object, but it is persisted only in the Controller based on the status returned. This util class also has a entityManager injected which is never used. Should we refactor this ?

mickel1138: I agree, this function is doing way too much. Returning an entire set of status fields... It should just return data or an exception. I'm not sure if now is the time to refactor it?

mickel1138:
> Tested by adding an exception in the try/catch block to force it to use the catch (crossref) option, and using a new doi each time.

Trying using the DOI specified in the trouble ticket.

src/Pelagos/Util/PubLinkUtil.php
```diff
@@ -34,37 +34,66 @@ public function __construct(EntityHandler $entityHandler)
...
+     * @param string       $doi         The DOI of the publication.
+     * @param string|null  $style       The textual style desired.
+     * @param string|null  $locale      The character representation desired.
+     * @param boolean|null $useCrossRef Use the alternative way to retrieve the citation.
+     *
+     * @throws \Exception When CONTENT TYPE is not text/bibliography.
+     *
+     * @return array
+     */
+    private function curlCitation($doi, $style = PublicationCitation::CITATION_STYLE_APA, $locale = 'en-US', $useCrossRef = false)
```
praneethpr: I think we can remove the arguments passed to this method

PR #210, opened 2019-03-27
Feature/pelagos 3514 rewrite dataset monitoring

mickel1138:
All the trouble was in hash change, and the new attributes from tree. All working now.


src/Pelagos/Bundle/AppBundle/Resources/public/js/tree.js
```diff
@@ -77,7 +77,7 @@ function updateTree(tree) {
 
     if (tree.selected) {
         selected_node = $("#" + tree.name).jstree('get_selected');
-        if (typeof(selected_node) != 'undefined' && typeof(selected_node.id) != 'undefined' && selected_node.id != 'tree') {
+        if (typeof(selected_node) != 'undefined' && typeof(selected_node.attr('id')) != 'undefined' && selected_node.attr('id') != 'tree') {
```
praneethpr: Type comparison missing and double quotes

src/Pelagos/Bundle/AppBundle/Resources/views/DatasetMonitoring/index.html.twig
```diff
@@ -4,6 +4,14 @@
 {% block head %}
     {{ parent() }}
 
+    {{
+        add_css (
```
praneethpr: Are we going to do a refactor later? and add all these to the layout.js file?

mickel1138: I don't think it's worth the trouble using webpack for these applications. There is nothing to gain.

PR #214, opened 2019-04-01
Add new QueryBuilder that sorts results.

fightingtexasaggie:
The fos:elastica:populate command has always skipped around 500 entries, inconsistently even given the same input. This is due to an internal paginator fos:elastica uses that expects an orderded result.

Details: 
https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1478
https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1478

The suggested fix was to override the querybuilder used by something that would always provide a sorted result. 

To test, run fos:elastica:populate and verify count.

curl -X GET "localhost:9200<INDEX NAME>/INDEXNAME/_count" -H 'Content-Type: application/json' -d'
{
    "query": {
        "match_all": {}
    }
}
'

To list all indexnames:
curl -X GET http://localhost:9200/_cat/indices?v


src/Pelagos/Bundle/AppBundle/Command/RebuildElasticsearchIndexCommand.php
```diff
@@ -1,83 +0,0 @@
-<?php
```
praneethpr: I think its better if we keep the custom rebuild elasticsearch command because it can also index individual UDIs

fightingtexasaggie: This can be done with:
bin/console fos:elastica:populate --index pelagos
bin/console fos:elastica:populate --index search_pelagos

PR #221, opened 2019-04-03
Feature/pelagos 3562 add abstract and theme to index and search query

praneethpr:
To test:
Search with theme keywords Ex: BTEX PIANO benzene toluene
Should return BP.x750.000:0005, BP.x750.000:0027 and few other BP datasets

Search with abstract Ex: Hydrographic and water column data were collected during cruises EN509 and EN510 (sequential legs) using a CTD-rosette system aboard the R/V Endeavor. We characterized the physical and chemical properties in the water column at the ECOGIG seep and other study sites. The ship departed Gulfport MS on for cruise EN509 on 05/25/2012 and collected samples at ECOGIG and other sites before returning to Gulfport on 06/20/2012. Cruise EN510 departed Pascagoula MS (relocated to avoid a hurricane) on 06/25/2012 and collected samples at ECOGIG seep and control sites before returning to Gulfport MS on 07/05/2012. Binned profiles and bottle files are available.

Should return udi: R1.x132.134:0003, Try putting in different parts of the abstract.

### DIFF
--- a/src/Pelagos/Util/Search.php
+++ b/src/Pelagos/Util/Search.php
@@ -33,6 +33,21 @@ class Search
     protected $entityManager;
 
     /**
+     * Elastic index title field.
+     */
+    const INDEX_FIELD_TITLE = 'title';
+
+    /**
+     * Elastic index abstract field.
+     */
+    const INDEX_FIELD_ABSTRACT = 'abstract';
+
+    /**
+     * Elastic index dataset submission authors field.
+     */
+    const INDEX_FIELD_DS_AUTHOR = 'datasetSubmission.authors';
+
+    /**
      * Constructor.
      *
      * @param TransformedFinder $finder        The finder interface object.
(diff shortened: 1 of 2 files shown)


src/Pelagos/Util/Search.php
```diff
@@ -32,6 +32,26 @@ class Search
      */
     protected $entityManager;
 
+    /**
+     * Elastic index title field.
+     */
+    const INDEX_FIELD_TITLE = 'title';
```
praneethpr: Any suggestions in naming these constants better? not really happy with them

mickel1138: An array in a constant could work. Then just assign it a named (key/pair) array.

src/Pelagos/Util/Search.php
```diff
@@ -35,28 +35,38 @@ class Search
...
     /**
      * Elastic index dataset submission theme keywords field.
      */
-    const INDEX_FIELD_DS_THEME_KEYWORDS = 'datasetSubmission.themeKeywords';
+    const DS_THEME_KEYWORDS = 'datasetSubmission.themeKeywords';
+
+    /**
+     * Elastic index mapping.
+     */
+    const ELASTIC_INDEX_MAPPING = array(
```
mickel1138: You don't really need those other constants now do you?

praneethpr: Yeah, i think so. removed the individual ones

mickel1138:
Allow works fine.

PR #222, opened 2019-04-04
Feature/pelagos 3558 add additional properties to research group

praneethpr:


praneethpr:
Travis builds which execute codesniffer phpcs are skipping few syntax rules which are enforced on gcs

src/Pelagos/Entity/ResearchGroup.php
```diff
@@ -86,6 +86,21 @@ class ResearchGroup extends Entity
      */
     protected $name;
 
+    /**
+     * Name of a research group.
+     *
+     * @var string $name
```
mickel1138: set to correct property name.

src/Pelagos/Entity/ResearchGroup.php
```diff
@@ -86,6 +86,21 @@ class ResearchGroup extends Entity
...
+     *
+     * @access protected
+     *
+     * @ORM\Column(type="citext", unique=true, nullable=true)
```
mickel1138: Citext is case insensitive text, also uniqueness must be enforced.

src/Pelagos/Entity/ResearchGroup.php
praneethpr: 👍

src/Pelagos/Entity/ResearchGroup.php
fightingtexasaggie: I tested and it does recognize "oyster" and "Oyster" as being the same, which I think is the right behavior for uniqueness. (the violation just needs to be handled)

praneethpr: Added the unique constraint validator

PR #223, opened 2019-04-04
Feature/pelagos 3554 add remotely hosted properties to dataset submission entity

mickel1138:
Added attributes for remotely hosted.


mickel1138:
probably needs a migration too.

PR #227, opened 2019-04-08
Will this blend?

mickel1138:
Just seeing if the tests run.


mickel1138:
This will not work on Travis, we have an empty database. Too much hinges around data.

PR #228, opened 2019-04-08
Feature/pelagos 3565 add funding org facet

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/css/search.css
```diff
@@ -121,3 +121,9 @@ h1{
     display:inline-block;
 }
 
+#resgrp-facet{
```
mickel1138: a class name would be better, then you can apply it to all facets?

src/Pelagos/Bundle/AppBundle/Controller/UI/SearchPageController.php
```diff
@@ -62,7 +65,8 @@ private function getRequestParams(Request $request): array
             'query' => $request->get('query'),
             'page' => $request->get('page'),
             'options' => array(
-                'rgId' => $request->get('rgId')
+                'rgId' => ($request->get('resGrp')) ? str_replace('rg_', '', $request->get('resGrp')) : null,
```
mickel1138: Why the value prefixes?

src/Pelagos/Bundle/AppBundle/Resources/views/Search/results.html.twig
```diff
@@ -16,24 +16,47 @@
...
 
-                <ul class="browse-list list-unstyled long-list" style="margin-bottom: 0;">
-                    <li for="{{ group.id }}">
-                        <div class="checkbox block" style="margin-top: 0; margin-bottom: 0;">
-                            <label>
-                                <input class="rg-checkbox" type="checkbox" id="{{ group.id }}">
+                        <li for="{{ group.id }}">
+                            <div class="checkbox block" style="margin-top: 0; margin-bottom: 0;">
+                                <label>
+                                    <input class="rg-checkbox" type="checkbox" id="rg_{{ group.id }}">
```
mickel1138: Have you considered attributes, or data?

src/Pelagos/Bundle/AppBundle/Resources/public/css/search.css
praneethpr: I'm not sure if other facets will have really long lists. I think we wouldn't want to have scrolls for all facets

src/Pelagos/Bundle/AppBundle/Controller/UI/SearchPageController.php
praneethpr: For uniqueness

PR #230, opened 2019-04-11
Added pelagos command for doi orphan clean up from Datacite

praneethpr:


src/Pelagos/Util/DoiRest.ini.example
```diff
@@ -0,0 +1,3 @@
+doi_api_user_name = "username"
+doi_api_password = "password"
+url = "https://website.here"
```
mickel1138: new line

praneethpr: Now, I feel that separate file for REST API isn't required, only the URL is changed, probably i'll add additional REST URL in the DOIUtil.ini file and consume that.

src/Pelagos/Bundle/AppBundle/Command/DoiUpdateOrphanDataciteCommand.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace Pelagos\Bundle\AppBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+use GuzzleHttp\Client;
```
mickel1138: It would be more apparent if you aliased this use statement `as GuzzleClient` because I was confused below what client it was.

src/Pelagos/Bundle/AppBundle/Command/DoiUpdateOrphanDataciteCommand.php
```diff
@@ -0,0 +1,97 @@
...
+use GuzzleHttp\Exception\GuzzleException;
+
+/**
+ * This Symfony Command updates the Datacite orphan dois.
+ *
+ * @see ContainerAwareCommand
+ */
+class DoiUpdateOrphanDataciteCommand extends ContainerAwareCommand
+{
+
```
mickel1138: Is this an extra line?

src/Pelagos/Bundle/AppBundle/Command/DoiUpdateOrphanDataciteCommand.php
praneethpr: I think we don't have/use any conflicting classes with that same name. If you think it is confusing, we can add an alias to it.

src/Pelagos/Bundle/AppBundle/Command/DoiUpdateOrphanDataciteCommand.php
praneethpr: Yes, will remove it

PR #237, opened 2019-04-19
Feature/pelagos 3472 divorce from drupal

mickel1138:


mickel1138:
There are several merge conflicts, maybe we should merge develop some day.

praneethpr:
> There are several merge conflicts, maybe we should merge develop some day.
Conflicts in composer files would be challenging to solve

PR #239, opened 2019-04-19
Feature/pelagos 3557 add additional remotely hosted fields and proper logic to iso xml generating templates

mickel1138:
Needs to have PELAGOS-3555 **branch** to work, just switch to it, and accept a dataset with remotely hosted url, and go back into it, and check the box and set the values. Then switch to this branch and check the data landing page for it, should be all green check and an "R" on last check. Then check the metadata.


src/Pelagos/Entity/Dataset.php
```diff
@@ -665,6 +665,36 @@ public function updateAvailabilityStatus()
...
     }
+    
+    /**
+     * Whether this Dataset is available.
+     *
+     * @return boolean
+     */
+    public function isAvailable() : bool
+    {
+        return $this->availabilityStatus === DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE_REMOTELY_HOSTED or $this->availabilityStatus === DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE;
```
praneethpr: For better readability, I think using in_array would be better here (For ex: DatalandUI Controller) and below functions

mickel1138: That works for me, in_array always return a bool, and has a strict option. I'll make the changes. Also should we (I) refactor Dataland?

PR #242, opened 2019-04-23
Added additional read only fields to Person twig template

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/template/Person.html.twig
```diff
@@ -64,6 +64,11 @@
                         <tr>
                             <td>
                                 <div>
+                                    <b>Funding Organization :</b> {{ PersonResearchGroup.researchGroup.fundingCycle.fundingOrganization.name }}
```
mickel1138: Note spaces before : ?

src/Pelagos/Bundle/AppBundle/Controller/UI/PersonController.php
```diff
@@ -56,8 +58,13 @@ public function defaultAction($id = null)
                     $ui['PersonResearchGroupForms'][$personResearchGroup->getId()] = $formView;
                     $ui['PersonResearchGroupEditLabel'][$personResearchGroup->getId()]
                         = new EntityProperty($personResearchGroup, 'label');
+                    $listOfPersonsInResearchGroup = $personResearchGroup->getResearchGroup()->getPersonResearchGroups();
+                    foreach ($listOfPersonsInResearchGroup as $personResearch) {
```
mickel1138: What is there are more? The last one will get picked, is this correct?

src/Pelagos/Bundle/AppBundle/Controller/UI/PersonController.php
```diff
@@ -56,8 +58,13 @@ public function defaultAction($id = null)
...
+                        if ($personResearch->getRole()->getName() === ResearchGroupRoles::LEADERSHIP) {
+                            $personInLeadership = $personResearch->getPerson();
```
fightingtexasaggie: since the ORM doesn't have any order specified in the PersonResearchGroup entity for PersonResearchGroups, there is no guarantee the first, last, or nth will will be selected. I know in almost all cases there is only one, but in the off case there isn't, it's just going to return any one (and may vary per fetch).

src/Pelagos/Bundle/AppBundle/Resources/views/template/Person.html.twig
praneethpr: 👍

src/Pelagos/Bundle/AppBundle/Controller/UI/PersonController.php
praneethpr: Changed logic to show multiple

src/Pelagos/Bundle/AppBundle/Controller/UI/PersonController.php
praneethpr: Changed the logic to show multiple PI's if present

PR #247, opened 2019-04-30
Changed logic to update doi with blank fields when doi is not in draf…

praneethpr:
…t state

### DIFF
--- a/src/Pelagos/Util/DOIutil.php
+++ b/src/Pelagos/Util/DOIutil.php
@@ -330,8 +330,6 @@ class DOIutil
             $ch = curl_init();
             curl_setopt($ch, CURLOPT_URL, $this->url . '/id/' . $doi);
             curl_setopt($ch, CURLOPT_USERPWD, $this->doiusername . ':' . $this->doipassword);
-            curl_setopt($ch, CURLOPT_POST, true);
-            curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
             curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'DELETE');
             $output = curl_exec($ch);
 
--- a/src/Pelagos/Util/DOIutil.php
+++ b/src/Pelagos/Util/DOIutil.php
@@ -319,7 +319,7 @@ class DOIutil
      * @throws HttpClientErrorException When there was an 4xx error negotiating with EZID.
      * @throws HttpServerErrorException When there was an 5xx error negotiating with EZID.
      *
-     * @return boolean
+     * @return void
      */
     public function deleteDOI($doi)
     {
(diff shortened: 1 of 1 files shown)


src/Pelagos/Util/DOIutil.php
```diff
@@ -319,42 +319,47 @@ public function getDOIMetadata($doi)
      * @throws HttpClientErrorException When there was an 4xx error negotiating with EZID.
      * @throws HttpServerErrorException When there was an 5xx error negotiating with EZID.
      *
-     * @return array
+     * @return boolean
```
mickel1138: Always returns true if successful, or throws exception, might as well return void?
Because you would have to call this in a try, checking the return value wouldn't help you much.
It's just a thought, I don't know what the best to handle this is, just my opinion.

praneethpr: That is right, there is no significance of the return value from this method, and even in the Consumer it doesn't expect a return value when this method is called

src/Pelagos/Util/DOIutil.php
```diff
@@ -319,42 +319,47 @@ public function getDOIMetadata($doi)
...
+            curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'DELETE');
+            $output = curl_exec($ch);
+
+            $httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+            curl_close($ch);
+
+            //check to see if it worked.
+            if (200 != $httpCode) {
+                $expMsg = "ezid failed with:$httpCode($output)";
+                if ($httpCode >= 400 and $httpCode <= 499) {
```
mickel1138: What error code would happen when a non-reserved (non-draft) would try to be deleted?

praneethpr: Should get a bad request error code

mickel1138: Could you determine that's in not in draft, then maybe do the else. It doesn't make much difference I guess.

src/Pelagos/Util/DOIutil.php
```diff
@@ -319,42 +319,47 @@ public function getDOIMetadata($doi)
...
-            }
-        }
-
-        $metadata = array();
-        foreach (explode("\n", $output) as $line) {
-            $split = preg_split('/:/', $line, 2);
-            if (count($split) > 1) {
-                $metadata[$split[0]] = trim($split[1]);
+        $doiMetadata = $this->getDOIMetadata($doi);
+        if ($doiMetadata['_status'] === 'reserved') {
```
praneethpr: This check is to determine, if the doi in Datacite is in Draft (in EZ id terms, 'reserved')

mickel1138: This works, just looking for shortcuts, but there aren't any.

PR #248, opened 2019-05-02
Feature/pelagos 3581 create pelagos command to compare dois between griidc and datacite

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/DoiComparisonCommand.php
```diff
@@ -0,0 +1,400 @@
...
+ * @see ContainerAwareCommand
+ */
+class DoiComparisonCommand extends ContainerAwareCommand
+{
+    /**
+     * The file output array which stores the data.
+     *
+     * @var array
+     */
+    protected $fileOutputArray;
```
mickel1138: Not used anywhere?

praneethpr: Removed it

PR #250, opened 2019-05-06
Fixed tree in dataset monitoring

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/js/tree.js
```diff
@@ -107,9 +107,9 @@ function updateTree(tree) {
                         } else {
                             url = Routing.generate("pelagos_api_tree_get_letters");
                         }
-                    }
-                    else {
-                        nodeId = node.id;
+                    } else {
+                        nodeId = node.a_attr['id'];
```
mickel1138: You should be able to get the id from node['id']. All nodes should have id. (Meaning, actually not required by jsTree, but we should have them everywhere.)

praneethpr: Not all twig templates have it, Either i can use node['id'] and add "id" to letters.twig, researchers.twig. All of them have a_attr['id'] except for reaserch_awars.twig, I can flip it if you want.

mickel1138: Not sure where the best place is. I also don't if a A tag needs an id.
This is what jsTree says about the properties: https://www.jstree.com/docs/json/

praneethpr: Okay, I'll remove the ids from A tags in the tree templates, and put it at the node level, so that we are consistent

PR #254, opened 2019-05-07
Fixed css for dataset monitoring

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/css/dataset-monitoring.css
```diff
@@ -41,3 +41,7 @@
 .display-table-cell {
     display: table-cell;
 }
+
+#pelagos-content table {
```
mickel1138: What "breaks" if we just change this in the main css. (The one we are overwriting)?

praneethpr: I don't know, didn't try it. That id is in layout.html.twig, we have to check every page if something has broken

PR #256, opened 2019-05-07
Feature/pelagos 3600 add correct logged in text

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/layout.html.twig
```diff
@@ -67,7 +67,7 @@
                         <div id="block-pelagos-pelagos-login-logout-links" class="block block-pelagos">
                             <div class="content">
                             {% if is_granted('ROLE_USER') %}
-                            <a href="{{ path('security_logout') }}">Logout</a>
+                            <a href="{{ path('security_logout') }}">Logged in as: {{ app.user.username }}  | Logout</a>
```
mickel1138: "logged in as: {username}" and "Logout" are two different links. The logout only does logout, but clicking on the username takes you to the Drupal User page. Maybe we could link it to the Person page? That way the user could update their info.

praneethpr: Thanks, I didn't really notice that. Probably discuss with the Product owners what they want to do?

mickel1138: Yes, sounds good. Or just split them for now.

praneethpr: I split them and gave the person page link to the logged in as username text

PR #257, opened 2019-05-09
Feature/pelagos 3603 move change password out of account route

mickel1138:


src/Pelagos/Bundle/AppBundle/Resources/views/Menus/main.html.twig
```diff
@@ -34,7 +34,7 @@
                                             <span class="sf-sub-indicator"> »</span></span>
                         <ul class="sf-hidden" style="float: none; width: 13.5em;">
                             <li class="first odd "><a href="{{ path('pelagos_app_ui_account_default') }}">Account Request</a></li>
-                            <li class="middle even "><a href="{{ path('pelagos_app_ui_account_passwordreset') }}">Password Reset</a></li>
+                            <li class="middle even "><a href="{{ path('pelagos_app_ui_account_password') }}">Password Reset</a></li>
```
fightingtexasaggie: We could change the verbage here to change/reset password, or base the verbage on whether or not they're logged in. I guess the safe bet is to see if the P.O. says anyting in testing. ;-)

src/Pelagos/Bundle/AppBundle/Controller/UI/AccountController.php
```diff
@@ -155,7 +155,7 @@ public function sendVerificationEmailAction(Request $request)
      * This verifies that the token has authenticated the user and that the user does not already have an account.
      * It then provides the user with a screen to establish a password.
      *
-     * @Route("/verify-email")
+     * @Route("/account/verify-email")
```
fightingtexasaggie: For some reason this is causing the token to not be recognized on a password reset action, resulting in the "The link you have selected is invalid. The link may have already been used to create an account. If you have forgotten your password please visit password reset. For any additional help please contact griidc@gomri.org. " error.

fightingtexasaggie: Fixed by removing old %pelagos-prefix% in firewall.

PR #258, opened 2019-05-09
Feature/pelagos 3605 fix css for login page

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/Security/login.html.twig
```diff
@@ -40,21 +40,28 @@
...
+                        </div>
+                    </div>
+                    <div class="row fl-controls-left">
+                        <div class="login-page-text form-item">
+                            {{ form_row(form._password) }}
+                            <div class="description">Enter the password that accompanies your username.</div>
+                        </div>
+
+                        <div>
+                            <label for="rememberMe" class="login-page-text" style="display:inline; float:none; vertical-align:15%;">
```
mickel1138: `for` is not needed, there is no related input that goes with this label.

PR #260, opened 2019-05-10
Feature/pelagos 3606 create unique doi locally

fightingtexasaggie:


src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +333,56 @@ function ($matches) {
...
+     * This function queries Datacite to determine if the DOI is already in use.
+     *
+     * @param string $doi The DOI to be checked.
+
+     * @throws HttpClientErrorException When there is a 4xx error negotiating with Datacite, other than a 400.
+     * @throws HttpServerErrorException When there is a 5xx error negotiating with Datacite.
+     *
+     * @return boolean True if returned if the DOI exists, false if it does not.
+     */
+    protected function checkDoiExists($doi)
```
praneethpr: I feel checking the doi on datacite is overkill. If we still want to check the DOI exists on datacite, we can use the getDoiMetadata method to verify it doesn't exist. Similar method exists in DoiConsumer.php (doiAlreadyExists() mehtod)

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +333,56 @@ function ($matches) {
...
+     */
+    public function generateDoi()
+    {
+        $encoder = new Base32CrockfordEncoder();
+        // 1099511627775 encodes to the longest 8 character Crockford 32 string.
+        $max = 1099511627775;
+        // Start at 1 (0 is problematic as library does not produce checksum for 0.)
+        $random = mt_rand(1, $max);
+        // Add shoulder and remove the checksum character.
+        $doi = $this->doishoulder . '/' . substr($encoder->encode($random), 0, -1);
```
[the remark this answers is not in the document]
fightingtexasaggie: The last character it generates is a checksum, which isn't needed for what we're doing. The substr -1 is simply chopping it off.

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: I tried using the getDoiMetadata method, as also Mike suggested, but since we are having to rely specifically on catching a 400 for detectection, I thought we'd end up relying on a catch as our detection method. I'll take a look at the DoiConsumer's method though.

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +333,56 @@ function ($matches) {
...
+        while ($this->checkDoiExists($doi)) {
+            $random = mt_rand(0, $max);
```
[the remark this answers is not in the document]
fightingtexasaggie: yep. good catch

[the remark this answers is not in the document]
fightingtexasaggie: Also, the range (729299307643, 729299307645) can be used to force a frequent collision with 10.33516/n76t0k3x which I have put in as a draft-mode DOI on datacite's test system. (and maybe a couple dump statements to test)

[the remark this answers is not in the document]
fightingtexasaggie: Also, to test non 400 conditions, I commented out the
#curl_setopt($ch, CURLOPT_URL, $this->url . '/id/' . $doi);
and used:
curl_setopt($ch, CURLOPT_URL, 'https://httpstat.us/404');

to insure the other errors still perculated through. This is otherwise difficult to test.

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: I replaced with a call to getDOIMetadata(), however, the check is still needed as the one made by the consumer will consider a collision an existing DOI, which is an error, where is here we just want to be able to pick a unique DOI.

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +334,51 @@ function ($matches) {
...
+    {
+        try {
+            $doiUtil = new DOIutil();
+            $doiUtil->getDOIMetadata($doi);
+        } catch (HttpClientErrorException $exception) {
+            if ($exception->getCode() == 400) {
+                // No, this DOI does not exist upstream (Datacite) thus can be used.
+                return false;
+            } else {
+                throw new HttpClientErrorException($exception->getMessage(), $exception->getCode());
```
praneethpr: We are throwing this exception class inside the same Exception class that is being caught. And this thrown exception is not being caught anywhere

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +334,51 @@ function ($matches) {
...
+     *
+     * @return boolean True if returned if the DOI exists, false if it does not.
+     */
+    public function checkDoiExistsExternal(string $doi): bool
+    {
+        try {
+            $doiUtil = new DOIutil();
+            $doiUtil->getDOIMetadata($doi);
+        } catch (HttpClientErrorException $exception) {
+            if ($exception->getCode() == 400) {
```
praneethpr: Will it be enough to just create a new DOI, if we get a Client side exception. I've not seen any other 4xx codes returned by the GET API

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: It's re-throwing all other 400's except 400 itself. It will perculate up to DoiConsumer's issueDoi() method, which would result in the attempt being logged. 500's aren't caught at all, and would also perculate up to this point for a retry 10 minutes later. Previously mintDOI would have thrown the 500. It still gets handled the same way.

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: I think the more proper response should have been a 404 from datacite, and is case they do this is the future, I'll check for both 400 and 404. If they ever do correct this in the future, we should probably re-throw the 400.

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +334,51 @@ function ($matches) {
...
+     *
+     * @throws HttpClientErrorException If a non 400 error is encountered.
+     *
+     * @return boolean True if returned if the DOI exists, false if it does not.
+     */
+    public function checkDoiExistsExternal(string $doi): bool
+    {
+        try {
+            $doiUtil = new DOIutil();
+            $doiUtil->getDOIMetadata($doi);
```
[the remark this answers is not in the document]
fightingtexasaggie: I guess I forgot where I was!

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: updated to look for either 404 or 400 here.

src/Pelagos/Util/DOIutil.php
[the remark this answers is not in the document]
praneethpr: I think you forgot to change this?

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +334,51 @@ function ($matches) {
...
+     */
+    public function generateDoi()
+    {
+        $encoder = new Base32CrockfordEncoder();
+        // 1099511627775 encodes to the longest 8 character Crockford 32 string.
+        $max = 1099511627775;
+        // Start at 1 (0 is problematic as library does not produce checksum for 0.)
+        $random = mt_rand(1, $max);
+        // Add shoulder and remove the checksum character.
+        $doi = $this->doishoulder . '/' . substr($encoder->encode($random), 0, -1);
```
praneethpr: The shoulder on prod is 10.7266/n7, so that '/' can be removed

fightingtexasaggie: I didn't realize N7 was part of our shoulder! I'll have to shorten the DOI generator by 2 characters.

src/Pelagos/Util/DOIutil.php
```diff
@@ -71,9 +72,6 @@ public function __construct()
      * @param string $status          Status of the DOI, by default is reserved.
      * @param string $resourcetype    Type for DOI Request, by default Dataset.
      *
-     * @throws HttpClientErrorException When there was an 4xx error negotiating with EZID.
```
mickel1138: If you don't handle (try) the calling function, then this function can throw exceptions.

src/Pelagos/Util/DOIutil.php
mickel1138: If I understand correctly, we own the entire doi prefix now.

wwnick: We always owned the entire prefix, the shoulder was just a EZID thing, carried over to EZAPI for compatibility.  Creating a DOI we can ditch the N7 if we want  See https://ezid.cdlib.org/learn/id_basics for more details

fightingtexasaggie: /N7 was not part of our prefix, and we really intermixed the wording on shoulder vs prefix in our prior code. I checked with the PO and we don't need to maintain the /N7 thing as that was specific to ezid's. All we need is 8 characters after the assigned prefix.

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: I'm stuck with having to catch a 400 (or 404) specifically. The rest I want to re-throw.

fightingtexasaggie:
For newest change, make sure and run composer install as I changed a parameter name to correct some (confusing) wording.

src/Pelagos/Util/DOIutil.php
```diff
@@ -85,50 +88,23 @@ public function mintDOI(
...
-        if (!in_array($httpCode, [200, 201])) {
-            $expMsg = "ezid failed with:$httpCode($output)";
-            if ($httpCode >= 400 and $httpCode <= 499) {
-                throw new HttpClientErrorException($expMsg, $httpCode);
-            } elseif ($httpCode >= 500 or $httpCode == 0) {
-                throw new HttpServerErrorException($expMsg, $httpCode);
-            }
+        try {
+            $doi = $this->generateDoi();
+        } catch (HttpClientErrorException $exception) {
```
mickel1138: So I looked around a bit, and rethrowing the same exception is not pretty.
For it to have meaning a different exception should be thrown at the right level.
i.e. DuplicateDoiException, HavingTroubleWithDataCiteException, CouldntCreateDoiException (examples).

Also, @praneethpr brought up that you can simplify this if you just put the doi (id) creation into the createDoi function, but leave the doi value as an option. (at the end or so).

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +352,72 @@ function ($matches) {
...
+        $doi = $this->doiprefix . '/' . substr($encoder->encode($random), 0, -1);
+        try {
+            $doiExists = $this->checkDoiExistsExternal($doi);
+        } catch (HttpClientErrorException $exception) {
+            throw new HttpClientErrorException($exception->getMessage(), $exception->getCode());
+        } catch (HttpServerErrorException $exception) {
+            throw new HttpServerErrorException($exception->getMessage(), $exception->getCode());
+        }
+        while ($doiExists) {
+            $random = mt_rand(1, $max);
```
praneethpr: This whole block of code is used twice, it can be created into a separate function

src/Pelagos/Util/DOIutil.php
```diff
@@ -375,4 +352,72 @@ function ($matches) {
...
+            $doi = $this->doiprefix . '/' . substr($encoder->encode($random), 0, -1);
+            try {
+                $doiExists = $this->checkDoiExistsExternal($doi);
+            } catch (HttpClientErrorException $exception) {
+                throw new HttpClientErrorException($exception->getMessage(), $exception->getCode());
+            } catch (HttpServerErrorException $exception) {
+                throw new HttpServerErrorException($exception->getMessage(), $exception->getCode());
+            }
+        }
+        return $doi;
```
mickel1138: Should the DOI be uppered?

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: yes, but each call would need to be run in a try/catch...

src/Pelagos/Util/DOIutil.php
```diff
@@ -85,50 +88,23 @@ public function mintDOI(
...
+            $doi = $this->generateDoi();
+        } catch (HttpClientErrorException $exception) {
+            throw new HttpClientErrorException($exception->getMessage(), $exception->getCode());
+        } catch (HttpServerErrorException $exception) {
+            throw new HttpServerErrorException($exception->getMessage(), $exception->getCode());
         }
 
-        $doi = preg_match('/^success: (doi:\S+)/', $output, $matches);
+        try {
+            $this->createDOI($doi, $url, $creator, $title, $publisher, $publicationYear, $status, $resourcetype);
```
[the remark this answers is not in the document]
fightingtexasaggie: That's what I ended up doing. much shorter too.

src/Pelagos/Util/DOIutil.php
```diff
@@ -158,6 +88,16 @@ public function createDOI(
         $status = 'reserved',
         $resourcetype = 'Dataset'
     ) {
+        if ('mint' === $doi) {
```
mickel1138: I think it's better if this was just null or blank. That way you can check empty

src/Pelagos/Util/DOIutil.php
```diff
@@ -199,7 +139,7 @@ public function createDOI(
             }
         }
 
-        return true;
+        return $doi;
```
mickel1138: The caller function createDoi() previous expected TRUE as return, however it doesn't appear any caller was checking for this. It's good to check if the function is still compatible.

src/Pelagos/Util/DOIutil.php
praneethpr: We don't have to check at all, if nothing is passed. No other function is going to call createDOI() with a doi parameter. We can just remove the first parameter

src/Pelagos/Util/DOIutil.php
fightingtexasaggie: I did check in the consumer to make sure I wasn't breaking anything. I'll take a second careful look just to be sure.

src/Pelagos/Util/DOIutil.php
mickel1138: > We don't have to check at all, if nothing is passed. No other function is going to call createDOI() with a doi parameter. We can just remove the first parameter

Don't we need to have a function to create existing DOI (in our system)?

praneethpr: Hmm.. I don't know if there is a use case for it. Post migration to Datacite, it was useful

PR #264, opened 2019-05-15
Feature/pelagos 3617 add padding around text in cells

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/css/dataset-monitoring.css
```diff
@@ -43,5 +43,9 @@
 }
 
 #pelagos-content table {
-     border-collapse: collapse;
+    border-collapse: collapse;
+}
+
+table.datasets td {
```
mickel1138: This could go in the layout.css (or common.css, whichever does ui) because it previously loaded in a global css template.

mickel1138: https://bitbucket.org/griidc/drupal-theme/src/amazee/css/pelagos-full.css

praneethpr: I don't know whether this file was the right one, because i tried adding it to pelagos-module.css and without the !important, it doesn't seem to work. And I don't know if changing it in the global css files would affect other things

mickel1138: If you look on production, and see where the cell padding comes from, you'll see it's from this file. In the end it doesn't matter that much. Just remember if you see anything else related to cell padding, we may want to move this higher up.

PR #266, opened 2019-05-15
Removed the link

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/views/layout.html.twig
```diff
@@ -67,7 +67,7 @@
                         <div id="block-pelagos-pelagos-login-logout-links" class="block block-pelagos">
                             <div class="content">
                             {% if is_granted('ROLE_USER') %}
-                                <a href="{{ path("pelagos_app_ui_person_default", {id: app.user.person.id }) }}">
+                                <a href="">
```
mickel1138: I had to look this up, but there is a difference between "" (blank) link and "#" (hash) link.
A blank link will refresh itself, while a hash link will not. But will put a has on the end of the url.

praneethpr: Changed it 👍

PR #268, opened 2019-05-15
Added tree to toggle for on click event

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/js/dataset-monitoring.js
```diff
@@ -6,12 +6,17 @@ $(document).ready(function() {
...
-    $("#tree").bind("loaded.jstree", function(e, data) {
+    let tree = $("#tree");
+    tree.bind("loaded.jstree", function(e, data) {
         if (window.location.hash != "") {
             $(window).trigger("hashchange");
         }
     });
 
+    tree.on('select_node.jstree', function (e, data) {
+        data.instance.toggle_node(data.node);
```
mickel1138: The ticket says to do nothing when already expanded, so may have to check the state of data.node.

praneethpr: They thought it would be more work for us to close the tree on click, i'll check with the PO

fightingtexasaggie: This works to detect it, if the PO really wants it to not re-close: if (data.instance.is_closed(data.node))

praneethpr: POs were okay with the expand and collapse on click

PR #269, opened 2019-05-15
Removed hyperlink and added style to match

praneethpr:


praneethpr:
> The logged-in section looks exactly like production now. Although this is what was asked for, I do have to wonder if they really want the underline, as that implies a link which we have removed. IDK...we'll see.

I checked with William, he said he doesn't want the hyperlink, but wants it to look the same

PR #275, opened 2019-05-21
Added check to exclude unavailable datasets from comparison

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/DoiComparisonCommand.php
```diff
@@ -384,14 +386,20 @@ private function isUrlValid(string $url, string $needle): bool
...
      *
      * @return void
      */
-    private function doesStringExist(array $metadataElement, string $comparisonElement): void
+    private function doesStringExist(array $metadataElement, string $comparisonElement, Dataset $dataset = null): void
     {
         if (empty($metadataElement[$metadataElement['field']])) {
+            if (!empty($dataset)
+                and $dataset->getAvailabilityStatus() === DatasetSubmission::AVAILABILITY_STATUS_NOT_AVAILABLE
+            and $metadataElement['field'] === 'author') {
```
mickel1138: Does this need to be tabbed over?

praneethpr: Yes, I don't know why GCS didn't catch it

PR #277, opened 2019-05-22
Made the controller redirect to "/" when not in debug mode.

mickel1138:
To test, invalidate the test, or change environment to non-development one.

### DIFF
--- a/src/Pelagos/Bundle/AppBundle/Controller/DefaultController.php
+++ b/src/Pelagos/Bundle/AppBundle/Controller/DefaultController.php
@@ -22,7 +22,11 @@ class DefaultController extends Controller
      */
     public function indexAction()
     {
-        return $this->render('PelagosAppBundle:Default:index.html.twig');
+        if ($this->get('kernel')->isDebug()) {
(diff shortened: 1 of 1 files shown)


mickel1138:
Turn on intercept redirects (to true) if you don't want to really redirected.

src/Pelagos/Bundle/AppBundle/Controller/DefaultController.php
```diff
@@ -22,7 +22,11 @@ class DefaultController extends Controller
      */
     public function indexAction()
     {
-        return $this->render('PelagosAppBundle:Default:index.html.twig');
+        if ($this->get('kernel')->isDebug()) {
+            return $this->render('PelagosAppBundle:Default:index.html.twig');
+        } else {
+            return $this->redirect('/', 302);
```
praneethpr: Should this be 301?

mickel1138: 301 is permanent redirect, i'm not sure how that might behave.

PR #282, opened 2019-05-29
Check to see if array is empty before accessing elements from it

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/DoiComparisonCommand.php
```diff
@@ -85,8 +85,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
                     'doi' => $doi['attributes']['doi'],
                     'url' => $doi['attributes']['url'],
                     'udi' => $this->getUdi($doi['attributes']['url']),
-                    'title' => str_replace(',', '', $doi['attributes']['titles'][0]['title']),
-                    'author' => str_replace(',', '', $doi['attributes']['creators'][0]['name']),
+                    'title' => str_replace(
+                        ',',
+                        '',
+                        $this->isArrayEmpty($doi['attributes']['titles'])
```
mickel1138: Wouldn't it be better is to use `array_key_exists` that way you can for sure tell if 'name' is there. Not empty doesn't mean the key exists...

praneethpr: To check if 'name' exists, I think i need to do a nested array_key_exists

praneethpr:
> tested against current data, worked without error. If you want to switch to the array_key_exists I'll re-test, otherwise I'm fine with it.

I added array_key_exists() inside the if loop, so that it looks for the key inside the nested array

src/Pelagos/Bundle/AppBundle/Command/DoiComparisonCommand.php
```diff
@@ -135,6 +145,25 @@ private function getResourceType(array $types): string
...
 
+    /**
+     * Checks if the array is empty.
+     *
+     * @param array  $doiMetadataElementArray Metadata element array to check if empty and key exists.
+     * @param string $keySearch               The key which needs to be checked if it exists in the array.
+     *
+     * @return boolean
+     */
+    private function doesKeyExist(array $doiMetadataElementArray, $keySearch): bool
```
mickel1138: Cool, I like this a lot better. Now there is no doubt that the value/key we need is there.

PR #284, opened 2019-05-29
Added google analytics, and needed config and parameters.

mickel1138:


mickel1138:
Not sure how to test this, but we can discuss and make a dummy/test ID.

PR #285, opened 2019-05-30
Feature/pelagos 3648 rewrite doi crud using rest api

praneethpr:


src/Pelagos/Util/DOIutil.php
```diff
@@ -84,68 +90,61 @@ public function generateDoi() : string
...
      */
     public function createDOI(
         $doi,
         $url,
         $creator,
         $title,
-        $publisher,
         $publicationYear,
-        $status = 'reserved',
+        $publisher = 'Harte Research Institute',
```
mickel1138: I liked it better when it was explicitly called. The utility shouldn't really know anything about who it's doing business for.

src/Pelagos/Entity/DOI.php
```diff
@@ -153,6 +153,6 @@ public function setPublicDate(\DateTime $publicDate)
      */
     public function __toString()
     {
-        return (string) 'doi:' . $this->doi;
```
mickel1138: doi is part of the display value. As recommended by DOI org. That's what makes this function unique, otherwise it returned the same thing as getDoi()

src/Pelagos/Util/DOIutil.php
```diff
@@ -264,63 +254,38 @@ public function getDOIMetadata($doi)
...
-                 'none supplied',
-                 '2019',
-                 'unavailable'
-             );
+            $this->updateDOI(
+                $doi,
+                'http://datacite.org/invalidDOI',
+                '(:null)',
+                'inactive',
+                '2019',
```
mickel1138: What are we going to do next year?

src/Pelagos/Entity/DOI.php
praneethpr: Yeah, I can change it back. We should probably use this at landing and other display pages, I assumed that it wasn't necessary to have the 'doi:'

src/Pelagos/Util/DOIutil.php
praneethpr: Haha. It just needs some year, and after the Dataset is deleted we can't get any attributes from the object, it won't take 0000, so I just gave it 2019

src/Pelagos/Util/DOIutil.php
praneethpr: Hmm.. I dont know I thought it is just another variable, which is going to be constant atleast for now. We don't need to explicitly call it

mickel1138: Yes, it's a variable but not a constant. In my opinion a utility should be "dumb" and now know anything. Also it's easy to overlook who sets this value. The caller should be responsible for setting it.

praneethpr: Okay I can change it back

PR #287, opened 2019-06-10
Added griidc banner stylesheets

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/css/griidc_site_banner_screen.css
```diff
@@ -0,0 +1,72 @@
...
+body.adminimal-menu.overlay {
+    padding-top: 80px!important;
+    margin-top: 30px!important;
+    margin-bottom: 30px!important;
+}
+
+#siteBannerFooterBanner,#siteBannerHeaderBanner{background-color:#EED202;color:#000000;}
+
+.atlassian-btns {
+    top: 20px;
```
mickel1138: I would make this margin-top, because it's already set at -2px, and it just barely overlaps to banner.

PR #288, opened 2019-06-11
Feature/pelagos 3655 create maintenance mode mechanism

mickel1138:


mickel1138:
Example file:
    maintenance_text = "text"
    maintenance_color = "#color"

src/Pelagos/Util/MaintenanceMode.php
```diff
@@ -0,0 +1,108 @@
...
+     * @var TransformedFinder
+     */
+    protected $fileName;
+
+    /**
+     * The kernel root path.
+     *
+     * @var string
+     */
+    private $kernelRootDir;
```
mickel1138: Not needed!

src/Pelagos/Util/MaintenanceMode.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+namespace Pelagos\Util;
+
+/**
+ * A utility to create and issue DOI from EZID API.
```
praneethpr: Fix doc block

src/Pelagos/Util/MaintenanceMode.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+class MaintenanceMode
+{
+    /**
+     * FOS Elastica Object to find elastica documents.
```
praneethpr: Fix doc block

src/Pelagos/Util/MaintenanceMode.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public function activateMaintenanceMode(string $text, string $color)
+    {
+        $contents = 'maintenance_text = "' . $text . '"' . PHP_EOL .
+            'maintenance_color = "' . $color . '"' . PHP_EOL;
+        file_put_contents($this->fileName, $contents);
+    }
+
+    /**
+     * Create the maintenance mode ini file.
```
praneethpr: Change annotation to delete

src/Pelagos/Util/MaintenanceMode.php
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @return void
+     */
+    public function deactivateMaintenanceMode()
```
praneethpr: return type not mentioned in function statement

src/Pelagos/Util/MaintenanceMode.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Create the maintenance mode ini file.
+     *
+     * @param string $text  The banner text.
+     * @param string $color The banner background color.
+     *
+     * @return void
+     */
+    public function activateMaintenanceMode(string $text, string $color)
```
praneethpr: return type not mentioned in function statement

src/Pelagos/Util/MaintenanceMode.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Is the system in maintenance mode.
+     *
+     * @return boolean If in maintenance mode.
+     */
+    public function isMaintenanceMode() : bool
+    {
+        return file_exists($this->fileName);
```
fightingtexasaggie: Maybe we should also require that the file exists and has a value set for either the maintenance_text or maintenance_color before returning true. This would allow us to put an ini file in place and comment out lines, perhaps having commented-out parameters for different situations with pre-agreed-upon wording.  We could also do this renaming the .ini file to something else though. I'm good with it as is, but wanted to see what y'all thought.

mickel1138: While this is a ini file, it's really a system file, controller by the system. Being able to put a ini file place should be considered a hack. As a future improvement this library can be used to create this file, and we should set this up in a controller, or command.

PR #290, opened 2019-06-12
Maintenance mode interface.

mickel1138:


src/Pelagos/Bundle/AppBundle/Controller/UI/MaintenanceModeController.php
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @param Request $request The Symfony request object.
+     *
+     * @Route("/")
+     *
+     * @return Response A Response instance.
+     */
+    public function defaultAction(Request $request)
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
```
praneethpr: I think we should give access only to DRPMs

src/Pelagos/Bundle/AppBundle/Resources/views/MaintenanceMode/maintenanceMode.html.twig
```diff
@@ -0,0 +1,40 @@
...
+<label for="bannertext">Banner Text:<label>
+<input type="text" size="200" id="bannertext" name="bannertext" placeholder="Your Banner Text Here" required value="{{ getMaintenanceModeText() }}">
+<label for="bannercolor">Banner Color:<label>
+<input type="color" id="bannercolor" name="bannercolor" value="{{ '#EED202'|maintenanceModeColor }}">
+<label for="bannermode">Mode:</label>
+<input type="radio" id="bannermodeon" name="bannermode" value="activate" {{isMaintenanceMode() ? 'checked' : ''}}>Active
+<input type="radio" id="bannermodeoff" name="bannermode" value="deactivate" {{not isMaintenanceMode() ? 'checked' : ''}}>Off<br>
+<input type="submit">
+
+{% endblock %}
```
praneethpr: New line at the end

src/Pelagos/Bundle/AppBundle/Twig/Extensions.php
```diff
@@ -243,6 +247,24 @@ public function getMaintenanceModeColor() : ? string
...
     }
 
+    /**
+     * Gets maintenance mode color.
+     *
+     * @param string $color The color text.
+     *
+     * @return string|null Returns maintenance mode banner color.
+     */
+    public function maintenanceModeColor(string $color) : string
```
praneethpr: We have two functions in this class which gets the maintenance mode color?

src/Pelagos/Bundle/AppBundle/Controller/UI/MaintenanceModeController.php
mickel1138: Yes, have to change that.

src/Pelagos/Bundle/AppBundle/Twig/Extensions.php
mickel1138: One of them is a filter, but this one can call the other one.

PR #293, opened 2019-06-14
Not needed since login no longer required for DL.

fightingtexasaggie:
The dl_attempt_cookie was somehow still being set although we're not requiring login. Rather than try to fix the un-needed code, I thought it would be best to remove it altogether. Download works properly without it now. (For purposes of testing, griidc-dev2 and griidc-test-c7 are populated with full griidc data for any dataset <= 100MB.)


praneethpr:
I think this bug exists in develop branch also, should we merge it into develop first and then merge develop back to this epic?

PR #298, opened 2019-06-24
Upgrade the jquery validator to 1.19, and removed formHash which was …

praneethpr:
…unused and undefined during form submission


src/Pelagos/Bundle/AppBundle/Resources/public/js/datasetReview.js
```diff
@@ -64,7 +64,7 @@ $(document).ready(function(){
         ignore: ".ignore,.prototype",
         submitHandler: function (form) {
             if ($(".ignore").valid()) {
-                formHash = regForm.serialize();
+                regForm.serialize();
```
mickel1138: This is not needed at all.

PR #299, opened 2019-06-24
Set/clear username session cookie via listeners.

fightingtexasaggie:
Look for GRIIDC_USERNAME cookie on your debug console. (will open a related PR for Drupal that'll use this cookie)


app/config/security.yml
```diff
@@ -39,6 +39,7 @@ security:
             logout:
                 path: security_logout
                 target: pelagos_homepage
+                handlers: [pelagos.logout_listener]
```
mickel1138: It's not really clear from documentation, but it looks like `handlers` will go way in Symfony 4.
You probably wanna use `success_handler` anyway because that doesn't just handle the logout, and actual logout success. And in that case you want to implement a `LogoutSuccessHandlerInterface`

mickel1138: Also, it's not a listner, but a handler.

app/config/services.yml
```diff
@@ -124,7 +124,16 @@ services:
         autowire: true
         tags:
             - { name: monolog.logger, channel: login_attempts }
- 
+
+    pelagos.login_listener:
+        class: Pelagos\Event\LoginEventListener
+        arguments: ["@security.token_storage"]
+        tags:
+          - { name: kernel.event_listener, event: security.interactive_login, method: onSecurityInteractiveLogin }
```
mickel1138: Instead of using this new event, you can just stick your code in LoginFormAuthenticator. It has a nice function for you:
`public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
    {
        dump('HELLO WORLD!');
    }`

src/Pelagos/Event/LoginEventListener.php
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @param InteractiveLoginEvent $event The event for successful login.
+     *
+     * @return void
+     */
+    public function onSecurityInteractiveLogin(InteractiveLoginEvent $event)
+    {
+        if ($event->getAuthenticationToken()->isAuthenticated()) {
+            $user = $event->getAuthenticationToken()->getUser();
+            setcookie('GRIIDC_USERNAME', $user->getUsername(), 0, '/');
```
mickel1138: Why not use: `Symfony\Component\HttpFoundation\Cookie` ? The symfony way of setting cookies...

app/config/security.yml
mickel1138: Turns out success_handler is called before logout handler, so no need to change it.

PR #303, opened 2019-06-26
Removed updateDoi method call inside doi issue method

praneethpr:


src/Pelagos/Bundle/AppBundle/Rabbit/Consumer/DoiConsumer.php
```diff
@@ -137,42 +137,41 @@ protected function issueDoi(Dataset $dataset, array $loggingContext)
...
-                $dataset->getTitle(),
-                $dataset->getReferenceDateYear(),
-                'Harte Research Institute'
-            );
-
-            $doi = new DOI($generatedDOI);
-            $doi->setCreator($dataset->getModifier());
-            $doi->setModifier($dataset->getModifier());
-            $dataset->setDoi($doi);
+            $this->logger->info('DOI Already issued for this dataset', $loggingContext);
```
mickel1138: Shouldn't we at least do an update for the DOI?

praneethpr: the last change what i made was exactly that, which led to this bug

src/Pelagos/Bundle/AppBundle/Rabbit/Consumer/DoiConsumer.php
```diff
@@ -137,42 +137,41 @@ protected function issueDoi(Dataset $dataset, array $loggingContext)
...
+                $issueMsg = ConsumerInterface::MSG_REJECT;
+            } catch (HttpServerErrorException $exception) {
+                $this->logger->error('Error requesting DOI: ' . $exception->getMessage(), $loggingContext);
+                //server down. wait for 10 minutes and retry.
+                sleep(self::DELAY_TIME);
+                $issueMsg = ConsumerInterface::MSG_REJECT_REQUEUE;
+            }
         }
-
+
```
mickel1138: extra spaces here.

praneethpr: I'm unable to fix it on phpstorm, if i remove the spaces, the line gets deleted

mickel1138:
I just unlocked an existing DIF, and submitted and approved it again. It issued another DOI.
But the doi issued was a test doi. I'm not sure if we check in our system if the DOI exists. Also this could just be a testing issue.

mickel1138:
Ok, did it another time, and issued another one, again. So doesn't look like a test environment issue.

mickel1138:
So it was my fault, I didn't refresh the consumer. Works as advertised.

src/Pelagos/Bundle/AppBundle/Rabbit/Consumer/DoiConsumer.php
mickel1138: vim?

src/Pelagos/Bundle/AppBundle/Rabbit/Consumer/DoiConsumer.php
mickel1138: We run update on a separate request after issue I think anyway. So that should work.

PR #305, opened 2019-06-27
Reverted back the change made to generate urls in doi consumer

praneethpr:


src/Pelagos/Bundle/AppBundle/Rabbit/Consumer/DoiConsumer.php
```diff
@@ -146,7 +146,7 @@ protected function issueDoi(Dataset $dataset, array $loggingContext)
             try {
                 $doiUtil->createDOI(
                     $generatedDOI,
-                    $this->generateUrl('pelagos_homepage') . '/tombstone/' . $dataset->getUdi(),
+                    'https://data.gulfresearchinitiative.org/tombstone/' . $dataset->getUdi(),
```
mickel1138: Yes, this is more data than a link. So in this case it's fine that it's hard coded.

PR #314, opened 2019-07-03
Changed regex to be case insensitive and accept period instead of col…

praneethpr:
…on in Udi

### DIFF
--- a/src/Pelagos/Util/DatasetIndex.php
+++ b/src/Pelagos/Util/DatasetIndex.php
@@ -82,7 +82,7 @@ class DatasetIndex
                 $text = trim(preg_replace($udiRegEx, '', $text));
                 foreach ($matches[1] as $udi) {
                     // Replacing the 11th position to ":"
-                    $udi = substr_replace($udi, ':',11, 1);
+                    $udi = substr_replace($udi, ':', 11, 1);
                     $textQuery->addShould(
                         new Query\MatchPhrase('udi', $udi)
                     );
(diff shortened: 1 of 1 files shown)


src/Pelagos/Util/DatasetIndex.php
```diff
@@ -77,10 +77,12 @@ protected function buildQuery(array $termsFilters = array(), $text = null, $geoF
 
         if (!empty($text)) {
             $textQuery = new Query\BoolQuery();
-            $udiRegEx = '/\b([A-Z\d]{2}\.x\d\d\d\.\d\d\d:\d\d\d\d)\b/';
+            $udiRegEx = '/\b([a-zA-Z\d]{2}\.x\d\d\d\.\d\d\d[:.]\d\d\d\d)\b/';
```
mickel1138: the 'x' is not  case sensitive. Recommend using /i in regex to make it case insensitive.

praneethpr: Made that change

src/Pelagos/Util/DatasetIndex.php
```diff
@@ -77,10 +77,12 @@ protected function buildQuery(array $termsFilters = array(), $text = null, $geoF
 
         if (!empty($text)) {
             $textQuery = new Query\BoolQuery();
-            $udiRegEx = '/\b([A-Z\d]{2}\.x\d\d\d\.\d\d\d:\d\d\d\d)\b/';
+            $udiRegEx = '/\b([A-Z\d]{2}\.x\d\d\d\.\d\d\d[:.]\d\d\d\d)\b/i';
```
mickel1138: i like that [:.] addition.

PR #315, opened 2019-07-09
Added field selector option to query on specific field

praneethpr:


mickel1138:
Everything works OK, but am unable to select -- All -- again once I selected something else?

praneethpr:
> Everything works OK, but am unable to select -- All -- again once I selected something else?

Fixed it

src/Pelagos/Bundle/AppBundle/Resources/views/Search/default.html.twig
```diff
@@ -1,6 +1,8 @@
 {% extends "PelagosAppBundle::layout.html.twig" %}
 
-{% block head %}
+{%block stylesheets %}
+    {{ parent() }}
+
     {{ add_css (
```
mickel1138: Maybe remove these add_css functions?

PR #317, opened 2019-07-10
Added total file size to stats.

mickel1138:


src/Pelagos/Entity/DatasetRepository.php
```diff
@@ -25,6 +25,22 @@ public function countRegistered()
...
 
+    /**
+     * Sum of all dataset file sizes.
+     *
+     * @return integer Size of data in bytes.
+     */
+    public function totalDatasetSize()
+    {
+        return $this->createQueryBuilder('dataset')
+            ->select('SUM(COALESCE(datasetSubmission.datasetFileColdStorageArchiveSize,datasetSubmission.datasetFileSize))')
```
praneethpr: I dont think sum is required here

src/Pelagos/Entity/DatasetRepository.php
```diff
@@ -25,6 +25,22 @@ public function countRegistered()
             ->getSingleScalarResult();
     }
 
+    /**
+     * Sum of all dataset file sizes.
+     *
+     * @return integer Size of data in bytes.
+     */
+    public function totalDatasetSize()
```
praneethpr: Return type

PR #319, opened 2019-07-15
Feature/pelagos 3693 clean dl dialog

praneethpr:


mickel1138:
The close icon can just put a plain X, right now it is overlapping the frame. Maybe it needs padding?

praneethpr:
> The close icon can just put a plain X, right now it is overlapping the frame. Maybe it needs padding?

I thought i'll rewrite it into a javascript dialog box, I made an another task for this user story

src/Pelagos/Bundle/AppBundle/Resources/views/Download/download-splash-screen.html.twig
```diff
@@ -67,39 +65,17 @@
...
-                        </div>
-                        <div style="display: table-cell;">
-                            <button class="close_button" title="Cancel this download attempt and close this dialog screen." class="dl_button" style="padding: 0px 0px;">
-                                {% image '@PelagosAppBundle/Resources/public/images/cancel.png' %}
-                                    <img style="vertical-align:middle;" src="{{ asset_url }}" border="0">
-                                {% endimage %}
-                            </button>
-                            Cancel Download
-                        </div>
+                        <a id="download-link"> Please click here to Download</a>
```
mickel1138: Maybe add a default HREF so the link will always be styled.

PR #322, opened 2019-07-22
Feature/pelagos 3704 add dataset status facet

praneethpr:


src/Pelagos/Util/Search.php
```diff
@@ -235,13 +265,85 @@ private function getFundingOrgInfo(array $aggregations): array
...
+            if (array_key_exists($status, $aggregations)) {
+                return $aggregations[$status];
+            }
+        };
+
+        $statusInfo = [
+            [
+                'id' => 1,
+                'name' => 'Unavailable & Expected Soon',
+                'count' => $datasetCount(DatasetSubmission::AVAILABILITY_STATUS_NOT_AVAILABLE)
```
mickel1138: This count sometimes return null, and not zero. That looks odd on the search page.

src/Pelagos/Bundle/AppBundle/Resources/public/js/search.js
```diff
@@ -4,11 +4,12 @@ $(document).ready(function () {
     "use strict";
     let pageSize = 10;
     let count = $("#count").attr("data-content");
-    let urlParts = window.location.search.split("?");
-    let queryParse = parseQueryString(urlParts[1]);
-    let startPage = `${queryParse.page ? `${queryParse.page}` : 1}`;
-    let rgId = `${queryParse.resGrp}`;
-    let foId = `${queryParse.fundOrg}`;
+    const parsed = queryString.parse(location.search);
+    console.log(parsed);
```
mickel1138: debug

src/Pelagos/Util/Search.php
praneethpr: Good catch, returning 0 in the else case now 👍

src/Pelagos/Bundle/AppBundle/Resources/public/js/search.js
praneethpr: Removed it

PR #326, opened 2019-07-30
Feature/pelagos 3678 remove assetic from composer and app

mickel1138:
Merge PELAGOS-3676 first!


mickel1138:
> I think updating the composer file, should be a different pull request. This should be just removing only the assetic bundle and its dependencies.
> `composer remove symfony/assetic-bundle`
> just removes the bundle without updating other libs

I can just remove the lock file?

PR #329, opened 2019-07-31
Changed status names and fixed unchecking of checkbox functionality

praneethpr:


praneethpr:
Need to merge PELAGOS-3676

praneethpr:
> Maybe odd to think of it now, but shouldn't status be a radio box, instead of checkbox? Either way, works now.

Hmm.. I don't know. We can ask the POs if they want it to be a radio button

PR #330, opened 2019-07-31
Feature/pelagos 3710 change datepicker

praneethpr:


praneethpr:
Need to merge PELAGOS-3676

fightingtexasaggie:
I'm not totally sure if this will be a problem, but sending a date in like 2012-12-01 works fine, but 2012-12-1 results in an exception being thrown by elastic.

praneethpr:
> I'm not totally sure if this will be a problem, but sending a date in like 2012-12-01 works fine, but 2012-12-1 results in an exception being thrown by elastic.

I tried it, the client-side validation shouldn't let you submit the form. I think you were able to get past because the JS libraries weren't found. yarn encore dev should solve the issue

src/Pelagos/Bundle/AppBundle/Resources/public/js/search.js
```diff
@@ -112,6 +113,54 @@ $(document).ready(function () {
         });
     });
 
+    // set up DatePickers
+    $("#collectionStartDate").datepicker({
+        //defaultDate: "",
```
mickel1138: comments.

src/Pelagos/Bundle/AppBundle/Resources/public/js/search.js
```diff
@@ -112,6 +113,54 @@ $(document).ready(function () {
...
+            collectionStartDate: "Collection Start Date is not a valid ISO date",
+            collectionEndDate: "Collection End Date is not a valid ISO date",
+        },
+        ignore: ".ignore,.prototype",
+        submitHandler: function (form) {
+            if ($(".ignore").valid()) {
+                form.submit();
+            }
+        }
+
```
mickel1138: extra line?

PR #331, opened 2019-08-02
Remove assetic and old router from production deploy script.

fightingtexasaggie:
Removing the assetic and old router dumps from bin/deploy

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -11,14 +11,10 @@ bin/console rabbitmq-supervisor:control stop
 
 # Run composer install to make sure all non-dev dependencies are installed and parameters are defined
 composer install --no-dev --optimize-autoloader
-echo
 
-# Dump assetic assets
-bin/console assetic:dump
-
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -11,14 +11,10 @@ bin/console rabbitmq-supervisor:control stop
...
 # Run composer install to make sure all non-dev dependencies are installed and parameters are defined
 composer install --no-dev --optimize-autoloader
-echo
 
-# Dump assetic assets
-bin/console assetic:dump
-
-# Dump FOSJsRouting routes
-echo "Dumping FOSJsRouting routes..."
-bin/console fos:js-routing:dump --target=web/js/routing
```
mickel1138: We ARE still using FOSJsRouting...

fightingtexasaggie: Ok, added it back. I also confirmed running this overwrites the previous routing file, so there's no need to clear it first.

PR #333, opened 2019-08-05
Upgraded jQuery to 3.4+

mickel1138:


mickel1138:
.toggle is on the deprecated list, but that's for the toggle event handler, not the toggle effect which is what we are using.

src/Pelagos/Bundle/AppBundle/Resources/public/js/download.js
```diff
@@ -8,12 +8,14 @@ function startDownload(id)
         return;
     }
 
-    $('<div id="download_splash" />').appendTo("#pelagos-content").load(
```
mickel1138: I didn't even have to remove .load, because only the .load event is deprecated, and the load function.

PR #334, opened 2019-08-05
Feature/pelagos 3716 drupal edit on admin menu

fightingtexasaggie:
The even/odd parity was already off..fixed and added admin link for better usability.


praneethpr:
Do we need to also add these links in the main drupal page ?

PR #335, opened 2019-08-05
Feature/pelagos 3702 improve ui design

praneethpr:


src/Pelagos/Bundle/AppBundle/Resources/public/js/download.js
```diff
@@ -4,42 +4,104 @@ $(document).ready(function()
...
-        return;
-    }
-
     $('<div id="download_splash" />').appendTo("#pelagos-content").load(
-        Routing.generate("pelagos_app_download_default", {"id": id}),
-        function () {
-            initializeDownload(id);
-        }
+        $.getJSON( Routing.generate("pelagos_app_download_default", {"id": id}), function( data ) {
+
```
mickel1138: instead of doing all that stuff in .done, you can just do it in your called back. Success is assumed.

mickel1138:
When I click download, I get a login dialog?

mickel1138:
> When I click download, I get a login dialog?

I forgot to pull, so not a problem.

src/Pelagos/Bundle/AppBundle/Resources/public/js/download.js
praneethpr: For the html div to be populated, i need the ajax request to process before the dialog box is created

src/Pelagos/Bundle/AppBundle/Resources/public/js/download.js
```diff
@@ -4,42 +4,129 @@ $(document).ready(function()
...
+                                }).done(function (data) {
+                                    window.location = `${data.downloadUrl}`;
+                                });
+                            }
+                        }}),
+                    $.extend({}, vex.dialog.buttons.NO, { text: "Cancel" })
+                ]
+            });
+            if (data.remotelyHosted) {
+                $(":button[type='submit']").remove();
```
mickel1138: Maybe it is wise to be a little more specific on the button if possible, or else any additional submit buttons will also be removed.

src/Pelagos/Bundle/AppBundle/Controller/DownloadController.php
```diff
@@ -115,4 +115,24 @@ public function httpAction($id)
...
+     * Get the dataset required information for the Download dialog box.
+     *
+     * @param Dataset $dataset A dataset instance.
+     *
+     * @return array
+     */
+    private function getDatasetDetails(Dataset $dataset): array
+    {
+        $datasetSubmission = $dataset->getDatasetSubmission();
+        return array(
```
fightingtexasaggie: What happens if the submission is null, as would be the case in a DIF only?

mickel1138: > What happens if the submission is null, as would be the case in a DIF only?

Is should probably get caught, but there is nothing to download with a DIF, so this function won't get called. The download button is grayed out.

praneethpr: i'll add a check if it is datasetSubmission object, just to be sure

src/Pelagos/Bundle/AppBundle/Controller/DownloadController.php
```diff
@@ -115,4 +115,24 @@ public function httpAction($id)
...
+            'udi' => $dataset->getUdi(),
+            'filename' => $datasetSubmission->getDatasetFileName(),
+            'fileSize' => TwigExtentions::formatBytes($datasetSubmission->getDatasetFileSize(), 2),
```
fightingtexasaggie: somehow fileSize isn't being populated.

praneethpr: Maybe it is a remotely hosted dataset?

PR #340, opened 2019-08-09
Removed details and size from data discovery page

praneethpr:


fightingtexasaggie:
I added a commit that alternates row colors, but I don't really care for my choice of color. Feel free to change the color in data-discovery.css to something better, (or revert my commit).

mickel1138:
Plain css striped is neater: https://www.w3schools.com/css/tryit.asp?filename=trycss_table_striped

src/Pelagos/Bundle/AppBundle/Resources/public/css/data-discovery.css
```diff
@@ -140,13 +140,15 @@ input.ui-button {
     border-top-right-radius: 4px;
 }
 
+.datasets tr:nth-child(odd) {
```
mickel1138: Odd or Even?

fightingtexasaggie: Coloring the odds so the first row gets colored

PR #343, opened 2019-08-12
refactor/nfsv4 acl compat

fightingtexasaggie:
To test:

The NetApp is already connected to griidc-dev, so this branch needs to be merged into anything else you're working on if you need files to work.

1. create/approve new dif, then submit a dataset with a small file to test.
2. Approve the submission, then download it from dataland.
3. Repeat for http pull and sftp uploaded file.

Check permissions on:
/san/data/store/FOLDER
/san/data/store/FOLDER/DATAFILE
/san/data/download/FOLDER
/san/data/download/FOLDER/DATAFILE

In order to see the NFS ACL permissions, use the nfs4_getfacl command.
pertinent info: 48 = UID of 'apache' user, 1008 = GID of 'griidc' ldap group.
r=read, x=execute (not traversal as with UNIX), t = traversal


fightingtexasaggie:
Let's keep this out of the Develop branch until we have a deployment strategy figured out.

src/Pelagos/Util/DataStore.php
```diff
@@ -424,4 +410,26 @@ protected function getStoreFileName($datasetId, $type)
...
+     * @param boolean $isGroup Flag set to true if a group, defaults to user.
+     *
+     * @return string|boolean
+     */
+    protected function getIdFromName(string $name, bool $isGroup = false)
+    {
+        $id = false;
+        if ($isGroup) {
+            // This call returns false on failure to resolve.
+            $obj = posix_getgrnam($name);
```
praneethpr: Is ext-posix required to be added to composer.json?

src/Pelagos/Util/DataStore.php
```diff
@@ -424,4 +410,26 @@ protected function getStoreFileName($datasetId, $type)
...
+            $id = $obj['gid'];
+        } else {
+            $obj = posix_getpwnam($name);
+            $id = $obj['uid'];
+        }
+        return $id;
```
praneethpr: Can id also have a string value?

src/Pelagos/Util/DataStore.php
fightingtexasaggie: No, this is the built-in PHP function.

src/Pelagos/Util/DataStore.php
fightingtexasaggie: no, just either a numeric value or a boolean false as posix_getpwnam and the other return false on failure. Come to think of it, it's probably better I throw an exception if that were to ever happen.

praneethpr: Oh then, we should probably check if array exists before trying to access the array $obj

app/config/config_drupal_dev.yml
```diff
@@ -10,6 +10,9 @@ framework:
         strict_requirements: true
     profiler: { only_exceptions: false }
 
+fos_js_routing:
+    request_context_base_url: /pelagos-symfony
```
mickel1138: I thought this was already in develop?

mickel1138: If you need it for _your_ development enviroment, add .local to that yaml file.

PR #347, opened 2019-08-14
Added Fileset entity

praneethpr:


src/Pelagos/Entity/Fileset.php
```diff
@@ -0,0 +1,75 @@
...
+    private $numberOfFiles;
+
+    /**
+     * Set totalFileSize.
+     *
+     * @param integer $totalFileSize Total file size in bytes.
+     *
+     * @return void
+     */
+    public function setTotalFileSize(int $totalFileSize)
```
mickel1138: While these properties are cool, are may we used, there is no requirement for them now. And if we are just adding them because we might use them, then you should add something like a hash too.

Another thing is that file size and number should probably be just calculated automatically, and not be able to set it, so we don't need the setters. And then, the getters are also not needed (just yet).

I don't mind if we leave this in for now, but it doesn't serve a purpose, or is needed as part as the requirements.

src/Pelagos/Entity/Fileset.php
```diff
@@ -0,0 +1,75 @@
...
+namespace Pelagos\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * Fileset Entity class.
+ *
+ * @ORM\Entity
+ */
+class Fileset extends Entity
```
mickel1138: Have we consider the reach back? Meaning, do you want to know which dataset submission a fileset goes with if you one have the fileset entity/id?

mickel1138:
I wonder how this entity would have looked if we used the doctine entity generator?
`doctrine:generate:entity`

src/Pelagos/Entity/Fileset.php
praneethpr: I think the bidirectional relationship could be useful for the Fileset entity and File entity, but i'm not sure if we need it for this one

src/Pelagos/Entity/Fileset.php
praneethpr: Yeah I agree that these attributes are not part of the requirements, but just having this entity solely to have a relationship with Files entity, we can just have Files entity instead

mickel1138: It's for all the things we -don't- know! ;-)

mickel1138:
Everything looks good, but not have an inverse relationship means that we can not find out from a fileset which dataset it belongs to. Maybe that's the way it should work?

src/Pelagos/Entity/DatasetSubmission.php
```diff
@@ -2861,4 +2870,26 @@ public function setRemotelyHostedFunction($remotelyHostedFunction)
...
+    }
+
+    /**
+     * Setter for fileset entity.
+     *
+     * @param Fileset $fileSet The fileset entity associated with this datasetSubmission instance.
+     *
+     * @return void
+     */
+    public function setFileSet(Fileset $fileSet) : void
```
mickel1138: I'd like to see this function go away when the files entity comes along. Instead there's a function maybe called addFile(s), and removeFile(s) which makes it's own fileset. We may have to use a managing utility for this?

PR #349, opened 2019-08-14
Stop using async on researchgroup.

mickel1138:


src/Pelagos/Bundle/AppBundle/Resources/public/js/entity/ResearchGroup.js
```diff
@@ -91,28 +93,25 @@ $(document).ready(function()
...
+            $.each(entities, function(seq, item) {
+                selectElement.append(
+                    $("<option></option>").val(item.id).html(item.name)
+                );
+            });
+            //selectElement.find("option[value=\"" + selectElement.attr(entity) + "\"]").attr("selected", true);
         });
-        //selectElement.find("option[value=\"" + selectElement.attr(entity) + "\"]").attr("selected", true);
+        
+        return this.resolve();
```
fightingtexasaggie: instead of this, it seems to work fine with chaining .promise() to the end of the outer return. Why the 2 returns?

mickel1138: Second return wasn't needed, removed it.

PR #350, opened 2019-08-19
Added populate research group short names command

praneethpr:


mickel1138:
One could use the postgres COPY command:
https://infinum.co/the-capsized-eight/superfast-csv-imports-using-postgresqls-copy

PR #351, opened 2019-08-19
Feature/pelagos 3740 add doi and clean up view

fightingtexasaggie:


src/Pelagos/Bundle/AppBundle/Resources/views/DatasetMonitoring/datasets.html.twig
```diff
@@ -137,7 +138,12 @@
 
             <tr udi="{{dataset.udi}}">
                 {% if not dataland | default %}
-                <td class="udi details"><a href="{{ path('pelagos_app_ui_dataland_default', { 'udi': dataset.udi }) }}" target="_blank">{{dataset.udi}}</a></td>
+                <td class="details"><a href="{{ path('pelagos_app_ui_dataland_default', { 'udi': dataset.udi }) }}" target="_blank">{{dataset.udi}}</a></td>
+                {% if dataset.datasetSubmissionStatus == constant('Pelagos\\Entity\\DatasetSubmission::STATUS_COMPLETE') or dataset.datasetSubmissionStatus == constant('Pelagos\\Entity\\DatasetSubmission::STATUS_IN_REVIEW') %}
+                <td class="details"><a href="https://doi.org/{{ dataset.doi | replace({'doi:': "", })}}">{{ dataset.doi | replace({'doi:': "", }) }}</a></td>
```
praneethpr: I don't think we store the doi with the doi prefix, so we can get rid of the replace

fightingtexasaggie: The DOI entity's __toString() method appends it.

fightingtexasaggie: removed the replace() and am using doi.doi to avoid the __toString() madness

PR #353, opened 2019-08-19
Feature/pelagos 3738 add locked attribute to research group entity

fightingtexasaggie:


src/Pelagos/Entity/ResearchGroup.php
```diff
@@ -652,13 +661,39 @@ public function setEmailAddress($emailAddress)
...
+    /**
+     * Setter for lockStatus.
+     *
+     * @param boolean $state Containing desired state of lock.
+     *
+     * @access public
+     *
+     * @return void
+     */
+    public function setLockStatus(bool $state)
```
mickel1138: Maybe a bit odd, but I like the functions: lock, unlock and isLocked. That way we have a bit more control. For instance when unlocked making sure it's unlockable, and not already unlocked, or something in the future that prevents locking.

src/Pelagos/Entity/ResearchGroup.php
```diff
@@ -293,6 +293,15 @@ class ResearchGroup extends Entity
...
     protected $datasets;
 
+    /**
+     * Lock flag for Research Group.
+     *
+     * @var boolean $lockStatus
+     *
+     * @ORM\Column(type="boolean", nullable=false)
+     */
+    protected $lockStatus = false;
```
mickel1138: Status in lockStatus but be a bit ambiguous. Maybe we should just call it 'locked' ?

src/Pelagos/Entity/ResearchGroup.php
```diff
@@ -652,13 +661,49 @@ public function setEmailAddress($emailAddress)
...
+    }
+
+    /**
+     * Method to check if Research Group is locked.
+     *
+     * @access public
+     *
+     * @return boolean Set to true if this Research Group is locked, false otherwise.
+     */
+    public function isResearchGroupLocked()
```
mickel1138: Is that extra "researchGroup" on these functions really needed? This is how it might look:
`if $researchGroup->isResearchGroupLocked()`
This is a long standing debate, but I think just isLocked is sufficient.

PR #358, opened 2019-08-21
Added files entity

praneethpr:


src/Pelagos/Entity/Fileset.php
```diff
@@ -11,5 +12,36 @@
...
+    }
+
+    /**
+     * Setter for files.
+     *
+     * @param Collection $files The collection of files in this fileset.
+     *
+     * @return void
+     */
+    public function setFiles(Collection $files) : void
```
mickel1138: In additional, or maybe instead of this, I would add functions: addFile, removeFile that need the required fields to make a file a file. Or else you can just add files, without any attributes.

praneethpr: Rename it to addFiles instead?

mickel1138: I was thinking adding one file at the time, but there could be a add multiple files one too?

praneethpr: I thought we can create all the files and then associate the collection to the Fileset entity

mickel1138: I think this is ok for now, things might change when we get further along.

PR #367, opened 2019-09-03
Added contraint to voter.

mickel1138:


mickel1138:
> If we move this up ahead of the DRPM check, this block will prevent DRPM actions as well. It's my understanding that "locked" applies to everyone and although DRPMs can unlock, will suffice as an additional safety net.

I will have to double check with the PO on that.

PR #374, opened 2019-09-12
Bugfix/pelagos 3767 add expiration check to pelagos authenticator

fightingtexasaggie:
To test, connect to database and find your person_id from the account table, then update the latest password modification date by that username so that it is more than 6 months ago. Try logging in. Set it back. Try logging in again.

ex: 
SQL> SELECT account_id FROM account WHERE user_id = 'mwilliamson';
10009

SQL> \x
SQL>SELECT * FROM password WHERE account_id = 10009;
...lots of records...use highest ID one.
id                      | 1917
...
modification_time_stamp | 2019-05-06 15:45:21-05

SQL>UPDATE password SET modification_time_stamp = '2019-03-06 15:45:21-05' WHERE id = 1917;

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -122,6 +122,8 @@ services:
     pelagos.security.login_form_authenticator:
         class: Pelagos\Bundle\AppBundle\Security\LoginFormAuthenticator
         autowire: true
+        arguments:
+            $maxPwAge: "%account_password_max_age%"
         tags:
             - { name: monolog.logger, channel: login_attempts }
 
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -123,7 +123,7 @@ services:
         class: Pelagos\Bundle\AppBundle\Security\LoginFormAuthenticator
         autowire: true
         arguments:
-            $maxPwAge: "%account_password_max_age%"
+            $maximumPasswordAge: "%account_password_max_age%"
         tags:
             - { name: monolog.logger, channel: login_attempts }
 
--- a/src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
+++ b/src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
@@ -78,7 +78,7 @@ class LoginFormAuthenticator extends AbstractFormLoginAuthenticator
      * @param EntityManagerInterface $entityManager An Entity Manager.
      * @param RouterInterface        $router        A Router.
      * @param LoggerInterface        $logger        A Monolog logger.
-     * @param string                 $maxPwAge      The max age for password paramater.
+     * @param string                 $maxPwAge      The max age for password parameter.
      */
     public function __construct(
         FormFactoryInterface $formFactory,
--- a/src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
+++ b/src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
@@ -264,8 +264,10 @@ class LoginFormAuthenticator extends AbstractFormLoginAuthenticator
     /**
      * Return to login page and add destination for redirect.
      *
-     * @param Request                 $request   A Symfony Request, req by interface.
-     * @param AuthenticationException $exception The exception thrown.
+     * @param Request                  $request   A Symfony Request, req by interface.
+     * @param AuthenticationException  $exception The exception thrown.
+     *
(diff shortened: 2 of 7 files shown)


src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
```diff
@@ -63,24 +64,34 @@ class LoginFormAuthenticator extends AbstractFormLoginAuthenticator
      */
     protected $logger;
 
+    /**
+     * String describing max PW age.
+     *
+     * @var string maxPwAge
+     */
+    protected $maxPwAge;
```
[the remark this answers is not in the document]
fightingtexasaggie: Perhaps because my first language was FORTRAN. j/k. I'll use better names.

src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
```diff
@@ -161,6 +173,14 @@ public function checkCredentials($credentials, UserInterface $user)
             throw new AuthenticationException('Too many login attempts');
         }
 
+        // Check for expired password.
+        $now = new \DateTime('now');
+        $expiration = $user->getPasswordEntity()->getModificationTimeStamp()->add(new \DateInterval($this->maxPwAge));
```
mickel1138: What if maxPwAge is blank/null?

app/config/services.yml
```diff
@@ -122,6 +122,8 @@ services:
     pelagos.security.login_form_authenticator:
         class: Pelagos\Bundle\AppBundle\Security\LoginFormAuthenticator
         autowire: true
+        arguments:
+            $maximumPasswordAge: "%account_password_max_age%"
```
mickel1138: Also add this to parameters dist file.

src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
fightingtexasaggie: Oh yeah, I want that to be the case for never expiring, blank or zero. Good catch.

fightingtexasaggie:
I've reviewed Mike's changes, but can't approve as I'm a contributer.

mickel1138:
Instead of the SQL thing, you can also set you max_password_age to P1D, then it expires in 1 day.

src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
```diff
@@ -63,24 +64,34 @@ class LoginFormAuthenticator extends AbstractFormLoginAuthenticator
...
+     * @param LoggerInterface        $logger             A Monolog logger.
+     * @param string|null            $maximumPasswordAge The max age for password, 0 or null means never expires.
      */
     public function __construct(
         FormFactoryInterface $formFactory,
         EntityManagerInterface $entityManager,
         RouterInterface $router,
-        LoggerInterface $logger
+        LoggerInterface $logger,
+        $maximumPasswordAge
```
mickel1138: You can use `? string` here i think.

src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
```diff
@@ -227,16 +267,51 @@ public function onAuthenticationSuccess(Request $request, TokenInterface $token,
      * @param Request                 $request   A Symfony Request, req by interface.
      * @param AuthenticationException $exception The exception thrown.
      *
+     * @throws AuthenticationException When the user is not found, and credentials is invalid.
+     *
      * @return Response The response or null to continue request.
      */
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
```
mickel1138: If we really wanted use a different exception for password expired, then `AuthenticationException` hint can be changed to `Exception` hint.

src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
praneethpr: We usually do something like this right?
`string $maximumPasswordAge=null`

src/Pelagos/Bundle/AppBundle/Security/LoginFormAuthenticator.php
mickel1138: NM, this won't work, the interface doesn't like that.

PR #376, opened 2019-09-17
Feature/pelagos 3750 modify datasetsubmission to use new fileset files entities

praneethpr:


src/Pelagos/Entity/File.php
```diff
@@ -79,20 +79,10 @@ class File extends Entity
...
-    /**
-     * File constructor.
-     *
-     * @param Fileset $fileSet Fileset entity which this file belongs to.
-     */
-    public function __construct(Fileset $fileSet)
-    {
-        $this->fileSet = $fileSet;
-    }
+    protected $fileset;
```
mickel1138: Need docblock and relationship?

praneethpr: I didn't understand what you meant

fightingtexasaggie:
Can you merge in a fresh develop? There are some conflicts that will have to be resolved in merging, but it's needed as the NFSv4 stuff is now needed for filer to run.

PR #379, opened 2019-09-17
Added boost to doi

praneethpr:


src/Pelagos/Util/DatasetIndex.php
```diff
@@ -92,21 +92,26 @@ protected function buildQuery(array $termsFilters = array(), $text = null, $geoF
...
                     $doiQuery = new Query\Nested();
                     $doiQuery->setPath('doi');
                     $doiQuery->setQuery(
                         new Query\MatchPhrase('doi.doi', $doi)
                     );
 
+                    $doiBoolQuery->addMust($doiQuery);
+                    $doiBoolQuery->setBoost(3.0);
+
+
```
mickel1138: Extra line?

src/Pelagos/Util/DatasetIndex.php
```diff
@@ -92,21 +92,26 @@ protected function buildQuery(array $termsFilters = array(), $text = null, $geoF
...
+                    $doiBoolQuery = new Query\BoolQuery();
                     // Query against dataset DOIs.
                     $doiQuery = new Query\Nested();
                     $doiQuery->setPath('doi');
                     $doiQuery->setQuery(
                         new Query\MatchPhrase('doi.doi', $doi)
                     );
 
+                    $doiBoolQuery->addMust($doiQuery);
+                    $doiBoolQuery->setBoost(3.0);
```
mickel1138: Why does the boost work here? Have tried it on the config file, but didn't work.

fightingtexasaggie: safer here than the config as that's global anyway imho

fightingtexasaggie: Try searhing for 10.7266/n7-fjhy-6w15 and you'll see it sorts 3rd instead of first. Are you sure the boost is working?

fightingtexasaggie: I had a bad local testing environment. I can see it's working now.

praneethpr: Index time boosting is deprecated as of 5.0.0, probably it is removed in the current version which we are using

PR #388, opened 2019-09-23
Fixed loaddump and added .idea folder to git ignore

praneethpr:


.gitignore
```diff
@@ -23,3 +23,5 @@ yarn-error.log
 .phpunit.result.cache
 /phpunit.xml
 ###< symfony/phpunit-bridge ###
+
+.idea
```
mickel1138: need new line

praneethpr: I think we will add more stuff to it

PR #396, opened 2019-09-27
Account Application - PELAGOS-3782

mickel1138:
Intentionally left commented out code in EventListener.php this construct does too much, which was convenient when we had to manually register every service, but now it's Autowired, and each service even if it's part of a base class, should ask for what they need at the appropriate time. So the commented out stuff is for reference, when other Listeners get added.
Yes, a listener still had to be registered in services,yml, we could refactor to EventSubscribers, but the mechanism (almost like a voter) is a bit different, and a refactor.


.env
```diff
@@ -43,5 +43,4 @@ HOSTNAME=host-env.tamucc.edu
 GOOGLE_MAPS_API_KEY=SECRET
 GOOGLE_ANALYTICS_ID=ID
 PELAGOS_READONLY_MESSAGE=MESSAGE
-MAINTENANCE_MODE_FILENAME="/var/maintenance.ini"
-ACCOUNT_PASSWORD_MAX_AGE=P180D
```
praneethpr: Did you revert the changes for this file by mistake?

templates/template/ErrorMessage.html.twig
```diff
@@ -0,0 +1,9 @@
+{% extends "PelagosAppBundle:template:UI.html.twig" %}
```
praneethpr: Change the file path

src/Controller/UI/AccountController.php
```diff
@@ -0,0 +1,514 @@
...
+    public function __construct(EntityHandler $entityHandler, ValidatorInterface $validator)
+    {
+        $this->entityHandler = $entityHandler;
+        $this->validator = $validator;
+    }
+
+    /**
+     * The index action.
+     *
+     * @Route("/account", methods={"GET"}, name="pelagos_app_ui_account_default")
```
praneethpr: I don't think we need to continue using the old names for the UI routes

src/Controller/UI/AccountController.php
```diff
@@ -0,0 +1,514 @@
...
+        $password = new Password($request->request->get('password'));
+
+        // Set the creator for password.
+        $password->setCreator($person);
+
+        // Attach the password to the account.
+        try {
+            $account->setPassword(
+                $password,
+                ((bool) ($this->container->hasParameter('account_less_strict_password_rules')) and
```
praneethpr: Can remove container

src/Controller/UI/AccountController.php
```diff
@@ -0,0 +1,514 @@
...
+        }
+
+        // Validate both Password and Account
+        $this->validateEntity($password, $validator);
+        $this->validateEntity($account, $validator);
+
+        $account = $this->entityHandler->update($account);
+
+        // Update LDAP
+        $this->get('pelagos.ldap')->updatePerson($person);
```
praneethpr: Here

src/Controller/UI/AccountController.php
```diff
@@ -0,0 +1,514 @@
...
+
+            // Persist Account
+            $account = $this->entityHandler->create($account);
+
+            try {
+                // Try to add the person to LDAP.
+                $this->get('pelagos.ldap')->addPerson($person);
+            } catch (LdapException $exception) {
+                // If that fails, try to update the person in LDAP.
+                $this->get('pelagos.ldap')->updatePerson($person);
```
praneethpr: Here

src/Controller/UI/AccountController.php
```diff
@@ -0,0 +1,514 @@
...
+            // Validate the entities.
+            $this->validateEntity($password);
+            $this->validateEntity($account);
+
+            // Persist Account
+            $account = $this->entityHandler->create($account);
+
+            try {
+                // Try to add the person to LDAP.
+                $this->get('pelagos.ldap')->addPerson($person);
```
praneethpr: Here

src/Controller/UI/AccountController.php
```diff
@@ -0,0 +1,514 @@
...
+            }
+
+            // Validate the entities.
+            $this->validateEntity($password, $validator);
+            $this->validateEntity($account, $validator);
+
+            // Persist Account
+            $account = $this->entityHandler->update($account);
+
+            $this->get('pelagos.ldap')->updatePerson($person);
```
praneethpr: Inject this as a service? and change usages below

templates/Account/Account.html.twig
```diff
@@ -0,0 +1,41 @@
+{% extends "template/UI.html.twig" %}
+
+{% block javascripts %}
+    {{ parent() }}
+    
+    <script type="text/javascript" src="{{ asset('build/js/Account.js') }}"></script>
```
praneethpr: Should remove build from the path?

src/Controller/UI/AccountController.php
mickel1138: In the future we should change them, i need them for now.

PR #399, opened 2019-10-02
Parse DB parameters from .env.local

fightingtexasaggie:
This is now working, however you'll need to comment out the rabbitmq commands and only use it to load the database, not elasticsearch as those components are not working at this time. This file shouldn't need to change when they are working later though.

### DIFF
--- a/bin/loaddump
+++ b/bin/loaddump
@@ -100,15 +100,10 @@ fi
 bin/console rabbitmq-supervisor:control stop
 
 # Drop existing database
-echo ONE
-echo "SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity WHERE pg_stat_activity.datname = '$parameters__database_name' AND pid <> pg_backend_pid(); $parameters__database_host"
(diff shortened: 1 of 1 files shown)


bin/loaddump
```diff
@@ -38,7 +38,27 @@ if [ ! -f $1 ]; then
     exit
 fi
 
-eval $(parse_yaml app/config/parameters.yml "")
+# Manually parse DATABASE_URL from .env.local file
+
+DATABASE_URL=$(grep "^DATABASE_URL" ./.env.local)
+IFS=':' read -ra ADDR <<< "$DATABASE_URL"
```
mickel1138: What is ADDR?

fightingtexasaggie: ok, I'll make less cryptic

fightingtexasaggie:
> Getting an authentication error, using the same password

Make sure you use localhost and not 127.0.0.1 in the .env.local file.

mickel1138:
@fightingtexasaggie We now have a field called database_name, you can use it if you like.

PR #400, opened 2019-10-03
Feature/pelagos 3785 add dataset monitoring

praneethpr:
Dataset Monitoring and Dataland should work in this pull request


mickel1138:
We fixed the API issues with this info:
https://stackoverflow.com/questions/30114152/fosrestbundle-setup-for-return-json-but-still-asking-for-twig-template

.env
```diff
@@ -52,3 +52,4 @@ LDAP_PEOPLE_OU='ou=users,dc=domain,dc=org'
 ACCOUNT_LESS_STRICT_PASSWORD_RULES=FALSE
 MAILER_FROM_ADDR='email@domain.null'
 MAILER_FROM_NAME='SenderName'
+PELAGOS_NAME=pelagos
```
mickel1138: This is just a weird parameter. Is this different in env.local?

praneethpr: The value is the pelagos name of our instance, which you need to add in the .env.local file
For example:
PELAGOS_NAME=pelagos.ppondicherry

config/routes/annotations.yaml
```diff
@@ -1,3 +1,5 @@
 controllers:
     resource: ../../src/Controller/
     type: annotation
+    options:
+        expose: true
```
mickel1138: This doesn't seem to make any difference? Where does it break if you change this?

config/packages/fos_elastica.yaml
```diff
@@ -3,4 +3,70 @@ fos_elastica:
     clients:
         default: { host: localhost, port: 9200 }
     indexes:
-        app: null
+        pelagos:
```
mickel1138: Should we call this 'app'?

composer.json
```diff
@@ -77,7 +77,8 @@
     "scripts": {
         "auto-scripts": {
             "cache:clear": "symfony-cmd",
-            "assets:install %PUBLIC_DIR%": "symfony-cmd"
+            "assets:install %PUBLIC_DIR%": "symfony-cmd",
+            "fos:js-routing:dump --format=json --target=public/js/fos_js_routes.json": "symfony-cmd"
```
mickel1138: we should use %PUBLIC_DIR% ?

config/routes/annotations.yaml
praneethpr: Try running the fos:js-routing:dump/debug, it should be empty without that config

composer.json
praneethpr: Will add it to the wiki List

config/routes/annotations.yaml
mickel1138: I can't seem to make it fail?

config/packages/fos_elastica.yaml
praneethpr: Will add it to the wiki List

PR #401, opened 2019-10-04
Feature/pelagos 3788 data discovery

praneethpr:
Need to merge dataset- monitoring before


praneethpr:
> I'm seeing a 500 upon trying to use the geo filter. Otherwise everything else seems to be working. The error is get_class() expects parameter 1 to be object, null given in DataDiscoveryController.php

Yes, i've seen that. I guess it is missing the Gml Controller. I'll add this to the list of things on Wiki, for now we can merge this.

PR #403, opened 2019-10-07
Migrated phpcs to sniffs 3

mickel1138:
Fixed up .travis a bit.


mickel1138:
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Version-3.0-Upgrade-Guide#upgrading-custom-sniffs

mickel1138:
PHPCS now working, try yourself:
./vendor/bin/phpcs --colors --standard=coding-standards/GRIIDC --extensions=php -s -n ./src

.env.test
```diff
@@ -3,3 +3,4 @@ KERNEL_CLASS='App\Kernel'
 APP_SECRET='$ecretf0rt3st'
 SYMFONY_DEPRECATIONS_HELPER=999999
 PANTHER_APP_ENV=panther
+DATABASE_URL=postgres://postgres@127.0.0.1:3306/postgres
```
fightingtexasaggie: please use localhost instead of 127.0.0.1. It's a separate db permission.

fightingtexasaggie: 3306 is the default port on mysql. 5432 is the one for postgres

mickel1138: Yes, that should be changed, but we are not creating a DB this time around, hopefully someday soon we can add fixtures and do proper regressions testing against the UI also.

fightingtexasaggie: ok, good point...

.env.test
```diff
@@ -3,3 +3,4 @@ KERNEL_CLASS='App\Kernel'
 APP_SECRET='$ecretf0rt3st'
 SYMFONY_DEPRECATIONS_HELPER=999999
 PANTHER_APP_ENV=panther
+DATABASE_URL=postgres://postgres@127.0.0.1:5432/postgres
```
praneethpr: Can be removed

src/Controller/Api/DataCenterController.php
```diff
@@ -157,7 +157,7 @@ public function getCollectionAction(Request $request)
     /**
      * Get a data center for a given id.
      *
-     * @param string $id The id of the Data center.
+     * @param integer $id The id of the Data center.
```
mickel1138: I changed this from string to integer, with the assumption that an ID always an integer is.

src/Entity/DIF.php
```diff
@@ -635,7 +635,7 @@ public function getAbstract()
      *
      * @return void
      */
-    public function setFieldOfStudyEcologicalBiological($fieldOfStudyEcologicalBiological)
+    public function setFieldOfStudyEcologicalBiological(bool $fieldOfStudyEcologicalBiological)
     {
         if ('boolean' !== gettype($fieldOfStudyEcologicalBiological)) {
```
mickel1138: I guess this won't be needed anymore...

These entities are littered with checks, they should be done by Assert or Assert\Type if needed.

mickel1138:
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file

New ignore syntax.

mickel1138:
There are TWO coding standard issues remaining:
/src/Controller/GmlController.php has a non-nullable parameter at the end the function. (That's a no-no).
/src/Command/MakeDRPMCommand.php need cleanup. @fightingtexasaggie isle one!

src/Controller/Api/DataCenterController.php
mickel1138: This will be a refactor.

src/Entity/DIF.php
mickel1138: Will remove in a later PR.

PR #404, opened 2019-10-08
Feature/pelagos 3792 submission and review

praneethpr:


praneethpr:
> I'm getting the BadRequestHttpException('The DIF has not yet been approved for this dataset.') exception throw as I try and submit a fully-completed form, with direct upload. (checked that dif was approved, and would not have been able to launch this on unapproved dif anyway. I went ahead and verified this was still properly preventing me from loading the form for an unapproved dif.)

It breaks at trying to publish a message, and we don't have the consumers setup yet. So it doesn't change the identified status of the dataset, even though DIF status is approved. Tested it by, commenting out the publish line in the DIFListener, able to submit the dataset. Should we go ahead with rabbitmq or try to move to Symfony messaging

src/Controller/UI/DatasetSubmissionController.php
```diff
@@ -0,0 +1,597 @@
...
+                $this->entityHandler->update($metadataContact);
+            }
+
+            $this->entityEventDispatcher->dispatch(
+                $datasetSubmission,
+                $eventName
+            );
+
+            foreach ($this->messages as $message) {
+                $this->get('old_sound_rabbit_mq.dataset_submission_producer')->publish(
```
mickel1138: Can't get this anymore, use injection in command.

templates/DatasetReview/submit.html.twig
```diff
@@ -0,0 +1,101 @@
...
+                <td>
+                    {{ DatasetSubmission.datasetFileUri }}
+                </td>
+            </tr>
+            <tr>
+                <td>
+                    <b>Submitted On:</b>
+                </td>
+                <td>
+                    {{ DatasetSubmission.getSubmissionTimeStamp.format('c') }}
```
mickel1138: This breaks when submissionTimeStamp is null

PR #407, opened 2019-10-14
Feature/pelagos 3798 add bin localmods

fightingtexasaggie:
added Debora and converted to use .env.local file instead of parameters.yml


bin/localmods
```diff
@@ -1,29 +1,32 @@
...
+
+dbportname=${database_url_parts[3]}
+IFS='/' read -ra partname_array <<< "$dbportname"
+dbport=${partname_array[0]} # unused
+parameters__database_name=${partname_array[1]}
 
 export PGPASSWORD="$parameters__database_password"
 
-echo "update person_data_repository set role_id = 1 where person_id in (select person_id from account where user_id in ('mvandeneijnden', 'mwilliamson', 'rrossi', 'bsigdel', 'wnichols', 'ppondicherry', 'lasirvadam' , 'undu'))" | /usr/bin/psql -h $parameters__database_host -U $parameters__database_user $parameters__database_name || exit
+echo "update person_data_repository set role_id = 1 where person_id in (select person_id from account where user_id in ('mvandeneijnden', 'mwilliamson', 'rrossi', 'bsigdel', 'wnichols', 'ppondicherry', 'lasirvadam' , 'undu', 'dlebel'))" | /usr/bin/psql -h $parameters__database_host -U $parameters__database_user $parameters__database_name || exit
```
mickel1138: Should we configure the userlist in a file, rather than changing this file everything, which is checked into source control.

fightingtexasaggie: How do you feel about an entry in the .env.local for it?

mickel1138: use you own file!

fightingtexasaggie: Honestly, this whole thing should probably be a Symfony command.

bin/localmods
```diff
@@ -1,29 +1,32 @@
...
-if [ ! -f app/config/parameters.yml ]; then
-    echo "parameters.yml not found"
+if [ ! -f .env.local ]; then
+    echo ".env.local not found"
     exit
 fi
+# Manually parse DATABASE_URL from .env.local file
 
-eval $(parse_yaml app/config/parameters.yml "")
+database_url=$(grep "^DATABASE_URL" ./.env.local)
```
mickel1138: There is a new parameter/variable now called: `DATABASE_NAME`

bin/localmods
mickel1138: This can be a refactor, and we can this another time.

src/Command/MakeDRPMCommand.php
```diff
@@ -0,0 +1,94 @@
...
+     * A Doctrine ORM EntityManager instance.
+     *
+     * @var EntityManagerInterface $entityManager
+     */
+    protected $entityManager;
+
+    /**
+     * Class constructor for dependency injection.
+     *
+     * @param EntityManagerInterface $em
```
mickel1138: i reject em

src/Command/MakeDRPMCommand.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    protected function configure()
+    {
+        $this
+            ->setDescription('Promote an existing user into a DRPM.')
+            ->addArgument('username', InputArgument::REQUIRED, 'User to make into DRPM.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
```
mickel1138: needs docblock

PR #408, opened 2019-10-14
Feature/pelagos 3799 add stats

fightingtexasaggie:


src/Controller/UI/StatsController.php
```diff
@@ -0,0 +1,299 @@
+<?php
+
+namespace App\Controller\UI;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Query;
+
+use Pelagos\Bundle\AppBundle\Twig\Extensions as TwigExtentions;
```
praneethpr: Need to change namespace for this

fightingtexasaggie: good catch. I wonder how it's even working, or if it's even needed.

praneethpr: If you goto the /stats/json route, it breaks and that is used in the drupal home page

webpack.config.js
```diff
@@ -35,7 +35,7 @@ Encore
     // will require an extra script tag for runtime.js
     // but, you probably want this, unless you're building a single-page app
     //.enableSingleRuntimeChunk()
-
```
mickel1138: ?

fightingtexasaggie: soft tab got removed. My editor cleans this stuff up automatically.

PR #409, opened 2019-10-15
Feature/pelagos 3801 add mdapp

fightingtexasaggie:
I ended up using the _defaults for the jiral link URL and the mdapp logfile name. It is working like this, but if anyone has ideas to make it work locally, I'm up for any suggestions.  I probably missed something, but it was acting weird not letting me bind to a controller directly.


.env
```diff
@@ -53,4 +53,6 @@ ACCOUNT_LESS_STRICT_PASSWORD_RULES=FALSE
 MAILER_FROM_ADDR='email@domain.null'
 MAILER_FROM_NAME='SenderName'
 PELAGOS_NAME=pelagos
+ISSUE_TRACKING_BASE_URL='https://griidc.atlassian.net/issues'
+MDAPP_LOGFILE='/opt/pelagos/var/logs/mdapp.log'
```
mickel1138: The folder logs doesn't exist anymore, it has been renamed to 'log' (new Symfony thing)

config/services.yaml
```diff
@@ -10,6 +10,9 @@ services:
     _defaults:
         autowire: true      # Automatically injects dependencies in your services.
         autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
+        bind:
```
mickel1138: Moving this to section: `App\Controller\UI\MdAppController:` works fine for me.

.env
mickel1138: BUT we can still use the same file path. 
Also you can use: `'%kernel.project_dir%/var/logs/...'`

src/Controller/UI/MdAppController.php
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * @param string $udi The dataset UDI identifier.
+     *
+     * @Route("/mdapp/getlog/{udi}",
+     *      name="app_ui_mdapp_getlog"
+     * )
+     *
+     * @return response
+     */
+    public function getlog($udi)
```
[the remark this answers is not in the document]
fightingtexasaggie: yeah, originally the logger was the writter, completely separate of the reader. I agree it belongs together.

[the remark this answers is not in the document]
fightingtexasaggie: Check it out now...I think this makes it considerably cleaner.

config/services.yaml
```diff
@@ -40,6 +44,10 @@ services:
             string $fromAddress: "%env(MAILER_FROM_ADDR)%"
             string $fromName: "%env(MAILER_FROM_NAME)%"
 
+    App\Util\MdappLogger:
+        arguments:
+            - '%kernel.project_dir%/var/log/mdapp.log'
```
praneethpr: Can use %kernel.logs_dir% to get to var/log/ folder
ex: %kernel.logs_dir%/mdapp.log

templates/AdminOnly.html.twig
```diff
@@ -0,0 +1,7 @@
+{% extends "template/UI.html.twig" %}
```
praneethpr: This file already exists in templates/template/AdminOnly

src/Controller/UI/MdAppController.php
```diff
@@ -0,0 +1,209 @@
...
+     *      methods={"GET"}
+     * )
+     *
+     * @return Response
+     */
+    public function defaultAction()
+    {
+        // Checks authorization of users
+        if (!$this->isGranted('ROLE_DATA_REPOSITORY_MANAGER')) {
+            return $this->render('AdminOnly.html.twig');
```
praneethpr: Need to change it to template/AdminOnly

config/services.yaml
```diff
@@ -40,6 +44,10 @@ services:
             string $fromAddress: "%env(MAILER_FROM_ADDR)%"
             string $fromName: "%env(MAILER_FROM_NAME)%"
 
+    App\Util\MdappLogger:
+        arguments:
```
mickel1138: Don't google what's the difference between Arguments and Bind

fightingtexasaggie: I'll just use bind to be consistent.

src/Controller/UI/MdAppController.php
```diff
@@ -0,0 +1,220 @@
...
+    /**
+     * Class constructor.
+     *
+     * @param EntityHandler $entityHandler         A Pelagos EntityHandler instance.
+     * @param string        $issueTrackingBaseUrl  Issue tracker base URL.
+     */
+    public function __construct(EntityHandler $entityHandler, string $issueTrackingBaseUrl)
+    {
+        $this->issueTrackingBaseUrl = $issueTrackingBaseUrl;
+        $this->entityHandler = $entityHandler;
```
praneethpr: Global variable in the class missing for entityHandler

fightingtexasaggie: added var

PR #410, opened 2019-10-16
Feature/pelagos 3800 add rabbitmq consumers

praneethpr:
Got it to work, Need to run
bin/console rabbitmq-supervisor:rebuild --wait-for-supervisord


config/packages/old_sound_rabbit_mq.yaml
```diff
@@ -2,14 +2,55 @@ old_sound_rabbit_mq:
...
-#            connection:       default
-#            exchange_options: { name: 'task', type: direct }
-#    consumers:
-#        task:
-#            connection:       default
-#            exchange_options: { name: 'task', type: direct }
-#            queue_options:    { name: 'task'}
-#            callback:         App\Consumer\TaskConsumer
+    producers:
+        dataset_file_hasher:
```
praneethpr: Removed unused producers

fightingtexasaggie:
reviewed all files and have compared diffs against develop, will begin testing...sorry this is taking a while

src/Consumer/DoiConsumer.php
```diff
@@ -0,0 +1,310 @@
...
+    /**
+     * Process a filer message.
+     *
+     * @param AMQPMessage $message A filer message.
+     *
+     * @return integer
+     */
+    public function execute(AMQPMessage $message)
+    {
+        // @codingStandardsIgnoreStart
```
mickel1138: Is this existing, or was this added? Also this directive will be depricated, it is now phpcs:enable (or something).

praneethpr: I think it was existing, should probably remove it

src/Consumer/DatasetFileHasherConsumer.php
```diff
@@ -0,0 +1,114 @@
...
+ * A consumer of dataset hash file request messages.
+ *
+ * Calculate a SHA256 hash of the dataset file and store it
+ * in the DatasetSubmission datasetFileSha256Hash attribute.
+ *
+ * @see ConsumerInterface
+ */
+class DatasetFileHasherConsumer implements ConsumerInterface
+{
+
```
mickel1138: I wonder why this new line didn't trigger a GCS issue.

src/Consumer/DoiConsumer.php
praneethpr: It is because of rabbitmq's variable delivery_info, which they haven't yet changed. Will change it to the new implementations of phpcs

src/Controller/UI/DataDiscoveryController.php
```diff
@@ -15,7 +15,7 @@
 
 use App\Entity\DatasetSubmission;
 use App\Entity\DIF;
-use App\Event\LogActionItemEventDispatcher;
+use App\EventListener\LogActionItemEventDispatcher;
```
mickel1138: Why the move? Should have put this is in it's own folder? EventDispatchers?

PR #414, opened 2019-10-21
Feature/pelagos 3804 add symfony commands

praneethpr:
Merge PELAGOS-3800 before this, Please let me know if I've left out any commands that we are currently using


src/Command/AccountExpirationNotifyCommand.php
```diff
@@ -0,0 +1,258 @@
...
+                'hostname' => $this->hostName,
+                'expireTimeStamp' => $expireTimeStamp,
+            );
+
+            $mailData['recipient'] = $person;
+
+            $this->mailer->sendEmailMessage(
+                $emailTemplate,
+                $twigData,
+                $person->getEmailAddress()
```
fightingtexasaggie: This seems to be expecting an array of addresses.

praneethpr: I refactored the Mailsender Utility class's method to take an array for the $toAddress, forgot to change this usage to array

src/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,155 @@
...
+                    // Set Modifier
+                    $datasetSubmission->setModifier($systemPerson);
+
+                    // Set filesize, hash, and original filename. (Cold Storage attributes).
+                    $datasetSubmission->setDatasetFileColdStorageAttributes($size, $hash, $originalFilename);
+
+                    // Set options for a new replacement datafile of the supplied Cold-Storage stubfile.
+                    $datasetSubmission->setDatasetFileTransferStatus(DatasetSubmission::TRANSFER_STATUS_NONE);
+                    $datasetSubmission->setDatasetFileName(null);
+                    $datasetSubmission->setDatasetFileSize(null);
```
fightingtexasaggie: I just noticed that before filer runs, the dataland template breaks because of a null here. I don't think we should fix on this ticket, and the bug exists in production, so I'll put it in the backlog. If for whatever reason we run this and filer isn't running, this will be a temporary issue.

mickel1138:
I think we should put every command we write into the Pelagos space... Thought? (question mark)

src/Command/ColdStorageFlagCommand.php
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Symfony command config section.
+     *
+     * @return void
+     */
+    protected function configure()
+    {
+        $this
+            ->setDescription('Marks dataset in input file is cold-stored and updates datafile')
+            ->addArgument('infofile', InputArgument::REQUIRED, 'Filename of the coldinfo-formatted textfile required.');
```
mickel1138: Where does this file come from? @fightingtexasaggie

fightingtexasaggie: it is generated by a python script that uses the hashdeep file.

fightingtexasaggie: verified (with an older db dump) this mailer works. 
(modify config/packages/dev/swiftmailer.yaml to test)

PR #415, opened 2019-10-22
Feature/pelagos 3806 add remaining ui applications

praneethpr:


public/js/dataset-summary.js
```diff
@@ -0,0 +1,102 @@
...
+            "datasetSubmissionHistory",
+            "datasetSubmissionHistory.creator",
+            "datasetSubmissionHistory.modifier",
+            "datasetPublications",
+            "datasetPublications.creator",
+            "datasetPublications.modifier",
+            "datasetPublications.publication"
+        ];
+        $.ajax({
+            url: "/api/datasets",
```
mickel1138: Should this be a relative address, or use FOS router like before?

src/Controller/DownloadController.php
```diff
@@ -43,15 +43,15 @@ public function __construct(EntityHandler $entityHandler)
...
      *
      * @Route("/download/{id}", name="pelagos_app_download_default")
      *
      * @return Response
      */
-    public function defaultAction(string $id)
+    public function defaultAction(int $id)
     {
-        $dataset = $this->entityHandler->get(Dataset::class, $id);
+        $dataset = $this->entityHandler->get(Dataset::class, (int) $id);
```
mickel1138: Why cast to int, it should be an int, because of the type hint.

src/Controller/DownloadController.php
```diff
@@ -124,7 +124,7 @@ public function httpAction(string $id, DataStore $dataStore, LogActionItemEventD
         $logActionItemEventDispatcher->dispatch(
             array(
                 'actionName' => 'File Download',
-                'subjectEntityName' => $em->getClassMetadata(get_class($dataset))->getName(),
+                'subjectEntityName' => 'Pelagos\Entity\Dataset',
```
mickel1138: I think the class name should stay, and instead we change the data. Just my $0.02.

src/Controller/DownloadController.php
praneethpr: It was debug code. Good catch

README.md
```diff
@@ -0,0 +1,61 @@
...
+* [RabbitMQ](https://www.rabbitmq.com/documentation.html) - Open source message broker software
+* [Composer 1.5.2+](https://getcomposer.org/) - Dependency Manager for PHP
+* [Yarn 1.15.2](https://yarnpkg.com/en/) - Package Manager
+
+### Installation
+
+Pelagos is a [Symfony 3.4](https://symfony.com/doc/3.4/index.html) project, please follow the normal configuration regarding setting up your webserver for a Symfony project.
+
+To install fullfill prerequisites then run:
+* `export SYMFONY-ENV=drupal_prod`
```
mickel1138: This is no longer needed now.

README.md
```diff
@@ -0,0 +1,61 @@
...
+Pelagos is a system for maintaining a repository of scientific research data.
+Developed and maintained by the The Gulf of Mexico Research Initiative Information and Data Cooperative (GRIIDC).
+URL: https://data.gulfresearchinitiative.org/
+## Getting Started
+
+These instructions will get you a copy of the project up and running on your local machine for development and testing purposes. See deployment for notes on how to deploy the project on a live system.
+
+### Prerequisites
+
+* [CENTOS 6/7](https://wiki.centos.org/) - Linux Distribution based on Red Hat
```
mickel1138: We should double check the versions, or add + to them.

README.md
```diff
@@ -0,0 +1,61 @@
...
+* [PHP 7.1](http://php.net/docs.php) - General-purpose scripting language
+* [PostgreSQL 9.6](https://www.postgresql.org/docs/9.6/static/release-9-6.html) - ORDBMS
+* [Elasticsearch 6.7.2+](https://www.elastic.co/products/elasticsearch) - ElasticSearch Document Indexer
+* [RabbitMQ](https://www.rabbitmq.com/documentation.html) - Open source message broker software
+* [Composer 1.5.2+](https://getcomposer.org/) - Dependency Manager for PHP
+* [Yarn 1.15.2](https://yarnpkg.com/en/) - Package Manager
```
mickel1138: We should double check the versions, or add + to them.

src/Command/AccountExpirationNotifyCommand.php
```diff
@@ -0,0 +1,258 @@
...
+                'hostname' => $this->hostName,
+                'expireTimeStamp' => $expireTimeStamp,
+            );
+
+            $mailData['recipient'] = $person;
+
+            $this->mailer->sendEmailMessage(
+                $emailTemplate,
+                $twigData,
+                $person->getEmailAddress()
```
fightingtexasaggie: this needs to be an array.

src/Controller/DefaultController.php
```diff
@@ -41,16 +46,15 @@ public function admin()
...
     public function showSiteMapXml()
     {
-        $container = $this->container;
-        $response = new StreamedResponse(function () use ($container) {
-
-            $entityManager = $container->get('doctrine.orm.entity_manager');
+        $response = new StreamedResponse(function () {
 
-            $datasets = $entityManager->getRepository(Dataset::class)->findBy(
+            $datasets = $this->getDoctrine()->getRepository(Dataset::class)->findBy(
```
mickel1138: No injection for the EM here?

src/Controller/UI/DatalandController.php
```diff
@@ -101,7 +101,7 @@ public function defaultAction(string $udi)
                 ->andwhere('a.subjectEntityName = ?2')
                 ->andwhere('a.actionName = ?3')
                 ->setParameter(1, $dataset->getId())
-                ->setParameter(2, 'App\Entity\Dataset')
+                ->setParameter(2, 'Pelagos\Entity\Dataset')
```
mickel1138: All this logic should be taken out, and dropped into a utility or something.
Also maybe we can resolve this by only using the main entity name like: "Dataset".

src/Controller/DefaultController.php
praneethpr: It is just another way to do it, but probably we should come up with a set of guidelines when we should inject and when we can use these symfony injected services

src/Controller/DownloadController.php
praneethpr: Probably we should do it as an improvement after we deploy flex

src/Controller/UI/SideBySideController.php
```diff
@@ -0,0 +1,259 @@
...
+        $this->entityHandler = $entityHandler;
+    }
+
+    /**
+     * The default action for Side by Side.
+     *
+     * @param Request     $request The Symfony request object.
+     * @param string|null $udi     The UDI of the Dataset to load.
+     *
+     * @Route("/sidebyside/{udi}", name="pelagos_app_ui_sidebyside_default")
```
mickel1138: You need a method "GET" in this route, are else it will conflict with the other similar route below. (same url, but expects post). If you don't specify a method, it will accept all.

praneethpr: We never specified the methods for GETs in the UI Controllers previously, Added GET methods to required UI Controllers

src/Controller/DownloadController.php
```diff
@@ -2,7 +2,7 @@
```
fightingtexasaggie: This is currently not working as it is hung up on a missing download_base_directory parameter, that seems to not be settable.

praneethpr: Also missing download_base_url, Good catch. I'll fix it

fightingtexasaggie: I'm playing around with the parameterBagInterface, to make things much much more sane imho.

PR #417, opened 2019-10-25
Added read only/maintenance mode

praneethpr:


templates/template/readonly.html.twig
```diff
@@ -0,0 +1,10 @@
+{% extends "template/UI.html.twig" %}
+
+{% block title %}
+    System is in read-only mode
```
fightingtexasaggie: This isn't really read-only though. We should probably ctall this maintenance mode across the board instead, especially since there's a separate ticket in the system for a true read-only mode.

praneethpr:
> In testing, I'm not able to see the it render the template/readonly.html.twig template regardless of setting the .env to true or not. I've tested both locally on on the webserver with the same results.
> 
> (Also, since there's a ticket for a read-only mode, it might prevent future confusion renaming this one and all internal references to maintenance mode altogether...or we can do that later.)

It should not be debug mode and should set .env variable to true. I should've mentioned that in the description. Yeah, I agree that we should change this to maintenance mode and rename the other one to a Banner or something like that

praneethpr:
> Re-tested with debug mode commented out and verified it working both on local server and apache. Code looks good.

Renamed it to Maintenance mode

src/EventListener/MaintenanceModeListener.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Kernel event listener method to enable/disbale maintenance mode.
+     *
+     * @param RequestEvent $event Symfony event variable.
+     *
+     * @return void
+     */
+    public function onKernelRequest(RequestEvent $event)
+    {
+        if ($this->maintenanceMode and !$this->debug) {
```
mickel1138: Why only in debug mode? Is this an additional safety?

templates/template/maintenanceMode.html.twig
```diff
@@ -0,0 +1,10 @@
+{% extends "template/UI.html.twig" %}
+
+{% block title %}
+    System is in read-only mode
```
mickel1138: Shouldn't this title also be "Maintenance Mode" ?

praneethpr: Fixed it

templates/template/maintenanceMode.html.twig
```diff
@@ -0,0 +1,10 @@
+{% extends "template/UI.html.twig" %}
+
+{% block title %}
+    Site is temporarily under maintenance
+{% endblock %}
+
+{% block body %}
+    <h1>Site is temporarily under maintenance</h1>
```
mickel1138: The verbiage could use some work, and it needs the guy with the shovel image too. ;-)

PR #418, opened 2019-10-28
Feature/pelagos 3805 add unit tests

praneethpr:
- Removed test suite for Form, which had only test cases for DatasetSubmissionType and were commented out. 

- Removed test suite for Exception, those exception classes aren't used in the project anymore. 

- Removed test cases which check for type, i think that is taken care by PHP7 and we shouldn't write individual test cases for it.

- Need help in PelagosEntityVoterTest, look at the commented out test cases in it


fightingtexasaggie:
I got the entity abstract class test working with a special mock for abstract clases, but still don't have the other two yet.

tests/unit/Security/PelagosEntityVoterTest.php
```diff
@@ -0,0 +1,371 @@
...
+                    $type,
+                    $this->mockTokens[$type][$roleName],
+                    $roleName
+                );
+            }
+        }
+    }
+
+//    /**
+//     * Test that all roles for associated entities can do or not do what they should be able to.
```
praneethpr: We still need to fix this test case?

tests/unit/Security/PelagosEntityVoterTest.php
```diff
@@ -0,0 +1,371 @@
...
+            $this->voter->vote(
+                $this->mockTokens['DataRepository'][DR_Roles::MANAGER],
+                $this->mockEntity,
+                array('THIS_IS_AN_UNSUPPORTED_ATTRIBUTE')
+            )
+        );
+    }
+
+//    /**
+//     * Test that the voter denies access for a token that does not resolve to a user.
```
praneethpr: and this

PR #421, opened 2019-11-07
Catch UidNumberInUseInLDAPException exception

fightingtexasaggie:


mickel1138:
Can you provide a description for the PR?

praneethpr:
> Can you provide a description for the PR?

@fightingtexasaggie posted on slack:
In the account controller, there seems to be a bug on changing a known password. It's LDAP related. The password is actually being changed in pelagos, but getting a UID already in use from LDAP type error, which is odd because it should not be creating a new ldap entry.

fightingtexasaggie:
What was actually happening was relying on an exception to be thrown if a create failed on a new ldap object,t hen it woudld just update. The rationale was originally that this would replace any missing ldap entries if they were missing. I just had to catch the right exception and it works correctly again.

PR #426, opened 2019-11-07
Misc/minor annoyances

fightingtexasaggie:
js sort tested on chrome, Firefox, and Edge. Since we're going to have a big testing session with lots of people in as DRPMs, I thought it nice to finally alphabatize the DIF title list.

Also, the link on the default menu now points to the generic change password logic that does a change if logged in, and a reset if not logged in...this mirrors what the drop down under the 'help' menu does.

### DIFF
--- a/public/js/dif.js
+++ b/public/js/dif.js
@@ -287,7 +287,18 @@ $(document).ready(function()
             }
         });
     }
+
+    var select = $('select#researchGroup');
+        select.html(select.find('option').sort(function(x, y) {
+            if ($(x).text() > $(y).text()) {
+                return 1;
+            } else {
+                return -1;
+            }
+    }));
+    $("select option:contains(PLEASE SELECT A PROJECT)").prependTo($("select#researchGroup"));
 });
+
 
 function difStatus(id, status)
 {
--- a/public/js/dif.js
+++ b/public/js/dif.js
@@ -287,18 +287,7 @@ $(document).ready(function()
             }
         });
     }
-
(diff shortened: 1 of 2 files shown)


templates/Default/index.html.twig
```diff
@@ -5,7 +5,7 @@
     <h3>Links to:</h3>
     <ul>
         <li><a href="{{  path("pelagos_app_ui_account_default") }}">Request Account</a></li>
-        <li><a href="{{  path("pelagos_app_ui_account_passwordreset") }}">Password Reset</a></li>
+        <li><a href="{{  path("pelagos_app_ui_account_password") }}">Change/Reset password</a></li>
```
praneethpr: Text change is fine, but the route name change is not necessary and it needs to be changed in the Controllers as well, if we want to make this change

public/js/dif.js
```diff
@@ -287,8 +287,19 @@ $(document).ready(function()
             }
         });
     }
+
+    var select = $('select#researchGroup');
+        select.html(select.find('option').sort(function(x, y) {
```
praneethpr: 👍

templates/Default/index.html.twig
praneethpr: Never mind, I thought this route dint exist

fightingtexasaggie:
It's only an annoying to DRPMs, and it's an unholy waste of everyone's time to deal with long unsorted lists, but I'll ping the PO.

PR #427, opened 2019-11-14
Catch UidNumberInUseInLDAPException exception.

fightingtexasaggie:
This was already approved, but set to merge to the upgrade/flex branch. I could not make it cleanly rebase, so I just saved out a patch, created a new branch from merge/flex3-changes, applied the patch to it, then opened this PR.  (gotta love diff and patch)

### DIFF
--- a/src/Controller/UI/AccountController.php
+++ b/src/Controller/UI/AccountController.php
@@ -14,6 +14,7 @@ use App\Entity\Password;
 use App\Entity\Person;
 use App\Entity\PersonToken;
 use App\Event\EntityEventDispatcher;
+use App\Exception\UidNumberInUseInLDAPException;
 use App\Handler\EntityHandler;
 use App\Util\Factory\UserIdFactory;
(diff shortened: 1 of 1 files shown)


src/Controller/UI/AccountController.php
```diff
@@ -440,7 +441,7 @@ public function changePasswordPostAction(Request $request, Ldap $ldap)
         try {
             // Try to add the person to LDAP, incase it needs to re-create.
             $ldap->addPerson($person);
-        } catch (LdapException $exception) {
+        } catch (UidNumberInUseInLDAPException $exception) {
```
mickel1138: So, an LdapException happens when the actual update/create method fails. This could happen because the LDAP object didn't exist. Maybe this should just be changed to regular "exception" to it catches both of them?

fightingtexasaggie: future improvement hopefully

PR #428, opened 2019-11-14
New ldap update command.

mickel1138:
Can update ldap passwords, and set password in database.

Option to command name suggestions.

### DIFF
--- a/src/Command/UpdateLdapAccountCommand.php
+++ b/src/Command/UpdateLdapAccountCommand.php
@@ -81,7 +81,6 @@ class UpdateLdapAccountCommand extends Command
                 InputOption::VALUE_NONE,
                 'Also flush password change to database.'
             );
-        ;
     }
 
     /**
--- a/src/Command/UpdateLdapAccountCommand.php
+++ b/src/Command/UpdateLdapAccountCommand.php
@@ -17,25 +17,35 @@ use App\Entity\Password;
 use App\Entity\Person;
 use App\Util\Ldap\Ldap;
 
+/**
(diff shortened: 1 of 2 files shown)


src/Command/UpdateLdapAccountCommand.php
```diff
@@ -0,0 +1,187 @@
...
+                if (empty($password)) {
+                    throw new \RuntimeException('Password cannot be empty.');
+                }
+
+                $password = new Password($password);
+
+                return $password;
+            });
+            $io->text('Setting Password');
+            $passwordsMatch = $user->getPasswordEntity()->comparePassword($password->getClearTextPassword());
```
mickel1138: I'm thinking I wanna take this bit out, it would allow for password guessing.

src/Command/UpdateLdapAccountCommand.php
```diff
@@ -0,0 +1,187 @@
...
+
+        if ($flush) {
+            $io->text('Flushing to database');
+            $this->entityManager->persist($user);
+            $this->entityManager->flush();
+        }
+
+        $io->success(sprintf('LDAP account for user %s has been updated.', $username));
+
+        return 0;
```
praneethpr: We return void on Commands

fightingtexasaggie:
I noticed this will break if you remove the ldapPublicKey object class and associated keys from the ldap as well as remove the public keys from the database. Not all POSIX users have the ldapPublicKey object class, and the Ldap.php on line 166 seems to require it.

PR #430, opened 2019-11-18
Removed base url from the string search to fix url comparison

praneethpr:


src/Pelagos/Bundle/AppBundle/Command/DoiComparisonCommand.php
```diff
@@ -410,6 +410,7 @@ private function getDoiStatus(string $state): string
      */
     private function isUrlValid(string $url, string $needle): bool
     {
+        $url = str_replace('https://data.gulfresearchinitiative.org/', '', $url);
```
fightingtexasaggie: Just in case someone sticks in (although they shouldn't) http instead of https, maybe we need to catch that as well.

fightingtexasaggie: future improvementt maybe, but this will work for the proper case

praneethpr: I don't think anyone else sets the landing page url other than the Rabbitmq DoiConsumer class, and we've hardcoded the url to point to our production website. I think it should be good

PR #442, opened 2019-12-04
This should fix the resourceGroup locked message.

mickel1138:


public/js/dif.js
```diff
@@ -938,9 +938,12 @@ function loadPOCs(researchGroup,ppoc,spoc)
             var researchGroupLocked = $("#researchGroup option[value=" + researchGroup + "]").attr("locked");
             if (researchGroupLocked == "true") {
                 $("#status").val("closedout");
-                setFormStatus();
+            } else {
+                $("#status").val(0);
```
praneethpr: It is previously set to a string and now an integer value, but i looked at the setFormStatus() function it checks if it is a 0 value. In future or it can be an improvement, we should probably stick to the same type for a variable

PR #443, opened 2019-12-04
Add --wait-for-supervisord to rabbit control script.

fightingtexasaggie:
just a quick fix for rabbitmq control

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -7,7 +7,7 @@ export SYMFONY_ENV=drupal_prod
 export SYMFONY_DEBUG=0
 
 # Stop RabbitMQ consumers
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -1,24 +1,17 @@
 #/bin/sh
 
-# Set Symfony environment to drupal_prod
-export SYMFONY_ENV=drupal_prod
-
 # Disable Symfony debug mode
 export SYMFONY_DEBUG=0
```
mickel1138: This is also not needed.

mickel1138: The export above does not do anything in Symfony 4. Unless it's used for something else, it can be removed.

PR #444, opened 2019-12-04
Removed usage of global $_ENV in Api/Account Controller

praneethpr:


config/services.yaml
```diff
@@ -12,6 +12,7 @@ parameters:
     router.request_context.host: "%env(HOSTNAME)%"
     router.request_context.scheme: "https"
     router.request_context.base_url: '%env(REQUEST_CONTEXT_BASE_URL)%'
+    account_less_strict_password_rules: '%env(ACCOUNT_LESS_STRICT_PASSWORD_RULES)%'
```
mickel1138: No bind, it's only used in the AccountController?

praneethpr: Yes, it is used only once. Should we use bind instead of this?

praneethpr: If we make a constructor, we need to initialize the parent constructors and it looks like there is a lot of inheritance

PR #445, opened 2019-12-05
Edge case (maybe not really), but covered the fact that loading exist…

mickel1138:
…ing locked or unlocked DIF shows the correct message.


fightingtexasaggie:
It may have been a caching issue on the javascript change. Seems to be working now.

PR #447, opened 2019-12-05
Added migration to remove audit tables and updated libraries

praneethpr:


src/Migrations/Version20191205223937.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->addSql('DROP TABLE dif_audit');
+        $this->addSql('DROP TABLE data_repository_audit');
+        $this->addSql('DROP TABLE distribution_point_audit');
+        $this->addSql('DROP TABLE person_data_repository_audit');
+        $this->addSql('DROP TABLE funding_cycle_audit');
+        $this->addSql('DROP TABLE person_audit');
+        $this->addSql('DROP TABLE funding_organization_audit');
+        $this->addSql('DROP TABLE person_funding_organization_audit');
+        $this->addSql('DROP TABLE research_group_audit');
+        $this->addSql('ALTER TABLE person_token ALTER valid_for TYPE VARCHAR(255)');
```
fightingtexasaggie: Were these 3 alter table lines just pending?

praneethpr: They are already in the previous migrations but i dont know why doctrine thinks that it needs to be run again

mickel1138: This may be one of the differences between Doctrine versions.

PR #449, opened 2019-12-16
Bugfix/testfile location

fightingtexasaggie:
No more need to have the testfile.bin in a webserver aliased location since we have a true public folder, per instance.

### DIFF
--- a/public/js/download.js
+++ b/public/js/download.js
@@ -87,7 +87,7 @@ function testDownload(fileSize) {
     let start = new Date().getTime();
     $.ajax({
         type: "GET",
-        "url": Routing.generate("pelagos_homepage") + "/images/testfile.bin?id=" + start,
+        "url": Routing.getBaseUrl() + "/images/testfile.bin?id=" + start,
         success: function(msg) {
             end = new Date().getTime();
             diff = (end - start) / 1000;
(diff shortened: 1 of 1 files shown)


public/js/download.js
```diff
@@ -87,7 +87,7 @@ function testDownload(fileSize) {
     let start = new Date().getTime();
     $.ajax({
         type: "GET",
-        url: "/download/testfile.bin?id=" + start,
+        "url": Routing.generate("pelagos_homepage") + '/images/testfile.bin?id=' + start,
```
praneethpr: Just to be consistent, use double quotes

public/js/download.js
```diff
@@ -87,7 +87,7 @@ function testDownload(fileSize) {
     let start = new Date().getTime();
     $.ajax({
         type: "GET",
-        url: "/download/testfile.bin?id=" + start,
+        "url": Routing.generate("pelagos_homepage") + "/images/testfile.bin?id=" + start,
```
mickel1138: Just use asset('path to file')

mickel1138:
Test file doesn't have to go into images folder, can be in the root, it's own folder, anywhere really.

public/js/download.js
mickel1138: Ok, can't use asset tag in Javascript, so you'll have to set it somewhere in the template.

mickel1138: https://symfony.com/doc/4.3/frontend/encore/server-data.html

mickel1138: Ok, forget all that, I think it's easier if you just want to the base path, just get it from Routing like this: `Routing.getBaseUrl()`

PR #457, opened 2019-12-19
Updated script with new symfony command name.

fightingtexasaggie:


fightingtexasaggie:
This is the only one I found in repository-commited scripts that call symfony commands.

PR #459, opened 2019-12-20
Added templating engine, this will now handle pages not found, and ot…

mickel1138:
…her error pages.

Only show real template in prod mode, but in debug the error will be correct now also.


mickel1138:
Don't merge yet, this is for testing only, as this may break the REST API.

fightingtexasaggie:
This has to be tested in PROD mode ;-) hehe

PR #464, opened 2020-01-03
Added udi field to search query

praneethpr:
To test, use this UDI: R5.x284.000:0009. It is used in the abstract of another dataset

### DIFF
--- a/src/Util/Search.php
+++ b/src/Util/Search.php
@@ -694,7 +694,6 @@ class Search
      */
     private function getUdiQuery(string $queryTerm): Query\MatchPhrase
     {
-        // Add title field to the query
         $udiQuery = new Query\MatchPhrase();
         $udiQuery->setFieldQuery(self::ELASTIC_INDEX_MAPPING_UDI, $queryTerm);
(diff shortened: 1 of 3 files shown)


src/Util/Search.php
```diff
@@ -678,4 +684,21 @@ private function getCollectionEndDateQuery(array $collectionDates): Query\Range
...
+
+    /**
+     * Get the UDI query.
+     *
+     * @param string $queryTerm Query term that needs to be searched upon.
+     *
+     * @return Query\Match
+     */
+    private function getUdiQuery(string $queryTerm): Query\Match
+    {
```
fightingtexasaggie: perhaps a regex in here to determine if the term is an UDI, and if it is, use the matchPhrase. On discovery, we also allowed the use of periods in place of the colon then substituted the colon back in.

praneethpr: I replaced this with the existing query with MatchPhrase, the way we do it  on data-discovery. I get only 1 result, even though that UDI is mentioned in another dataset's abstract (this is how it works on Data-discovery as well)

praneethpr: I changed it to Matchphrase, and i figured that we were nulling the query term in data-discovery. Now it works the way we want. Its ready for test

PR #465, opened 2020-01-06
ID must by cast to int, because RabbitPublisher does not seem to like…

mickel1138:
… the "object" being send. And crashes.


mickel1138:
> In testing, I get this upon submission:
> An instance of Symfony\Component\Templating\EngineInterface must be injected in FOS\RestBundle\View\ViewHandler to render templates.

This is fixed is another pull request, but this shouldn't happen. Try clearing cache, composer install etc.

mickel1138:
On the Templating error, that's not the actual error that happens. The e-mail templates are trying to use doi, but doi can be null. So in templates dataset.doi.doi should not be used, but just dataset.doi, because that one stringyfies the DOI.

mickel1138:
> tested and verified email received upon submission

@fightingtexasaggie are you getting a filer e-mail?

mickel1138:
This seems to no longer be needed now, killing the PR.
Found in DMS-679, PELAGOS-3859

PR #470, opened 2020-01-08
Feature/pelagos 3862 make method field required

fightingtexasaggie:


src/Form/DatasetSubmissionType.php
```diff
@@ -132,7 +132,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ))
             ->add('suppMethods', Type\TextareaType::class, array(
                 'label' => 'Supplemental Information - Methods',
-                'required' => false,
+                'required' => true,
```
praneethpr: We use the same Form type in Dataset-review tool, and the PO's don't want it to be required in the Dataset-review tool

fightingtexasaggie:
Separating the datasetSubmission and datasetSubmissionReview types creates a heck of a lot of code duplication, but neither inherit_data, php traits, nor inheritance seems to be able to help.  I'm open to ideas.

src/Form/DatasetSubmissionType.php
fightingtexasaggie: I've separated them now, but if you have a suggestion for code-reuse, let me know. It's a painful amount of duplication.

src/Entity/DatasetSubmission.php
```diff
@@ -334,7 +334,7 @@ class DatasetSubmission extends Entity
             'description' => 'energy, water and waste systems and communications infrastructure and services. Examples: hydroelectricity, geothermal, solar and nuclear sources of energy, water purification and distribution, sewage collection and disposal, electricity and gas distribution, data communication, telecommunication, radio, communication networks',
         ],
     ];
-    
+
```
praneethpr: Lot of whitespaces on this file

PR #472, opened 2020-01-13
Added Bcc to all emails

praneethpr:


src/Util/MailSender.php
```diff
@@ -73,6 +83,7 @@ public function sendEmailMessage(
             ->setSubject($emailTwigTemplate->renderBlock('subject', $mailData))
             ->setFrom($this->from)
             ->setTo($toAddresses)
+            ->setBcc($this->bccAddress)
```
mickel1138: What happens if this is null, blank, not set? Error, or just no e-mail?

fightingtexasaggie: well, .env is checked into git and contains a default, so is this still a concern?

praneethpr: I don't think there would be a scenario where it would not be set, the Constructor of MailSender expects a string and it is by default set in the .env file. I tried with empty string, there is no email sent

.env
```diff
@@ -36,6 +36,7 @@ RABBITMQ_URL=amqp://guest:guest@localhost:5672
 # For a generic SMTP server, use: "smtp://localhost:25?encryption=&auth_mode="
 # Delivery is disabled by default via "null://localhost"
 MAILER_URL=null://localhost
+BCC_MAILER_ADDR='email@domain.null'
```
mickel1138: Should this be moved to line 56, to be with the other variables regarding the mailer? Maybe even put them in a ### block. Also named the same? If so, maybe something like "MAILER_BCC_ADDR" or "MAILER_BCC_TO" ?

praneethpr: Done

PR #474, opened 2020-01-14
Added must not match query to all fields

praneethpr:
To test:

For ex: Title field
Deepwater Horizon 
Deepwater -Horizon (Horizon word shouldn't appear on the results)

For ex.: Authors field
Steven Murawski (lesser results)
Steven -Murawski (more results)

### DIFF
--- a/src/Util/Search.php
+++ b/src/Util/Search.php
@@ -139,6 +139,30 @@ class Search
         // Bool query to get range temporal extent dates
         $collectionDateBoolQuery = new Query\BoolQuery();
 
+        // Check if exclude term exists in the given query term
+        $mustNotQueryTerm = '';
+        if (preg_match_all('/-\b(\w*\w*)\b/', $queryTerm, $matches)) {
+            $mustNotQueryTerm = $matches[1][0];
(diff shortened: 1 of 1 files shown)


src/Util/Search.php
```diff
@@ -139,6 +139,30 @@ public function buildQuery(array $requestTerms): Query
         // Bool query to get range temporal extent dates
         $collectionDateBoolQuery = new Query\BoolQuery();
 
+        // Check if exclude term exists in the given query term
+        $mustNotQueryTerm = '';
+        if (preg_match_all('/-\b(\w*\w*)\b/', $queryTerm, $matches)) {
```
mickel1138: Only one `\w` seems to word too. And one could instead of the last word boundry search for `\S*` (any space/new line etc). (optional)

src/Util/Search.php
```diff
@@ -139,6 +139,30 @@ public function buildQuery(array $requestTerms): Query
...
+            $mustNotQueryTerm = $matches[1][0];
```
mickel1138: This only supports one word, I think it should support all words that start with `-`

praneethpr:
Closing this branch, typo in the branch name

PR #476, opened 2020-01-15
Feature/pelagos 3871 add must not match functionality

praneethpr:
To test:

For ex: Title field
Deepwater Horizon
Deepwater -Horizon (Horizon word shouldn't appear on the results)

For ex.: Authors field
Steven Murawski (lesser results)
Steven -Murawski (more results)

### DIFF
--- a/src/Util/Search.php
+++ b/src/Util/Search.php
@@ -144,24 +144,15 @@ class Search
         $splitUpQueryTerms = $this->splitQueryTerms($queryTerm);
         if (!empty($splitUpQueryTerms)) {
             $queryTerm = $splitUpQueryTerms['mustMatch'];
-            $mustNotQueryTerm = $splitUpQueryTerms['mustNotMatch'];
+            $mustNotQueryTerms = $splitUpQueryTerms['mustNotMatch'];
         }
 
         // If exclude term exists add the must not query
-        if ($mustNotQueryTerm) {
-            $mustNotMultiMatch = new Query\MultiMatch();
-            $mustNotMultiMatch->setFields(
-                [
-                    self::ELASTIC_INDEX_MAPPING_ABSTRACT,
-                    self::ELASTIC_INDEX_MAPPING_TITLE,
-                    self::ELASTIC_INDEX_MAPPING_UDI,
-                    self::ELASTIC_INDEX_MAPPING_DOI,
-                    self::ELASTIC_INDEX_MAPPING_AUTHORS,
-                    self::ELASTIC_INDEX_MAPPING_THEME_KEYWORDS
-                ]
-            );
-            $mustNotMultiMatch->setQuery($mustNotQueryTerm);
-            $subMainQuery->addMustNot($mustNotMultiMatch);
+        if (!empty($mustNotQueryTerms)) {
+            foreach ($mustNotQueryTerms as $mustNotQueryTerm) {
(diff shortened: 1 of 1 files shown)


src/Util/Search.php
```diff
@@ -765,4 +790,29 @@ private function doesUdiExistInQueryTerm(string $queryTerm, Query\BoolQuery $fie
...
+        $splitUpQueryTerms = array();
+        if (preg_match_all('/-\b(\w*)\b/', $queryTerm, $matches)) {
+            $splitUpQueryTerms = array(
+                'mustNotMatch' => '',
+                'mustMatch' => ''
+            );
+            foreach ($matches[1] as $match) {
+                $splitUpQueryTerms['mustNotMatch'] = trim($splitUpQueryTerms['mustNotMatch'] . ' ' . $match);
+            }
+            $splitUpQueryTerms['mustMatch'] = str_replace('-', '', $queryTerm);
```
fightingtexasaggie: I think this is maybe a tad too aggressive, removing embedded-between-words instead of just -excludes.  For example, a title contained the string 'July-October' which didn't work as expected when trying to exclude it. I think adding the word boundry as above would fix.

praneethpr: Fixed it

src/Util/Search.php
```diff
@@ -765,4 +790,29 @@ private function doesUdiExistInQueryTerm(string $queryTerm, Query\BoolQuery $fie
...
+    private function splitQueryTerms(string $queryTerm): array
+    {
+        $splitUpQueryTerms = array();
+        if (preg_match_all('/\s-\b(\w*)\b/', $queryTerm, $matches)) {
+            $splitUpQueryTerms = array(
+                'mustNotMatch' => '',
+                'mustMatch' => ''
+            );
+            foreach ($matches[1] as $match) {
+                $splitUpQueryTerms['mustNotMatch'] = trim($splitUpQueryTerms['mustNotMatch'] . ' ' . $match);
```
fightingtexasaggie: Did you mean to append here instead?

praneethpr: I'm not sure what you are asking

fightingtexasaggie: Wouldn't this result in only the very last term in the match array being used?

fightingtexasaggie: I see it now...my bad...

src/Util/Search.php
```diff
@@ -765,4 +779,60 @@ private function doesUdiExistInQueryTerm(string $queryTerm, Query\BoolQuery $fie
...
+     * Split the query terms into must match and must not match terms.
+     *
+     * @param string $queryTerm Query term that needs to be searched upon.
+     *
+     * @return array
+     */
+    private function splitQueryTerms(string $queryTerm): array
+    {
+        $splitUpQueryTerms = array();
+        if (preg_match_all('/\s-\b(\w*)\b/', $queryTerm, $matches)) {
```
mickel1138: Regex still a bit wrong, if you start your search with a negative item first, it will ignore the dash.

src/Util/Search.php
```diff
@@ -168,7 +184,7 @@ public function buildQuery(array $requestTerms): Query
 
         $mainQuery->setQuery($subMainQuery);
         $mainQuery->setFrom(($page - 1) * 10);
-
+
```
mickel1138: These spaces may cause GCS issues?

PR #479, opened 2020-01-16
Refactor/nf sv4 acl compat homedir consumer

fightingtexasaggie:


config/services.yaml
```diff
@@ -136,6 +136,7 @@ services:
     App\Consumer\CreateHomedirConsumer:
         bind:
             Symfony\Bridge\Monolog\Logger $logger: '@monolog.logger.create_homedir'
+            $fsType: "%env(FILESYSTEM_TYPE)%"
```
mickel1138: I like fully spelled out variable names, so you instantly can see what it's for.

fightingtexasaggie:
similarly, I have a bash script ready that can be used to apply ACL/perms/owners to existing homedirs once they're rsynced to the NetApp.

src/Consumer/CreateHomedirConsumer.php
```diff
@@ -54,6 +64,133 @@ public function __construct(
     * @return boolean True if success, false otherwise.
     */
     public function execute(AMQPMessage $message)
+    {
+        if ('NFS' === $this->filesystemType) {
+            return($this->processNFS($message));
+        } else {
```
mickel1138: Not a huge deal, but the other setting could be anything else, but is expected to be "Linux" as set in the .env file.

src/Consumer/CreateHomedirConsumer.php
```diff
@@ -54,6 +64,133 @@ public function __construct(
     * @return boolean True if success, false otherwise.
     */
     public function execute(AMQPMessage $message)
+    {
+        if ('Linux' === $this->filesystemType) {
+            return($this->processNormalFS($message));
+        } else {
```
mickel1138: My point was more the second else, should be an `if`?

PR #480, opened 2020-01-21
Changed logic and refactored the query to match all words in either o…

praneethpr:
…ne of the fields

### DIFF
--- a/src/Util/Search.php
+++ b/src/Util/Search.php
@@ -76,6 +76,11 @@ class Search
     );
 
     /**
+     * Index boost for Title, Authors, Theme Keywords.
+     */
+    const BOOST = '^2';
+
+    /**
      * Constructor.
      *
      * @param TransformedFinder      $finder        The finder interface object.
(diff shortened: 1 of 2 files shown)


src/Util/Search.php
```diff
@@ -807,32 +737,21 @@ private function splitQueryTerms(string $queryTerm): array
...
-     * @return Query\BoolQuery
+     * @return Query\MultiMatch
      */
-    private function getMustNotIncludeTermsQuery(string $mustNotQueryTerm): Query\BoolQuery
+    private function getMustNotIncludeTermsQuery(string $mustNotQueryTerm): Query\MultiMatch
     {
-        $mustNotBoolQuery = new Query\BoolQuery();
-
         $mustNotMultiMatch = new Query\MultiMatch();
         $mustNotMultiMatch->setFields(
```
mickel1138: Don't we need the same cross_field type here?

praneethpr: No. We don't need it, so that the given term does not match in any field

PR #481, opened 2020-01-21
Integrate the static styled website with GRIIDC theme

praneethpr:
jQuery validate error messages styling is kind of ugly, probably we should look into bootstrap validation methods


praneethpr:
> Looks nice, but I did notice the datepicker isn't quite working right. If you change the month or year in the drop-down selector, it doesn't update the field unless you click a day again.

I think that's how it always used to work? Its the same jQuery datepicker which we use on DIF, and other places. We can implement a more modern one?

praneethpr:
> jQuery validate error messages styling is kind of ugly, probably we should look into bootstrap validation methods

Removed jQuery validate, I don't think it is necessary for this application

templates/Search/form.html.twig
```diff
@@ -1,69 +1,104 @@
...
-                   value = "{{ query|default }}">
-            {% if page is defined and page is not null%}
-                <input type="hidden" id="pageNo" name="page" value="{{ page | default }}">
-            {% endif %}
-            <span class="input-group-btn">
-                <button id="searchSubmit" class="btn btn-danger" type="submit">
-                    <span class=" glyphicon glyphicon-search"></span>
-                </button>
-            </span>
+<!-- ========================= SECTION CONTENT ========================= -->
```
mickel1138: Should probably remove all these. Doesn't seem to make much sense.

templates/Search/form.html.twig
```diff
@@ -1,69 +1,104 @@
...
+                               value = "{{ query|default }}"
+                               required>
+                        {% if page is defined and page is not null%}
+                            <input type="hidden" id="pageNo" name="page" value="{{ page | default }}">
+                        {% endif %}
+                    </div>
+                    <div class="col-sm-3 btn-toolbar">
+                        <button id="searchSubmit" type="submit" class="btn btn-primary mx-2 w-50">Search
+                            <i class="fa fa-search"></i></button>
+                        <button type="reset" class="btn btn-dark mx-2 w-25">Clear</button>
```
mickel1138: This button doesn't work?

templates/Search/form.html.twig
```diff
@@ -1,69 +1,104 @@
...
-            {% endif %}
-            <span class="input-group-btn">
-                <button id="searchSubmit" class="btn btn-danger" type="submit">
-                    <span class=" glyphicon glyphicon-search"></span>
-                </button>
-            </span>
+<!-- ========================= SECTION CONTENT ========================= -->
+<section class="section-content bg pt-5">
+    <div class="container">
+        <div class="card card-header">
```
mickel1138: Card may not be the right thing to use here. If you just want a different background there are other ways to do it. Also the card doesn't scale right it seems.

templates/Search/form.html.twig
praneethpr: Removed it

templates/Search/form.html.twig
praneethpr: It does a basic HTML reset, I'm not sure what would the PO's want for that button to do

PR #484, opened 2020-01-27
Clear button simply reloads blank form now.

fightingtexasaggie:


public/js/search.js
```diff
@@ -172,6 +172,11 @@ $(document).ready(function () {
         $("#collectionEndDate").datepicker('show');
     });
 
+    $("#search-clear").click(function (e) {
+        e.preventDefault();
```
mickel1138: Instead of doing this, just turn the button into a button, not a reset button.

templates/Search/form.html.twig
```diff
@@ -18,7 +18,7 @@
                     <div class="col-sm-3 btn-toolbar">
                         <button id="searchSubmit" type="submit" class="btn btn-primary mx-2 w-50">Search
                             <i class="fa fa-search"></i></button>
-                        <button type="reset" class="btn btn-dark mx-2 w-25">Clear</button>
+                        <button type="reset" id="search-clear" class="btn btn-dark mx-2 w-25">Clear</button>
```
mickel1138: Just make button type into "button" so it won't do reset functionality.

PR #490, opened 2020-01-28
Feature/pelagos 3890 size erddap button

fightingtexasaggie:
I got rid of the images of buttons and replaced with buttons, but in loading, multiple classes get applied and they go throw 2 different themes. It's functional, but beyond ugly.


praneethpr:
Remove the divs and the <br> tags, the buttons should be in the same row

fightingtexasaggie:
See what y'all think all on a single line. The first 2 buttons have routes associated with them, the erddap one does not so I had to get add a data attribute in the template to pass to the js. What I don't like is how the button theme is applied late, and you can see the buttons transition. I guess I could hide them until document.ready to prevent this nastiness.

praneethpr:
> See what y'all think all on a single line. The first 2 buttons have routes associated with them, the erddap one does not so I had to get add a data attribute in the template to pass to the js. What I don't like is how the button theme is applied late, and you can see the buttons transition. I guess I could hide them until document.ready to prevent this nastiness.

For those download buttons, we can use font-awesome icons which will give it a better appearance ?

mickel1138:
I not sure if I like the plain buttons, they are not very prominent.

templates/Dataland/summary.html.twig
```diff
@@ -14,24 +14,23 @@
...
+                            <button id="downloaddsden" data-download-url="{{dataset.datasetSubmission.erddapurl}}" >Download Data</button>
                     {% endif %}
                     {% if dataset.datasetStatus == constant('App\\Entity\\Dataset::DATASET_STATUS_ACCEPTED') %}
-                        <img width="40px" height="40px" id="metadatadl" src="{{ asset('images/xmlc.png') }}">
+                            <button id="metadatadl" data-download-url="{{dataset.datasetSubmission.erddapurl}}" >XML Metadata</button>
                     {% else %}
-                        <img width="40px" height="40px" id="metadatadl-dis" src="{{ asset('images/xmlc-bw.png') }}">
+                            <button id="downloadds" data-download-url="{{dataset.datasetSubmission.erddapurl}}" >XML Metadata</button>
                     {% endif %}
                     {% set TEMPORYDISABLED = true %}
```
mickel1138: This line can go.

mickel1138: No, it can stay for now.

PR #492, opened 2020-01-29
Pelagos 3854 add e mail logger to monolog

mickel1138:


src/Util/MailerLoggerUtil.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Returns the LogLevel according to Swift Event Result.
+     *
+     * @param Swift_Events_SendEvent $event The SwiftMailer Event.
+     *
+     * @return string
+     */
+    private function getLogLevel(Swift_Events_SendEvent $event)
+    : string
```
praneethpr: It can be in the same line

src/Util/MailerLoggerUtil.php
```diff
@@ -0,0 +1,114 @@
...
+            // Sending has yet to occur
+            case Swift_Events_SendEvent::RESULT_PENDING:
+                return LogLevel::WARNING;
+
+            // Email is spooled, ready to be sent
+            case Swift_Events_SendEvent::RESULT_SPOOLED:
+                return LogLevel::NOTICE;
+
+            // Sending failed
+            default:
```
praneethpr: Just for consistency, do we want to add break statements for each case and move the default case to the end

src/Util/MailerLoggerUtil.php
```diff
@@ -0,0 +1,114 @@
...
+      * @var Logger
+      */
+    protected $logger;
+
+    /**
+     * MailerLoggerUtil constructor.
+     *
+     * @param LoggerInterface $logger The Logger Interface.
+     */
+    public function __construct(LoggerInterface  $logger)
```
praneethpr: We don't need to use Psr\Log\LoggerInterface, monolog already implements that in its class. So we can just create a Monolog logger object, like how we have done in the Consumers, etc

config/services.yaml
```diff
@@ -200,3 +200,9 @@ services:
     App\Controller\Api\DatasetPublicationController:
         bind:
             FOS\ElasticaBundle\Persister\ObjectPersister $objectPersister: "@fos_elastica.object_persister.pelagos.dataset"
+
+    App\Util\MailerLoggerUtil:
+        arguments: ["@logger"]
+        tags:
+            - { name: monolog.logger, channel: mailer }
```
praneethpr: I think this is the old way of how we used to do it, we can use bind to the respective logger channel

praneethpr:
> Logging works, but emails aren't getting sent. Not sure if its an environment issue

Emails are working, got delivered late.

src/Util/MailerLoggerUtil.php
mickel1138: Somewhere i'm thinking one of these ways is deprecated.

mickel1138: This is how it's documents in Symfony 5.0:
https://symfony.com/doc/current/components/console/logger.html

src/Util/MailerLoggerUtil.php
```diff
@@ -0,0 +1,122 @@
...
+      * @var Logger
+      */
+    protected $logger;
+
+    /**
+     * MailerLoggerUtil constructor.
+     *
+     * @param LoggerInterface $logger The Logger Interface.
+     */
+    public function __construct(Logger $logger)
```
praneethpr: I don't know how I made the other loggers work with Monolog\Logger class injection, but according to the documentation whatever you did was right . Should use the LoggerInterface from Psr\Log
[https://symfony.com/doc/current/logging/channels_handlers.html#how-to-autowire-logger-channels](url)

src/Util/MailerLoggerUtil.php
```diff
@@ -0,0 +1,122 @@
...
+use Swift_Events_SendEvent;
+use Swift_Events_SendListener;
+
+/**
+ * A utility to log Swift e-mails.
+ */
+class MailerLoggerUtil implements Swift_Events_SendListener
+{
+     /**
+      * A Monolog logger.
```
praneethpr: Change to LoggerInterface

src/Util/MailerLoggerUtil.php
```diff
@@ -0,0 +1,122 @@
...
+      * @var Logger
+      */
+    protected $logger;
+
+    /**
+     * MailerLoggerUtil constructor.
+     *
+     * @param LoggerInterface $logger The Logger Interface.
+     */
+    public function __construct(LoggerInterface $logger)
```
praneethpr: `Psr\Log\LoggerInterface $<channel>Logger` Can be used to specify the channel for the logger and the arguments in the servers.yaml can be removed

PR #496, opened 2020-02-04
Added post update event listener for dataset to update the doi accord…

praneethpr:
…ingly


src/Consumer/DoiConsumer.php
```diff
@@ -123,6 +123,15 @@ public function execute(AMQPMessage $message)
             } else {
                 $this->logger->warning("Unknown routing key: $routingKey", $loggingContext);
             }
+
+            if (preg_match('/^doi/', $routingKey)) {
```
mickel1138: Remove above issue match, because it's throwing a warning of bad routing key. It works fine however.

praneethpr: Fixed it

PR #498, opened 2020-02-05
Feature/pelagos 3736 create fileset entity

praneethpr:


config/packages/doctrine.yaml
```diff
@@ -3,7 +3,7 @@ doctrine:
         # configure these for your database server
         driver: 'pdo_pgsql'
         charset: utf8
-        server_version: 9.6
+        server_version: 12
```
fightingtexasaggie: We probably should leave this at 9.6 as is checked in in code (stash it after making the migration), or will that make the migration not compatible?

praneethpr: Tried it, migration works at 9.6. But we should specify the correct version of postgres in the future

src/Entity/File.php
```diff
@@ -0,0 +1,283 @@
...
+    /**
+     * Doctrine identifier for File entity.
+     *
+     * @var integer
+     *
+     * @ORM\Id()
+     * @ORM\GeneratedValue()
+     * @ORM\Column(type="integer")
+     */
+    protected $id;
```
fightingtexasaggie: Are we overriding the base class Entity's ID?

src/Entity/File.php
```diff
@@ -0,0 +1,283 @@
...
+    /**
+     * File path or location.
+     *
+     * @var string
+     *
+     * @ORM\Column(type="text", nullable=true)
+     */
+    protected $filePath;
+
+    /**
```
fightingtexasaggie: possibly not needed if we're using to base class getter/setter methods

src/Entity/Fileset.php
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Fileset Entity class.
+ *
+ * @ORM\Entity(repositoryClass="App\Repository\FilesetRepository")
+ */
+class Fileset
+{
+
+    /**
```
fightingtexasaggie: base class?

src/Entity/File.php
praneethpr: I created the entity using bin/console make entity command, should've removed this. Thanks Mike

fightingtexasaggie:
I put output of \d file \d fileset \d dataset_submission (trimmed) on the ticket for PO review is there is nothing to test.

src/Entity/File.php
```diff
@@ -0,0 +1,262 @@
...
+{
+    /**
+     * Fileset which contains this file.
+     *
+     * @var Fileset
+     *
+     * @ORM\ManyToOne(targetEntity="Fileset", inversedBy="File")
+     * @ORM\JoinColumn(nullable=false)
+     */
+    protected $fileset;
```
mickel1138: I don't think this relationship is needed, and may complicate things later.

PR #499, opened 2020-02-06
Pelagos 3914 upgrade to symfony 4 4

mickel1138:


mickel1138:
This PR could be ready to merge, but please give it a jolly good test.

composer.json
```diff
@@ -31,20 +31,20 @@
         "php-http/httplug-bundle": "^1.10",
         "sensio/framework-extra-bundle": "^4.0",
         "simplethings/entity-audit-bundle": "^1.0",
-        "symfony/console": "4.3.*",
-        "symfony/dotenv": "4.3.*",
-        "symfony/flex": "^1.4",
-        "symfony/form": "4.3.*",
-        "symfony/framework-bundle": "4.3.*",
-        "symfony/ldap": "4.3.*",
+        "symfony/console": "^4.3",
```
praneethpr: We should probably change all the libraries to 4.4 as well

mickel1138: The up arrow means 4.3 or higher, before we had it set to exactly this version. So we actually have version 4.4.4 installed if you look at the lock file.

mickel1138:
> Why do we have multiple .gitignore files

The .gitignore files are to make sure these folder are persisted when they are empty.

composer.json
praneethpr: Yeah, I know the caret means till <5.0.0, thought we can be specific

mickel1138: Yes, it wouldn't hurt. I only changed those because otherwise it had upgrade conflicts. Even if there is a 4.5 (if) then it would upgrade to it.  I'm ok with whatever looks best?

mickel1138: Actually I take that back, it wouldn't go to 4.5, because in the json file under extra it is constrained to 4.4.

fightingtexasaggie:
The issues I was seeing with a dataset not becoming available were related to my testing environment (needed to restart consumers after switching branches hehe). All is well, tested all major sections including publink and pods. All seems the same.

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -27,23 +29,20 @@ class DatasetSubmissionController extends EntityController
...
-     *     }
-     *   },
-     *   output = "array<Pelagos\Entity\DatasetSubmission>",
-     *   statusCodes = {
-     *     200 = "The requested collection of Dataset Submissions was successfully retrieved.",
-     *     500 = "An internal error has occurred.",
-     *   }
+     * @Operation(
+     *     tags={"Dataset Submission"},
+     *     summary="Get a collection of Dataset Submissions.",
```
praneethpr: I don't know if this ever worked? Dataset submission collection throws an error 
"Could not convert database value \"0001-12-31 18:09:24-05:50:36 BC\" to Doctrine Type datetimetz. Expected format: Y-m-d H:i:sO"
And we aren't using this anywhere in our code base

mickel1138: This is because of a zero date that is being converted to UTC (less time) and makes it a weird date with BC in it. (Before Christ).

PR #500, opened 2020-02-07
Removed AUDIT bundle that we no longer need.

mickel1138:


composer.json
```diff
@@ -30,7 +30,6 @@
         "php-http/guzzle6-adapter": "^1.1",
         "php-http/httplug-bundle": "^1.10",
         "sensio/framework-extra-bundle": "^4.0",
```
praneethpr: Can be removed from composer and it can be added as an annotation, https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/index.html

config/bundles.php
```diff
@@ -13,7 +13,6 @@
     Symfony\Bundle\TwigBundle\TwigBundle::class => ['all' => true],
     Doctrine\Bundle\DoctrineBundle\DoctrineBundle::class => ['all' => true],
```
praneethpr: We aren't using Nelmio Bundle (Api Doc) as well, should be removed

composer.json
```diff
@@ -30,7 +30,6 @@
         "php-http/guzzle6-adapter": "^1.1",
```
praneethpr: Not sure, who requires these libraries php-http/guzzle6-adapter and php-http/httplug-bundle. Doi Util uses guzzlehttp/guzzle bundle for making HTTP API requests

config/bundles.php
mickel1138: I think that's how /api creates the doc?

praneethpr: Yes, I thought no one uses it? and we are on an older version as well

mickel1138: We use it. And we can upgrade to 3.0, but maybe not now. Would rather put effort into upgrading to API Platform.

composer.json
fightingtexasaggie: looks like just the httphotplugbundle which is removed below. :+1:

composer.json
mickel1138: Annotations is an alias of FrameworkExtraBundle, so it's the same.

composer.json
mickel1138: The plug bundle is still there. I believe plug is either recommended or required by guzzle.

PR #42, opened 2018-10-01
Feature/dif modify/pelagos 3324 add unit tests for dif entity

praneethpr:

PR #501, opened 2020-02-07
Changed events and twig extensions to use new class.

mickel1138:


src/Twig/Extensions.php
```diff
@@ -37,7 +40,7 @@ class Extensions extends \Twig_Extension
      */
     public function __construct(KernelInterface $kernel, MaintenanceMode $maintenanceMode)
     {
-        $this->kernelRootDir = $kernel->getRootDir();
+        $this->kernelRootDir = $kernel->getProjectDir();
```
mickel1138: Rootdir is not the same as ProjectDir, so dataland breaks.

fightingtexasaggie: also new search is throwing a 500 now..looking into it

mickel1138: Adding ` . '/src';` would give the same result.

fightingtexasaggie: https://pastebin.com/fHjwy0ju  
Looks like our LogActionItemEvent isn't worknig as an Eventdispatcher Event.

mickel1138: That's just a type hinting issue. It's parent must be updated too.

mickel1138: It was more than that, the dispatch function signature changed.

src/Event/EntityEventDispatcher.php
```diff
@@ -2,7 +2,7 @@
 
 namespace App\Event;
 
-use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
 use App\Entity\Entity;
```
mickel1138: Gotta switch the dispatch method around also on line #42

mickel1138:
> Is src/Event/DatasetIndexSubscriber.php needing to be changed as well?

That stays in the Component namespace for now, as per Symfony instructions.

PR #502, opened 2020-02-10
Modified publink form to validate DOI.

mickel1138:
Also modified backend to reject non-doi.

### DIFF
--- a/public/js/publink.js
+++ b/public/js/publink.js
@@ -3,7 +3,6 @@ var $ = jQuery.noConflict();
 var valid_publication = false;
 var valid_dataset = false;
 var last_retrieved = { dataset: "", publication: "" };
-var doiForm;
 
 $(document).ready(function() {
     jQuery.validator.addMethod("doiFormat", function(value, element) {
--- a/public/js/publink.js
+++ b/public/js/publink.js
@@ -3,14 +3,36 @@ var $ = jQuery.noConflict();
 var valid_publication = false;
 var valid_dataset = false;
 var last_retrieved = { dataset: "", publication: "" };
+var doiForm;
 
 $(document).ready(function() {
-    $('#retrieve_publication').button().click(function () {
-        retrievePublicationCitation();
+    jQuery.validator.addMethod("doiFormat", function(value, element) {
+        var regPattern = /^10.\d{4,9}\/[-._;()\/:A-Z0-9]+$/i
+        return this.optional(element) || (regPattern.test(value));
+    },function (params, element) {
+        return "Please enter a valid DOI"
     });
-    $('#retrieve_dataset').button().click(function () {
-        retrieveDatasetCitation();
+
+    $("#doiForm").validate({
+        submitHandler: function(form) {
+            retrievePublicationCitation();
+        },
+        rules: {
+            doi: {
+                    doiFormat: false
+            }
+        }
     });
+
+    $("#udiForm").validate({
+        submitHandler: function(form) {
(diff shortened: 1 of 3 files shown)


public/js/publink.js
```diff
@@ -5,12 +5,33 @@ var valid_dataset = false;
 var last_retrieved = { dataset: "", publication: "" };
 
 $(document).ready(function() {
-    $('#retrieve_publication').button().click(function () {
-        retrievePublicationCitation();
+    jQuery.validator.addMethod("doiFormat", function(value, element) {
+        var regPattern = /^10.\d{4,9}\/[-._;()\/:A-Z0-9]+$/i
```
mickel1138: This DOI regex is fairly strict. While the one on API is less strict. (https://github.com/griidc/pelagos/pull/502/files#diff-3462843700e70b69066a685d5c23b761R141)

fightingtexasaggie: https://www.doi.org/doi_handbook/2_Numbering.html#2.2.1 Yeah, that regex is way too strict per the DOI specification 2.2.1, "There is no defined limit on the length of the DOI name, or of the DOI prefix or DOI suffix." Also, any printable character is valid. Also "The DOI name is case-insensitive and can incorporate any printable characters from the legal graphic characters of Unicode" so we've got expand allowable characters.

We talked about it once upon a time and agreed on 10.* It think

mickel1138: This regex is suggested by Crossref. Maybe it should be a bit more strict here. But i'm ok with relaxing it to `10\..*`

src/Controller/Api/PublicationController.php
```diff
@@ -138,7 +138,7 @@ public function postAction(Request $request, PubLinkUtil $pubLinkUtil)
 
         $doi = $request->request->get('doi');
 
-        if (false === preg_match('/^10\./', $doi)) {
+        if (false == preg_match('/^10\..*/', $doi)) {
```
praneethpr: is there a reason for removing the type-hint?

public/js/publink.js
```diff
@@ -5,12 +5,33 @@ var valid_dataset = false;
 var last_retrieved = { dataset: "", publication: "" };
 
 $(document).ready(function() {
-    $('#retrieve_publication').button().click(function () {
-        retrievePublicationCitation();
+    jQuery.validator.addMethod("doiFormat", function(value, element) {
```
praneethpr: I think we should stop adding more jQuery stuff, if we are planning to move away from it. We can use vanilla JS and packages for it, for validation we can use this alternative
https://validatejs.org/

fightingtexasaggie: at some point, this is going to be a big epic

src/Controller/Api/PublicationController.php
fightingtexasaggie: https://www.php.net/manual/en/function.preg-match.php -  PHPs nastiness - possible return of different types that both evaluate to false.

mickel1138: It's only false if an error occurred. That seems unlikely to happen.
https://github.com/griidc/pelagos/commit/29af7783e0642b0b21e1e3f0edb4a0d9d0fc377b#diff-98d3078053a7ffd5bb3afaf91ae5dc5fL58-R58

Example: https://3v4l.org/UK5Mt

public/js/publink.js
mickel1138: I'm all for no more jQuery, html 5 input have `pattern` property.

PR #508, opened 2020-02-14
Misc/upgrade nelmio doc

praneethpr:


praneethpr:
I did a good end to end test for this branch, this could be merged to Symfony 4.4 upgrade. If we do it at another time, we should do an end-to end test again because these changes affect the APIs @mickel1138 @fightingtexasaggie

composer.json
```diff
@@ -41,11 +42,12 @@
         "symfony/security-bundle": "4.4.*",
         "symfony/swiftmailer-bundle": "^3.1",
         "symfony/templating": "4.4.*",
+        "symfony/twig-pack": "^1.0",
         "symfony/validator": "4.4.*",
         "symfony/webpack-encore-bundle": "^1.1",
         "symfony/yaml": "4.4.*",
         "twig/extensions": "^1.5",
-        "willdurand/hateoas-bundle": "^1.1"
+        "willdurand/hateoas-bundle": "^2.1"
```
mickel1138: Maybe make this 2.0, so we know that we specifically needed 2.1 (or higher)

config/packages/nelmio_api_doc.yaml
```diff
@@ -1,22 +1,9 @@
...
-#            license:              null
-#            licenseUrl:           null
-#    cache:
-#        enabled:              false
-#        file:                 '%kernel.cache_dir%/api-doc.cache'
+nelmio_api_doc:
+    documentation:
+        info:
+            title: Pelagos
+            description: This is an awesome app!
```
mickel1138: Not sure what the description should be, or the version?

config/routes/nelmio_api_doc.yaml
```diff
@@ -1,3 +1,12 @@
...
-    resource: "@NelmioApiDocBundle/Resources/config/routing.yml"
-    prefix:   /api/doc
+# Expose your documentation as JSON swagger compliant
+app.swagger:
+    path: /api/doc.json
+    methods: GET
+    defaults: { _controller: nelmio_api_doc.controller.swagger }
+
+## Requires the Asset component and the Twig bundle
+## $ composer require twig asset
```
mickel1138: Should we require this (twig) separately? It's already installed because of other packages.

mickel1138: Actually, I see it has been added.

composer.json
praneethpr: ^2.1 does mean that we specifically needed 2.1 or higher right?

config/packages/nelmio_api_doc.yaml
praneethpr: we can start with 1.0.0 for version and its anyways for internal viewing

config/routes/nelmio_api_doc.yaml
praneethpr: This installs a different package called symfony/twig-pack, I think its different from the twig package we use?

composer.json
mickel1138: Yes, it does mean that. But this makes me thing that 2.0 was no good. What is there is no 2.1? It was more of a type-o I think.

config/packages/nelmio_api_doc.yaml
mickel1138: The API is public, some people may make use of it. However, we don't publish it the URL publicly I thnk.

config/routes/nelmio_api_doc.yaml
mickel1138: The symfony pack is just a bundle/package of commonly used bundles/packages that Symfony uses when using twig.

PR #509, opened 2020-02-17
Fixed citation year at dataset entity level

praneethpr:


src/Entity/Dataset.php
```diff
@@ -573,7 +573,7 @@ public function getCitation()
         $title = preg_replace('/\.$/', '', $title);
         $udi = $this->getUdi();
         $author = $this->getAuthors();
-        $year = $this->getReferenceDateYear();
+        $year = ($this->getAcceptedDate()) ? $this->getAcceptedDate()->format('Y') : null;
```
mickel1138: Maybe add instanceOf ?

praneethpr: Maybe we should wait till we decide on what dates to use, and we can probably make changes to all affecting applications in this ticket

PR #510, opened 2020-02-17
Feature/pelagos 3922 modify udi minter to use new udi list

mickel1138:


src/Entity/Udi.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->uniqueDataIdentifier = $uniqueDataIdentifier;
+    }
+
+    /**
+     * A Pretty Print sort of formatting in the string context.
+     *
+     * @return string The formatted UDI string.
+     */
+    public function __toString(): string
```
praneethpr: If this property is declared as a text, why do we need a toString() method

src/Entity/Udi.php
```diff
@@ -0,0 +1,51 @@
...
+    protected $uniqueDataIdentifier;
+
+    /**
+     * Constructor.
+     *
+     * @param string $uniqueDataIdentifier The UDI string.
+     *
+     * Sets to UDI identifier for the UDI entity.
+     */
+    public function __construct(string $uniqueDataIdentifier): void
```
praneethpr: I don't think constructors are allowed to have a return type

src/Entity/Udi.php
mickel1138: Or else there is not getter for this. Which be may not need anyway. But it's a just in case. Or if we would like to improve this entity with some pattern matching.

src/Entity/Udi.php
mickel1138: It doesn't seem to break, but it's also not needed.

mickel1138: It broke! I changed it after I tested. Gonna fix.

src/Entity/Udi.php
mickel1138: So I actually need this. Instead of a getter. It's used in the utility, line 67.

src/Util/Udi.php
```diff
@@ -50,26 +51,31 @@ public function mintUdi(Dataset $dataset)
...
+            ->where('u.uniqueDataIdentifier LIKE :udi')
+            ->orderBy('u.uniqueDataIdentifier', 'DESC')
             ->setParameter('udi', "$udi%")
             ->getQuery();
 
-        $datasets = $query->getResult();
+        $udis = $query->getResult();
 
-        if (count($datasets) == 0) {
+        if (count($udis) == 0) {
```
praneethpr: can add type comparison

src/Entity/Udi.php
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @ORM\Entity(repositoryClass="App\Repository\UdiRepository")
+ *
+ * @UniqueEntity(
+ *     fields={"uniqueDataIdentifier"},
+ *     errorPath="uniqueDataIdentifier",
+ *     message="This UDI already exists"
+ * )
+ */
+class Udi extends Entity
```
mickel1138: This really doesn't need to extend Entity. There is no need for modifier, on modified/created dates?
What do you guys think? @fightingtexasaggie @praneethpr

fightingtexasaggie: As we talked about at stand up, I'm in agreement we don't need it. Would we have to add an ID to the class?

mickel1138: No, not needed. UDI can be the identifier.

praneethpr: I think its better to have ID to the class, because UDI isnt a good identifier

mickel1138: UDI will be unique, it's just a list.

src/Util/Udi.php
```diff
@@ -50,26 +51,48 @@ public function mintUdi(Dataset $dataset)
...
+                ->getResult();
 
-        if (count($datasets) == 0) {
+        if (count($datasets) === 0 and count($udis) === 0) {
             // If this is the first dataset for this Research Group, we start at 1.
             $sequence = 1;
         } else {
-            // Grab the sequence from the UID.
+            // Find the latest dataset submitted.
             preg_match('/:(\d{4})$/', $datasets[0]->getUdi(), $matches);
```
praneethpr: Do we need to add the same count($datasets) here as well?

praneethpr: But I dont think a scenario would exist

src/Util/Udi.php
```diff
@@ -66,30 +66,26 @@ public function mintUdi(Dataset $dataset)
...
-            $sequence = ($lastSequence + 1);
+        if (count($udis) !== 0) {
+            // Grab the last sequence from the UDI list.
+            preg_match('/:(\d{4})$/', $udis[0], $matches);
+            $lastUdiSequence = intval($matches[1]);
+        } else {
+            $lastUdiSequence = 0;
         }
+
+        $lastSequence = max($lastDatasetSequence, $lastUdiSequence);
```
fightingtexasaggie: :sunglasses:

PR #511, opened 2020-02-18
Added dispatch event for put action if save and continue later is used

praneethpr:


mickel1138:
This is not correct, now every time you click save (draft), it will send an e-mail. Have to discuss with PO.

PR #514, opened 2020-02-21
Feature/pelagos 3926 add checks to remotely hosted app

fightingtexasaggie:


src/Controller/UI/RemotelyHostedDatasetsController.php
```diff
@@ -155,4 +158,48 @@ private function dispatchLogEvent(Dataset $dataset, string $actor)
...
     }
+
+    /**
+     * Check to see if the remotely-hosted fields are all populated.
+     *
+     * @param DatasetSubmission $datasetSubmission A DatasetSubmission.
+     *
+     * @return array $errors Containing strings of each unpopulated attribute, or empty array for compliant.
+     */
+    public function remotelyHostedPrerequsiteCheck(DatasetSubmission$datasetSubmission) : array
```
mickel1138: need space?

src/Controller/UI/RemotelyHostedDatasetsController.php
```diff
@@ -155,4 +158,48 @@ private function dispatchLogEvent(Dataset $dataset, string $actor)
...
+    {
+        $errors = array();
+
+        if (!$datasetSubmission instanceof DatasetSubmission) {
```
praneethpr: I feel this check should be placed in the previous function, because if suppose DatasetSubmission object doesn't exist and this function is called, PHP would throw an error for incorrect parameter type

src/Controller/UI/RemotelyHostedDatasetsController.php
```diff
@@ -155,4 +162,45 @@ private function dispatchLogEvent(Dataset $dataset, string $actor)
...
     }
+
+    /**
+     * Check to see if the remotely-hosted fields are all populated.
+     *
+     * @param DatasetSubmission $datasetSubmission A DatasetSubmission.
+     *
+     * @return array $errors Containing strings of each unpopulated attribute, or empty array for compliant.
+     */
+    public function remotelyHostedPrerequsiteCheck(DatasetSubmission$datasetSubmission): array
```
mickel1138: No space between typehint and parameter?

fightingtexasaggie: and yet, somehow that worked! hehe oops

src/Controller/UI/RemotelyHostedDatasetsController.php
```diff
@@ -84,20 +85,26 @@ public function postAction(Request $request)
...
             $datasetStatus = $dataset->getDatasetStatus();
-
-            if ($datasetStatus === Dataset::DATASET_STATUS_ACCEPTED) {
-                if (DatasetSubmission::TRANSFER_STATUS_REMOTELY_HOSTED !== $datasetSubmission->getDatasetFileTransferStatus()) {
-                    $datasetSubmission->setDatasetFileTransferStatus(DatasetSubmission::TRANSFER_STATUS_REMOTELY_HOSTED);
-
-                    $this->dispatchLogEvent($dataset, $this->getUser()->getUserId());
-                    return new Response('Dataset UDI ' . $udi . ' has been successfully set to remotely hosted.', Response::HTTP_OK);
+            $datasetSubmission = $dataset->getDatasetSubmission();
+            if (null !== $datasetSubmission) {
```
mickel1138: InstanceOf would work better here I think.

mickel1138:
Works, but the Noty warning only shows for a few seconds, maybe this should be a model error. That stays up.

src/Controller/UI/RemotelyHostedDatasetsController.php
praneethpr: Yeah I agree

public/js/remotelyHostedDatasets.js
```diff
@@ -25,28 +25,57 @@ $(document).ready(function(){
             url: Routing.generate("pelagos_app_ui_remotelyhosteddatasets_post", {udi : $("#udiInput").val().trim()}),
         }).done(function(data, textStatus, jqXHR){
             var messageType = "success";
+            var modalBool = false;
             //return informative message for 202 code
             if (202 === jqXHR.status) {
+                console.log(data);
```
mickel1138: delete debug code

public/js/remotelyHostedDatasets.js
```diff
@@ -25,28 +25,57 @@ $(document).ready(function(){
...
                 messageType = "warning";
+                if (data.includes("CONFIRM:")) {
+                    data = data.replace("CONFIRM:","");
+                    var confirmBool = true;
+                }
             } else {
                 //reset table
                 $("#remotelyHostedDatasetsTable").pelagosDataTable();
             }
-            var n = noty(
```
mickel1138: Understand you are passing JSON, these are 90% the same, so you can make a base json, then add or remove properties based on the condition.

public/js/remotelyHostedDatasets.js
```diff
@@ -25,28 +25,57 @@ $(document).ready(function(){
...
+                    type: messageType,
+                    text: data,
+                    modal: true,
+                    animation: {
+                        open: "animated fadeIn", // Animate.css class names
+                        close: "animated fadeOut", // Animate.css class names
+                        easing: "swing", // unavailable - no need
+                        speed: 500 // unavailable - no need
+                    },
+                    buttons: [
```
mickel1138: Buttons are not needed, click will close it. (Rather than timeout).

src/Controller/UI/RemotelyHostedDatasetsController.php
```diff
@@ -90,7 +90,7 @@ public function postAction(Request $request)
             if ($datasetSubmission instanceof DatasetSubmission) {
                 $preReqs = $this->remotelyHostedPrerequsiteCheck($datasetSubmission);
                 if (count($preReqs) > 0) {
-                    $message = implode(', ', $preReqs);
+                    $message = 'CONFIRM:' . implode(', ', $preReqs);
```
mickel1138: This is not a great way to do this. Maybe change the response HTTP code on line L113 to something else, 400 ish or so.

mickel1138:
> Works fine. In my opinion, we could've used Vex-js modal box instead of Noty (like Dataland download box)

If changes the noty if ugly, I say we just undo that part, and move it along the way it is. We can make improvements later on the suggestion of the PO.

PR #519, opened 2020-02-26
Refactor/additional deprecated

mickel1138:


config/packages/doctrine.yaml
```diff
@@ -13,7 +13,7 @@ doctrine:
                 commented: false
     orm:
         auto_generate_proxy_classes: true
-        naming_strategy: doctrine.orm.naming_strategy.underscore
+        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
```
praneethpr: Cannot create DIF, throws a doctrine error : column doesn't exist

composer.json
```diff
@@ -15,18 +15,18 @@
         "doctrine/orm": "^2.5",
         "friendsofsymfony/elastica-bundle": "^5.1.1",
         "friendsofsymfony/jsrouting-bundle": "^2.1",
-        "friendsofsymfony/rest-bundle": "^2.5",
+        "friendsofsymfony/rest-bundle": "dev-master",
```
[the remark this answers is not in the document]
mickel1138: It is. As soon as the new version comes out, we should change it.

config/packages/doctrine.yaml
mickel1138: May need to run a migration.

mickel1138: It did need a migration.

composer.json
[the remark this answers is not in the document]
fightingtexasaggie: If we really have to do this, we should put it on a particular commit instead, at least it won't be a moving target.

config/packages/prod/monolog.yaml
```diff
@@ -5,6 +5,7 @@ monolog:
             action_level: error
             handler: nested
             excluded_http_codes: [404, 405]
+            buffer_size: 50 # How many messages should be saved? Prevent memory leaks
```
fightingtexasaggie: Isn't this way, way way too few? maybe 50,000 messages might be more reasonable. Is this really a memory leak issue or a drive space thing with runaway logs?

src/Migrations/Version20200302174819.php
```diff
@@ -0,0 +1,43 @@
+<?php declare(strict_types=1);
```
fightingtexasaggie: Why is this migration in here?

config/packages/prod/monolog.yaml
mickel1138: This is how it's set by Symfony, not my doing.

src/Migrations/Version20200302174819.php
```diff
@@ -0,0 +1,43 @@
...
+final class Version20200302174819 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('DROP SEQUENCE revisions_id_seq CASCADE');
+        $this->addSql('DROP TABLE revisions');
+        $this->addSql('ALTER TABLE dataset_submission ADD dataset_file_sha256_hash TEXT DEFAULT NULL');
```
mickel1138: Use: `ALTER TABLE dataset_submission RENAME COLUMN dataset_file_sha256hashTO dataset_file_sha256_hash ;`

fightingtexasaggie: as discussed in chat, reminder to rename instead of add/drop

composer.json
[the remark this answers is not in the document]
mickel1138: took it out for now

mickel1138:
I removed the FOS/rest upgrade, in time the fix will be released, we won't have to make any changes.
This can be merged if desired.

config/packages/prod/monolog.yaml
mickel1138: So,  buffer size is just that, it's not a limit of messages. It will write keep 50 messages in it's buffer, then write. It is not a limitation of how many messages.

src/Migrations/Version20200302174819.php
```diff
@@ -0,0 +1,38 @@
...
+        $this->addSql('ALTER TABLE dataset_submission RENAME COLUMN dataset_file_cold_storage_archive_sha256hash TO dataset_file_cold_storage_archive_sha256_hash');
+        $this->addSql('ALTER TABLE dataset_submission RENAME COLUMN metadata_file_sha256hash TO metadata_file_sha256_hash');
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
```
mickel1138: Remove

PR #520, opened 2020-02-27
Feature/vue search app

praneethpr:


assets/scss/bootstrap.scss
```diff
@@ -1,3 +1,9 @@
-div.bootstrap {
-    @import "~bootstrap";
+div.bootstrap, .tooltip {
```
mickel1138: This means tooltip class gets bootstrap applied, is that really needed?

praneethpr: The tooltip div spawns outside the parent div element by design, and none of the styles get applied to it

mickel1138:
Looks and works great, would like a explanation of how Vue works here.

PR #523, opened 2020-03-04
Feature/pelagos 3934 create dataset links entity

mickel1138:


src/Entity/DatasetLinks.php
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /**
+     * Set the function code for the Dataset Link.
+     *
+     * @param string $functionCode The function code for the Dataset Link.
+     *
+     * @return self
+     */
+    public function setFunctionCode(string $functionCode): self
```
fightingtexasaggie: Should we check that the setter only accept values in the const array above?

src/Migrations/Version20200304165655.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->addSql('ALTER TABLE dataset_links ADD CONSTRAINT FK_A16DD1C261220EA6 FOREIGN KEY (creator_id) REFERENCES person (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('ALTER TABLE dataset_links ADD CONSTRAINT FK_A16DD1C2D079F553 FOREIGN KEY (modifier_id) REFERENCES person (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
```
fightingtexasaggie: can remove this extra line that gets generated - CREATE SCHEMA public

praneethpr: Good catch! I think we let this slip in the UDI Entity migration, we should remove this line in that too

PR #525, opened 2020-03-05
Feature/pelagos 3939 create forms and ui

mickel1138:


mickel1138:
Needs to have unit tests fixed.

src/Entity/DatasetLink.php
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /**
+     * Set the function code for the Dataset Link.
+     *
+     * @param string $functionCode The function code for the Dataset Link.
+     *
+     * @return self
+     */
+    public function setFunctionCode(?string $functionCode): self
```
praneethpr: Same here

src/Entity/DatasetLink.php
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /**
+     * Set the dataset submission for this link.
+     *
+     * @param DatasetSubmission|null $datasetSubmission The dataset submission for this link.
+     *
+     * @return self
+     */
+    public function setDatasetSubmission(?DatasetSubmission $datasetSubmission): self
```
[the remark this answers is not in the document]
mickel1138: This is what the maker bundle did. I questioned it myself at first too, but I think it's so you can do things like this:

`$myObject = $myObject->changeItem($item);`

I guess that can be useful is certain situations, but if you think we should do things they same everywhere, then that's ok too. Writing the test is harder now.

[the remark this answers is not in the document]
mickel1138: I think this also help with New-ing an Class.

src/Entity/DatasetSubmission.php
```diff
@@ -1155,6 +1166,18 @@ public function __construct(Entity $entity, PersonDatasetSubmissionDatasetContac
                 $newDistributionPoint->setDataCenter($distributionPoint->getDataCenter());
                 $this->addDistributionPoint($newDistributionPoint);
             }
+
+            // Copy the original Dataset Submission's Dataset Links.
+            foreach ($entity->getDatasetLinks() as $datasetLink) {
+                $newDatasetLink = new DatasetLink();
+                $newDatasetLink->setUrl($datasetLink->getUrl());
```
praneethpr: If we are going to use return types for setters, then we should use method chaining as well! I think if aren't going to use method chaining, we shouldn't use return type for setters.

For example:

`class fakeString
{
    private $str;
    function __construct()
    {
        $this->str = "";
    }

    function addA()
    {
        $this->str .= "a";
        return $this;
    }

    function addB()
    {
        $this->str .= "b";
        return $this;
    }

    function getStr()
    {
        return $this->str;
    }
}


$a = new fakeString();


echo $a->addA()->addB()->getStr();`

mickel1138: How would you improve this particular scenario?

praneethpr:
````
$newDatasetLink->setUrl($datasetLink->getUrl())->setName($datasetLink->getName())->setDescription($datasetLink->getDescription()) ......
````
We can do this, but I feel chaining methods like this is tough to understand

File: public/js/datasetReview.js [hunk 4]
```diff
@@ -114,6 +114,41 @@ $(document).ready(function(){
...
+        var value = parseFloat($(this).attr("index"));
+        datasetLinksCount = (value > datasetLinksCount) ? value : datasetLinksCount;
+    });
+
+    $("#addLink")
+        .button()
+        .click(function(){
+            datasetLinksCount++;
+
+            var newLink = $("#links-prototype table")
```
praneethpr: We should validate the Url using our API. Its done with the ERDDAP field

src/Form/DatasetLinkType.php
```diff
@@ -0,0 +1,71 @@
...
+     * @return void
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('url', Type\TextType::class, array(
+                'label' => 'Link Url:',
+                'required' => false,
+            ))
+            ->add('name', Type\TextType::class, array(
```
praneethpr: We can try using Datalist here, and the ticket describes this to be a predefined list.

src/Entity/DatasetLink.php
fightingtexasaggie: I didn't see your prior comment referred to, but would suggest checking that the past code is in the list of constants above.

src/Entity/DatasetSubmission.php
```diff
@@ -1028,6 +1028,15 @@ class DatasetSubmission extends Entity
...
     protected $remotelyHostedFunction;
 
+    /**
+     * The Point of Contact for this Dataset Submission.
+     *
+     * @var Collection
+     *
+     * @ORM\OneToMany(targetEntity="App\Entity\DatasetLink", mappedBy="datasetSubmission", cascade={"persist"}, orphanRemoval=true)
+     */
+    private $datasetLinks;
```
fightingtexasaggie: private vs protected?

src/Entity/DatasetSubmission.php
fightingtexasaggie: I'm actually ok with them too, but think we should decide one way or another and stick with it.

src/Entity/DatasetSubmission.php
mickel1138: This was added by the Maker bundle, but changed it.

templates/DatasetReview/dataset-links.html.twig
```diff
@@ -0,0 +1,140 @@
...
+            </tr>
+        </table>
+        </fieldset>
+        {% endfor %}
+    </div>
+
+    <button id="addLink" type="button">Add Link</button>
+</fieldset>
+
+<datalist id="dataLinkNames">
```
praneethpr: Should we also add Dataland ? It is mentioned in the ticket
I think this should be a list of constants in the DatasetLink Entity

src/Form/DatasetLinkType.php
```diff
@@ -0,0 +1,69 @@
...
+     * @param array                $options The options.
+     *
+     * @see FormTypeExtensionInterface::buildForm()
+     *
+     * @return void
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('url', Type\TextType::class, array(
```
praneethpr: It might be a question for the PO, Should these fields be required?

PR #527, opened 2020-03-16
Added identified status in the elastic config to not show all dataset…

praneethpr:
…s which do not have approved Difs

### DIFF
--- a/config/packages/fos_elastica.yaml
+++ b/config/packages/fos_elastica.yaml
@@ -120,6 +120,8 @@ fos_elastica:
                             properties:
                                 doi:
                                     type: 'text'
+                        identifiedStatus:
+                            type: 'integer'
                     persistence:
                         driver: orm
                         model: App\Entity\Dataset
--- a/src/Util/Search.php
+++ b/src/Util/Search.php
@@ -762,7 +762,9 @@ class Search
                 $subMainQuery->addMust($this->getFieldsQuery($queryTerm, $specificField, $collectionDateRange));
             }
         } else {
-            $subMainQuery->addMust(new Query\MatchAll());
+            $allDatasetsQuery = new Query\Term();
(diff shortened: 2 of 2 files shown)


mickel1138:
Is there a way to only index Identified?

praneethpr:
> Is there a way to only index Identified?

We should add a check in the DatasetIndexSubscriber class, to only index datasets which are identified. But we use the same Class to populate both data-discovery and search index

mickel1138:
> > Is there a way to only index Identified?
> 
> We should add a check in the DatasetIndexSubscriber class, to only index datasets which are identified. But we use the same Class to populate both data-discovery and search index

I looked around a bit, and you could use an Analyzer with a filter. But I think that's more complicated to configure.

praneethpr:
> > > Is there a way to only index Identified?
> > 
> > 
> > We should add a check in the DatasetIndexSubscriber class, to only index datasets which are identified. But we use the same Class to populate both data-discovery and search index
> 
> I looked around a bit, and you could use an Analyzer with a filter. But I think that's more complicated to configure.

I think using a term filter is the simplest solution

praneethpr:
> An exception has been thrown during the rendering of a template ("Could not find the entry "app" in "/home/mwilliamson/pelagos/public/build/entrypoints.json". Found: entrypoints.").
> 
> is this just my environment, or can anyone else confirm?

Did you clear the cache? I'm unable to reproduce this error

mickel1138:
> > An exception has been thrown during the rendering of a template ("Could not find the entry "app" in "/home/mwilliamson/pelagos/public/build/entrypoints.json". Found: entrypoints.").
> > is this just my environment, or can anyone else confirm?
> 
> Did you clear the cache? I'm unable to reproduce this error

That's usually when you forgot `yarn dev`

PR #528, opened 2020-03-17
Error caused due to missing use statement.

mickel1138:


mickel1138:
The exception wasn't caught on line 311 https://github.com/griidc/pelagos/pull/528/files#diff-769ce891dc823d63b488bf0b070b185eR311

PR #530, opened 2020-03-18
Bugfix/pelagos 3945 enable browser back forward button for new search

praneethpr:


praneethpr:
> Almost perfect - searched for 'water' then selected the checkbox for 'available'. Clicking back showed the number of results without the 'available' contraint, but it remained visible checked.

Good catch! I was trying with 2 checkboxes checked, it works for that. Didnt try this scenario 👍

praneethpr:
> Almost perfect - searched for 'water' then selected the checkbox for 'available'. Clicking back showed the number of results without the 'available' contraint, but it remained visible checked.

Fixed it

assets/js/vue/components/search/SearchForm.vue
```diff
@@ -153,14 +153,21 @@
                 this.form.page = newPageNo;
                 this.onSubmit();
             },
+            detectHashChange: function () {
+                this.route = window.location.hash;
+                if (this.route) {
+                    const urlHashSplit = this.route.split("#")[1].split("&").map(value => value.split("="));
+                    this.form = Object.fromEntries(urlHashSplit);
```
mickel1138: If there are no entries, it won't work. For instance: start with a clean search, then search for something, then go back. (Basically a blank page) and it won't work.

mickel1138:
Don't know where this is coming from, what when you search for words, add another word, it will add %20 (coming from the url) in the search bar.

assets/js/vue/components/search/SearchForm.vue
praneethpr: Will fix it, didn't think about this scenario

praneethpr:
> Don't know where this is coming from, what when you search for words, add another word, it will add %20 (coming from the url) in the search bar.

I think we should decode the url, because it is populating the form input box from the hash

praneethpr:
> Don't know where this is coming from, what when you search for words, add another word, it will add %20 (coming from the url) in the search bar.

Fixed it

assets/js/vue/components/search/SearchForm.vue
praneethpr: Fixed it

mickel1138:
Back still shows %20...
This is what I did:
* search for fish (go back, go forward) (works)
* add oil (got back and forward) (back is ok, forward adds %20

praneethpr:
> Back still shows %20...
> This is what I did:
> 
> * search for fish (go back, go forward) (works)
> * add oil (got back and forward) (back is ok, forward adds %20

Forgot to add the decode URI, Fixed it now

assets/js/vue/components/search/SearchForm.vue
```diff
@@ -173,7 +173,7 @@
             route: function (value) {
                 if (!this.submitted) {
                     if (this.route) {
-                        const urlHashSplit = this.route.split("#")[1].split("&").map(value => value.split("="));
+                        const urlHashSplit = decodeURI(this.route).split("#")[1].split("&").map(value => value.split("="));
```
mickel1138: Why const, and not `let`? Or `var`? is this a vue thing?

praneethpr: Its just a regular Javascript constant, not a vue thing. If the value of variable isn't going to be reassigned we should use `const`. `const` and `let` are pretty much the same except that it can't be reassigned after it is declared

PR #534, opened 2020-03-23
Removed bootstrap tooltip and v-tooltip library which is more reactiv…

praneethpr:
…e to mouse hover


praneethpr:
> Works, but only TOP placement looks good, or even shows.
> Also a lot of additional CSS, and will start to look messy/hard to maintain.
> If we can help we only wanna load CSS supplied by the Library.

Yes, I dint want to add custom CSS but didn't find a better option. We can look into other wrapper libraries which use popper.js that have built-in CSS for the tooltip

PR #536, opened 2020-03-25
Feature/pelagos 3966 standardize collection date formats to use utc

praneethpr:


praneethpr:
> While everything looks logical (using the same offset in and out effectively), when conducting a date range search and no other criteria, I get nothing for the current year. Going back a few years, I only get a few hundred.

Did you do a fos elastica reset and populate?

src/Event/DatasetIndexSubscriber.php
```diff
@@ -103,8 +103,18 @@ public function populateCalculatedFields(TransformEvent $event)
 
         if ($dataset->hasDatasetSubmission()) {
             if ($dataset->getDatasetSubmission()->getTemporalExtentBeginPosition() and $dataset->getDatasetSubmission()->getTemporalExtentEndPosition()) {
-                $document->set('collectionStartDate', $dataset->getDatasetSubmission()->getTemporalExtentBeginPosition()->format('Y-m-d'));
-                $document->set('collectionEndDate', $dataset->getDatasetSubmission()->getTemporalExtentEndPosition()->format('Y-m-d'));
+                $collectionStartDate = new \DateTime(
+                    $dataset->getDatasetSubmission()->getTemporalExtentBeginPosition()->format('Y-m-d H:i:s'),
```
mickel1138: You probably don't have to format it here?

praneethpr: TemporalBeginPosition timestamp is stored in that format along with a timezone, if i don't get it in that format and convert it to UTC here, there will be discrepancies in search results and dataland

PR #537, opened 2020-03-25
Removed hyperlink associated class and text change

praneethpr:


praneethpr:
> hyperlink is removed on the default card. Ticket critera met.
> 
> I think we should consider removing it from the results cards too, and maybe make a specific link to dataland instead of a clickable card because it's nearly impossible to copy/paste text from the results. (food for thought anyway)

William wanted the whole result row to be a clickable item, but we can ask the PO's again

PR #538, opened 2020-03-26
Added dataset links on dataland.

mickel1138:


templates/Dataland/datasetlinks.html.twig
```diff
@@ -0,0 +1,14 @@
+<h2>Dataset Links:</h2>
+<div>
+    {% for link in dataset.datasetSubmission.datasetLinks %}
+        <p>
+            <b>Name:</b>{{ link.name }}<br>
+            <b>Description:</b>{{ link.description }}<br>
+            <b>Function:</b>{{ link.functionCode }}<br>
```
fightingtexasaggie: we just showing the raw function code here, or turning it into it's textual representation?

templates/Dataland/datasetlinks.html.twig
```diff
@@ -0,0 +1,14 @@
+<h2>Dataset Links:</h2>
+<div>
+    {% for link in dataset.datasetSubmission.datasetLinks %}
+        <p>
+            <b>Name:</b>{{ link.name }}<br>
+            <b>Description:</b>{{ link.description }}<br>
+            <b>Function:</b>{{ link.functionCode }}<br>
+            <b>URL:</b>
+            <a href="{{ link.url }}" target="_blank">
```
fightingtexasaggie: Should we decompose the link and strip out any included protocol, then add the protocol + link?

templates/Dataland/datasetlinks.html.twig
mickel1138: Was working on that, now fixed.

templates/Dataland/datasetlinks.html.twig
mickel1138: No, the protocol of the link may not be the same as the protocol. It's just a thing.

templates/Dataland/datasetlinks.html.twig
```diff
@@ -0,0 +1,15 @@
+<h2>Dataset Links:</h2>
+<div>
+    {% set linkcodes = constant('App\\Entity\\DatasetLink::ONLINE_FUNCTION_CODES') %}
+    {% for link in dataset.datasetSubmission.datasetLinks %}
+        <p>
+            <b>Name:</b>{{ link.name }}<br>
```
praneethpr: Need to add space for all the elements

templates/Dataland/datasetlinks.html.twig
```diff
@@ -0,0 +1,15 @@
+<h2>Dataset Links:</h2>
+<div>
+    {% set linkcodes = constant('App\\Entity\\DatasetLink::ONLINE_FUNCTION_CODES') %}
+    {% for link in dataset.datasetSubmission.datasetLinks %}
+        <p>
+            <b>Name:</b>{{ link.name }}<br>
+            <b>Description:</b>{{ link.description }}<br>
+            <b>Function:</b>{{ attribute(linkcodes[link.functionCode], 'name') }}<br>
```
praneethpr: Should we add the same style as summary tab?

mickel1138: Which style do you suggest? I tried h2, didn't look right. And the new lines use too much space I think.

praneethpr: We can ask the PO because it is a public facing feature

PR #545, opened 2020-04-01
Bugfix/pelagos 3923 additional upgrades

mickel1138:


src/Util/DataStore.php
```diff
@@ -140,17 +142,17 @@ public function addFile(string $fileUri, string $datasetId, string $type)
         if (preg_match('/^http/', $fileUri)) {
             // Decode any characters escaped in the URL.
             $fileName = urldecode($fileName);
-            $browser = new \Buzz\Browser();
-            $result = $browser->head($fileUri);
-            $status = $result->getHeaders()[0];
-            if (!preg_match('/200/', $status)) {
+            $client = new GuzzleClient();
+            $result = $client->request('HEAD', $fileUri);
```
fightingtexasaggie, line 146: if HEAD method is not supported by the server, what does request return?

src/Util/DataStore.php
```diff
@@ -140,17 +142,17 @@ public function addFile(string $fileUri, string $datasetId, string $type)
...
+            if (200 !== $status) {
                 throw new \Exception("File could not be downloaded from $fileUri ($status)");
             }
-            $contentType = $result->getHeader('Content-Type');
+            $contentType = $result->getHeader('Content-Type')[0];
             if (preg_match('#^text/html#', $contentType)) {
                 throw new HtmlFoundException("HTML file found at $fileUri");
             }
-            $contentDisposition = $result->getHeader('Content-Disposition');
+            $contentDisposition = $result->getHeader('Content-Disposition')[0];
```
fightingtexasaggie, line 155: This one didn't exist at https://www.tamucc.edu/robots.txt
For available headers, see curl -I https://www.tamucc.edu/robots.txt

src/Util/DataStore.php
mickel1138: HEAD is just the header, so anyone should support that.

PR #546, opened 2020-04-02
Feature/pelagos 3980 create switching template and global twig var and use for base template in apps

praneethpr:


templates/nas-grp-base.html.twig
```diff
@@ -0,0 +1,150 @@
...
+                                    <a href="{{ path('security_login', {destination: app.request.uri}) }}">Log in</a>
+                                {% endif %}
+                            </div>
+                        </div>
+                    </div>
+
+                    <div id="branding" class="container">
+                        <div id="logo-wrapper">
+                            <a href="http://gulfresearchinitiative.org" title="National Academies | Gulf Research Program" rel="home" id="logo">
+                                <img src="{{asset('images/nas-grp-logo.png')}}" alt="Home" style="width: 233px; height: 80px">
```
mickel1138, line 75: This image is a bit squeezed.

templates/nas-grp-base.html.twig
```diff
@@ -0,0 +1,150 @@
...
+                            </a>
+                        </div>
+
+                        <div id="name-and-slogan">
+
+                            <div id="site-slogan">Investigating the effect of oil spills <br> on the environment and public health.</div>
```
mickel1138, line 81: This slogan has to be changed or removed.

templates/nas-grp-base.html.twig
praneethpr: Do you want to merge it to your branch? I think you've fixed styling right

PR #549, opened 2020-04-06
Upgraded minimist to 1.2.5

mickel1138:
If we like, we can do a --deep upgrade, and update all indirect libraries.


mickel1138:
> Should we instead make this minimum version required in composer.json?

This is yarn, not composer.

PR #555, opened 2020-04-08
Moved colors up to the template from each app css

praneethpr:


assets/css/griidc.css
```diff
@@ -30,15 +30,15 @@
     --color-black:#000;
     --color-gradientStart: #022f3a;
     --color-gradientEnd: #4f6e74;
+    --color-transperantGreen: #ddffdd;
```
mickel1138, line 33: TranspArant

assets/css/griidc-app.css
```diff
@@ -29,4 +29,21 @@
...
+table.datasets tr:hover td {
+    background-color: var(--color-table-striped);
+}
+
+table.datasets tr td.highlight {
+    background-color:  var(--color-table-striped);
+}
+
+.datasets tr:nth-child(odd) {
+    background-color: #f6f5f4;
```
mickel1138, line 43: Is this a relevant color to be named?

praneethpr: I found only one instance of this

public/css/data-discovery-datasets.css
```diff
@@ -7,14 +7,6 @@ table.datasets td {
     padding: 5px 10px;
 }
 
-table.datasets tr:hover td {
```
mickel1138, line 10: In this case, I think the CSS should stay here, just change the color name to something like: `--dataset-table-hover-color' and set that in each theme template.

mickel1138:
Maybe it's me, but data discovery is not working (hover)

public/css/data-discovery-datasets.css
mickel1138: Actually, never mind my second part of the comment, you did that.

public/css/dataland-details.css
```diff
@@ -107,11 +107,6 @@
 	background-color: #DFD;
 }
 
-tr:hover.dsrow
```
mickel1138, line 110: This doesn't even do anything in Dataland, so maybe just move it back, or remove?

mickel1138: I guess it is removed now.

PR #559, opened 2020-04-10
Changed to variable rather than bool

mickel1138:


src/Twig/Extensions.php
```diff
@@ -116,9 +116,40 @@ public function getFilters()
...
+                'orTemplateIfNotExists',
+                [$this, 'doesTwigFileExist']
+            ),
         );
     }
 
+    /**
+     * Does the template exist, or else return base template.
+     *
+     * @return boolean If in maintenance mode.
```
fightingtexasaggie, line 129: Update return annotation

src/Twig/Extensions.php
```diff
@@ -116,9 +116,40 @@ public function getFilters()
...
+            ),
         );
     }
 
+    /**
+     * Does the template exist, or else return base template.
+     *
+     * @return string Filename of basepath, or default.
+     */
+    public function doesTwigFileExist(string $file, $default) : string
```
praneethpr, line 131: Two parameters in the method's signature, but only 1 argument passed when this method is called?

src/Twig/Extensions.php
```diff
@@ -116,9 +116,40 @@ public function getFilters()
...
             ),
+            new \Twig\TwigFilter(
+                'orTemplateIfNotExists',
+                [$this, 'doesTwigFileExist']
+            ),
         );
     }
 
+    /**
+     * Does the template exist, or else return base template.
```
praneethpr, line 127: Need to add annotation

src/Twig/Extensions.php
mickel1138: Right you are. Not sure why GCS didn't catch that.

mickel1138: Oh, misread that, yes, so because this is a filter, the first argument is the value being filtered, and the second argument are the options passed in the function name.

PR #572, opened 2020-04-22
Removed ERDAPP link and attribute, rerouted getters to new function.

mickel1138:


src/Entity/DatasetSubmission.php
```diff
@@ -2764,14 +2754,38 @@ public function getDistributionPoints()
...
+
+        return null;
+    }
+
     /**
      * Getter for the erddap url.
      *
      * @return string|null
      */
     public function getErddapUrl(): ?string
```
praneethpr, line 2780: Are we using this getter anywhere?

mickel1138: No, not currently, but can be used later on the data landing page or, or whoever wants the link. I should probably make getErddapDatasetLink public, incase someone wants to use that.

src/Entity/DatasetSubmission.php
```diff
@@ -2764,14 +2754,38 @@ public function getDistributionPoints()
...
 
+     /**
+     * Getter for the (first) erddap url link.
+     *
+     * @return DatasetLink|null
+     */
+    private function getErdappUrlLink() : ?DatasetLink
+    {
+        $datasetLinks = $this->getDatasetLinks()->filter(function (DatasetLink $datasetLink) {
+            return $datasetLink->getName() == "ERDDAP";
```
praneethpr, line 2765: Can use the constant LINK_NAME_CODES from DatasetLinks Entity

src/Entity/DatasetSubmission.php
praneethpr: Okay, makes sense. I think all the ERDDAP stuff should be moved to DatasetLinks entity

src/Entity/DatasetSubmission.php
```diff
@@ -992,15 +992,6 @@ class DatasetSubmission extends Entity
...
     protected $distributionPoints;
 
-    /**
-     * ERDDAPP Url for the dataset.
-     *
-     * @var string
-     *
-     * @ORM\Column(type="text", nullable=true)
-     */
-    protected $erddapUrl;
```
praneethpr, line 1002: I think there is a missing migration, and what do we want to do about the existing data in those fields?

src/Entity/DatasetSubmission.php
mickel1138: Good one!

src/Entity/DatasetSubmission.php
mickel1138: Yes, forgot to put that in.

src/Entity/DatasetSubmission.php
mickel1138: So these are actually used in the ISO metadata generator.
Also in Dataland, but the button is not visible.

PR #574, opened 2020-04-22
Add citation to mdgen template.

fightingtexasaggie:
Screenshot from 2020-04-22 16-04-18
Included screenshot of validation.

This simply calls dataset's getCitation method, and if that method is made to call the upcoming new citation-generating utility, this won't need to be modified later.

### DIFF
--- a/templates/MetadataGenerator/MD_DataIdentification.xml.twig
+++ b/templates/MetadataGenerator/MD_DataIdentification.xml.twig
@@ -72,6 +72,13 @@
             }
         %}
     </gmd:descriptiveKeywords>
+    <gmd:resourceConstraints xlink:title="Cite As">
+        <gmd:MD_Constraints>
+            <gmd:useLimitation>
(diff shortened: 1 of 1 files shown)


templates/MetadataGenerator/MD_DataIdentification.xml.twig
```diff
@@ -89,6 +89,13 @@
             </gmd:otherConstraints>
         </gmd:MD_LegalConstraints>
     </gmd:resourceConstraints>
+    <gmd:resourceConstraints xlink:title="Cite As">
```
praneethpr, line 92: Do we want to move this up? According to the order in the ticket

fightingtexasaggie: yeah, I will.

mickel1138: also merge in develop

PR #575, opened 2020-04-22
Feature/pelagos 4028 create citation formatting utility and link get getcitation function

mickel1138:
Little bit of moving around of stuff. Kinda wanted to use twig, wadda you guys think?

### DIFF
--- a/src/Entity/Dataset.php
+++ b/src/Entity/Dataset.php
@@ -7,6 +7,8 @@ use Doctrine\ORM\Mapping as ORM;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\ArrayCollection;
 
+use App\Util\DatasetCitationUtil;
+
 /**
  * Dataset Entity class.
  *
@@ -569,26 +571,7 @@ class Dataset extends Entity
      */
     public function getCitation()
     {
-        $title = $this->getTitle();
-        $title = preg_replace('/\.$/', '', $title);
-        $udi = $this->getUdi();
-        $author = $this->getAuthors();
-        $year = null;
-        if ($this->getAcceptedDate() instanceof \Datetime) {
-            $year = $this->getAcceptedDate()->format('Y');
-        }
-        $doi = $this->getDoi();
-
-        $citationString = $author . ' (' . $year . ') ' . $title . '.' .
-            ' Distributed by: GRIIDC '
-            . '(GRIIDC), Harte Research Institute, Texas A&M University-Corpus Christi. ';
-
-        if ($doi instanceof DOI) {
-            $citationString .= 'doi:' . $doi->getDoi();
-        } else {
-            $citationString .= "Available from: http://data.gulfresearchinitiative.org/data/$udi";
-        }
-        return $citationString;
+        return DatasetCitationUtil::getCitation($this);
     }
 
     /**
--- a/src/Util/DatasetCitationUtil.php
+++ b/src/Util/DatasetCitationUtil.php
@@ -28,8 +28,12 @@ class DatasetCitationUtil
         }
         $doi = $dataset->getDoi();
 
-        $citationString = $author . ' (' . $year . ') ' . $title . '.' .
-            ' Distributed by: GRIIDC '
+        $citationString = '';
+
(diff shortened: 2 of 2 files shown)


src/Util/DatasetCitationUtil.php
```diff
@@ -0,0 +1,46 @@
...
+     * Creates the and return the Dataset Citation Text.
+     *
+     * @param Dataset $dataset The Dataset that has the data.
+     *
+     * @return string The Dataset Citation string.
+     */
+    public static function getCitation(Dataset $dataset) :string
+    {
+        $title = $dataset->getTitle();
+        $title = preg_replace('/\.$/', '', $title);
```
fightingtexasaggie, line 22: Do we really have titles that are just a period? If so, why should we be cleaning it up here?

src/Util/DatasetCitationUtil.php
```diff
@@ -0,0 +1,46 @@
...
+        $citationString .= (!empty($author) ? "$author. " : '');
+        $citationString .= (!empty($year) ? "($year). " : '');
+        $citationString .= "$title . ";
+        $citationString .= 'Distributed by: GRIIDC '
+            . '(GRIIDC), Harte Research Institute, Texas A&M University-Corpus Christi. ';
+
+        if ($doi instanceof DOI) {
+            $citationString .= 'doi:' . $doi->getDoi();
+        } else {
+            $citationString .= "Available from: http://data.gulfresearchinitiative.org/data/$udi";
```
fightingtexasaggie, line 42: we've hardcoded the production site like this before in references, so I don't see an issue with doing so.

src/Util/DatasetCitationUtil.php
```diff
@@ -0,0 +1,46 @@
...
+        if ($dataset->getAcceptedDate() instanceof \Datetime) {
+            $year = $dataset->getAcceptedDate()->format('Y');
+        }
+        $doi = $dataset->getDoi();
+
+        $citationString = '';
+
+        $citationString .= (!empty($author) ? "$author. " : '');
+        $citationString .= (!empty($year) ? "($year). " : '');
+        $citationString .= "$title . ";
```
fightingtexasaggie, line 35: Why the extra space between $title and the period?

src/Util/DatasetCitationUtil.php
mickel1138: This is how it was, it removes the last period, we may want to remove this.

src/Entity/Dataset.php
```diff
@@ -569,26 +571,7 @@ public function getIssueTrackingTicket()
      */
     public function getCitation()
```
praneethpr, line 572: Publink application uses the Dataset GET Citation API, which uses this method from the Entity. So changing this would change the behavior of publink

src/Util/DatasetCitationUtil.php
```diff
@@ -0,0 +1,45 @@
...
+        $citationString .= (!empty($author) ? "$author. " : '');
+        $citationString .= (!empty($year) ? "($year). " : '');
+        $citationString .= "$title. ";
+        $citationString .= 'Distributed by: GRIIDC '
+            . '(GRIIDC), Harte Research Institute, Texas A&M University-Corpus Christi. ';
+
+        if ($doi instanceof DOI) {
+            $citationString .= 'doi:' . $doi->getDoi();
+        } else {
+            $citationString .= "Available from: http://data.gulfresearchinitiative.org/data/$udi";
```
praneethpr, line 41: https?

src/Util/DatasetCitationUtil.php
```diff
@@ -0,0 +1,45 @@
...
+            $year = $dataset->getAcceptedDate()->format('Y');
+        }
+        $doi = $dataset->getDoi();
+
+        $citationString = '';
+
+        $citationString .= (!empty($author) ? "$author. " : '');
+        $citationString .= (!empty($year) ? "($year). " : '');
+        $citationString .= "$title. ";
+        $citationString .= 'Distributed by: GRIIDC '
```
praneethpr, line 35: Should we get rid of one of them  "GRIIDC (GRIIDC)" ?

PR #578, opened 2020-04-24
Removed ERDDAP generated section.

mickel1138:
Also captured the old description in the LINK_NAME_CODES for preservation. (Not currently used).
Fixed codeSpace it would have been always 001.

### DIFF
--- a/src/Entity/DatasetLink.php
+++ b/src/Entity/DatasetLink.php
@@ -50,7 +50,7 @@ class DatasetLink extends Entity
     const LINK_NAME_CODES = [
         'erddap' => [
             'name' => 'ERDDAP',
-            'description' => 'An ERDDAP link.',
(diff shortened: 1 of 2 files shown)


src/Entity/DatasetLink.php
```diff
@@ -50,7 +50,7 @@ class DatasetLink extends Entity
     const LINK_NAME_CODES = [
         'erddap' => [
             'name' => 'ERDDAP',
-            'description' => 'An ERDDAP link.',
+            'description' => "ERDDAP's version of the OPeNDAP .html web page for this dataset. Specify a subset of the dataset and download the data via OPeNDAP or in many different file types.",
```
praneethpr, line 53: Is this wording mentioned in the ticket?

mickel1138: No, not used anywhere. I just put it here to capture the old description. May never be used, just like the current description is not used anywhere.

PR #579, opened 2020-04-24
Fixed return type on Dataset Submission entity

praneethpr:


src/Entity/DatasetSubmission.php
```diff
@@ -1586,9 +1586,9 @@ public function getMetadataContacts()
      *
      * @see self::getDatasetContacts()
      *
-     * @return string
+     * @return string|null
      */
-    public function getPointOfContactName() : string
+    public function getPointOfContactName() : ? string
```
fightingtexasaggie, line 1591: Since this can be null, you might want to add |default or otherwise handle the possible null cases in:
templates/DatasetReview/submit.html.twig
templates/DatasetSubmission/submit.html.twig

src/Entity/DatasetSubmission.php
```diff
@@ -1614,9 +1614,9 @@ public function getPointOfContactName() : string
      *
      * @see self::getDatasetContacts()
      *
-     * @return string
+     * @return string|null
      */
-    public function getPointOfContactEmail() : string
+    public function getPointOfContactEmail() : ? string
```
fightingtexasaggie, line 1619: same 2 templates above have the possible null case for this attribute too.

src/Entity/DatasetSubmission.php
praneethpr: Good catch! Fixed it

fightingtexasaggie: this typehint is correct, but should we also make the template handle the null gracefully?

PR #586, opened 2020-04-28
Release/5.38.0

praneethpr:


src/Migrations/Version20200422204033.php
```diff
@@ -0,0 +1,28 @@
...
+
+        $this->addSql('ALTER TABLE dataset_submission DROP erddap_url');
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE dataset_submission ADD erddap_url TEXT DEFAULT NULL');
```
fightingtexasaggie, line 26: Normally I'd think we would want to create code to populate this field from the other representation, but since our data doesn't have any and we're only concerned with this one particular install, this is fine.

PR #588, opened 2020-04-29
Add link button to discovery.

fightingtexasaggie:


templates/DataDiscovery/index.html.twig
```diff
@@ -85,6 +85,9 @@
                             <td valign="top">
                                 <input type="button" id="clear-button" class="button" value="Clear/Show All" onclick="clearAll();">
                             </td>
+                            <td valign="top">
+                                <input type="button" id="searchlink-button" class="button" value="Search" onclick="parent.location=Routing.generate('pelagos_app_ui_searchpage_default')">
```
mickel1138, line 89: It's say the onClick is a nono. If you are going to use jQuery, generate the link in javascript.

mickel1138: And also would use window, rather than parent.

assets/static/css/data-discovery.css
```diff
@@ -241,3 +241,22 @@ input.ui-button {
 #expand-collapse.collapsed {
     left: -12px;
 }
+
+.teal-button
+{
+    background-image:none;
+    background-color:darkcyan;
```
mickel1138, line 248: This color should be: `var(--color-headerMiddle)`
It's a bad name, but it's the right one.

If you don't this won't work in the other template.

assets/static/css/data-discovery.css
```diff
@@ -241,3 +241,22 @@ input.ui-button {
...
+    color:white;
+}
+.teal-button:hover
```
mickel1138, line 251: If you made no changes, why is there a hover, and active?

templates/DataDiscovery/index.html.twig
```diff
@@ -85,6 +85,9 @@
                             <td valign="top">
                                 <input type="button" id="clear-button" class="button" value="Clear/Show All" onclick="clearAll();">
                             </td>
+                            <td valign="top">
+                                <input type="button" id="searchlink-button" class="button teal-button" value="Search V2" onclick="navigateToNewsearch();">
```
mickel1138, line 89: Still has onClick, use jQuery.on, or if you wanna be cool use: `document.getElementById("searchlink-button").addEventListener("click", navigateToNewsearch);`

mickel1138: Can even use an anonymous function in the event listener.

assets/static/css/data-discovery.css
fightingtexasaggie: hover was still needed, active not

assets/static/css/data-discovery.css
fightingtexasaggie: ok, works as suggested.

assets/static/css/data-discovery.css
mickel1138: I was mistaken, and active can even be used. But the point was that the color didn't change, so you can't see that you are hovering. Which may be ok.

PR #589, opened 2020-04-29
Switched menu from discovery to search.

mickel1138:
The Drupal part still have to be done.


templates/Menus/nas-main.html.twig
```diff
@@ -64,4 +64,4 @@
             </div>
         </li>
     </ul>
-</div>
+</div>
```
praneethpr, line 67: Do you want to revert this change?

mickel1138: Did I break something? Is it just a new line?

praneethpr: Just the new line

PR #592, opened 2020-04-30
Misc/startup scripts

fightingtexasaggie:
This goes along with DMS-710. Since it's directly supporting pelagos, I wanted to put it in the share directory we started for these types of things.

### DIFF
--- a/share/bash/control-pelagos-consumers.sh
+++ b/share/bash/control-pelagos-consumers.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+#
+# chkconfig: 35 98 02
+# description: Pelagos RabbitMQ Consumer Supervisor
+#
+# This script to be copied to or symlinked at /usr/local/bin and set to executable.
+#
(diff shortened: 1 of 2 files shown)


share/bash/control-pelagos-consumers.sh
```diff
@@ -0,0 +1,61 @@
...
+        fi
+        echo
+}
+
+# Stop the service
+stop() {
+	cd /opt/pelagos
+	pwd
+        echo -n $"Stopping $prog: "
+        if su - $runuser -c "$basecmd stop $postcmd" ; then
```
praneethpr, line 36: I don't think we need the postcmd for stopping the service. It works without it as well

fightingtexasaggie: doesn't seem to always though

praneethpr: I tested stopping without the wait for supervisord, it works but it doesn't hurt to have that

PR #594, opened 2020-05-01
Renamed base to griidc-base and made base template as the switch to d…

praneethpr:
…ecide the template theme


templates/base.html.twig
```diff
@@ -1,150 +1,10 @@
-<!DOCTYPE html>
-<html style="height:100%;">
-<head>
-{% block head %}
-    <title>GRIIDC | {% block title %}Pelagos{% endblock %}</title>
+{% extends custom_base_template | orTemplateIfNotExists('griidc-base.html.twig') %}
 
-    <link rel="icon" type="image/x-icon" href="{{ asset('build/images/griidc_fav.png') }}" />
+{%block stylesheets %}
+    {{ parent() }}
```
mickel1138, line 4: Which parent?

templates/base.html.twig
```diff
@@ -1,150 +1,10 @@
...
-                        $("#atlwdg-suggest").click(function(e) {
-                            e.preventDefault();
-                            showCollectorDialog();
-                        });
-                    }
-                },
-            };
-        });
-    </script>
+    {{ parent() }}
```
mickel1138, line 9: I don't think it has a parent.

templates/base.html.twig
```diff
@@ -1,150 +1,10 @@
-<!DOCTYPE html>
-<html style="height:100%;">
-<head>
-{% block head %}
-    <title>GRIIDC | {% block title %}Pelagos{% endblock %}</title>
+{% extends custom_base_template | orTemplateIfNotExists('griidc-base.html.twig') %}
 
-    <link rel="icon" type="image/x-icon" href="{{ asset('build/images/griidc_fav.png') }}" />
+{%block stylesheets %}
```
mickel1138, line 3: These blocks are not needed, there is nothing to add or overwrite, these blocks will be inherited from the extended template. Everything works as expected.

templates/base.html.twig
mickel1138, line 9: Can remove.

PR #595, opened 2020-05-01
Changed button color and margin to center

praneethpr:


assets/js/vue/components/search/SearchForm.vue
```diff
@@ -219,6 +219,13 @@
         border-radius: calc(0.25rem - 1px) calc(0.25rem - 1px) 0 0;
         background-clip: border-box;
         border: 1px solid rgba(0, 0, 0, 0.125);
+        .button-toolbar {
+            .map-search-button {
+                color: var(--color-white);
+                background-color: var(--color-headerMiddle);
```
mickel1138, line 225: Maybe we should give these another alias, like Primary, Secondary or so? Or maybe even start a bootstrap theme?
It's confusing, because I don't know what color they are.

But it works fine, looks good.

praneethpr: William picked a color, which matches to headerMiddle color, even if I feel this is confusing. Should we use the color variable name? We have headerMiddle = var(--color-dark-cyan), so instead of using headerMiddle variable name, we could use var(--color-dark-cyan). Should i make that change?

mickel1138: Maybe an improvement. No change needed.

PR #596, opened 2020-05-05
Feature/pelagos 4045 change layout of result row

praneethpr:


src/Event/DatasetIndexSubscriber.php
```diff
@@ -82,10 +83,19 @@ public function populateCalculatedFields(TransformEvent $event)
...
                 $document->set('year', $dataset->getAcceptedDate()->format('Y'));
                 $document->set('updatedDateTime', $dataset->getAcceptedDate()->format('Ymd\THis\Z'));
+                $document->set('acceptedDate', $dataset->getAcceptedDate()->format('Y-m-d'));
             } else {
                 $document->set('year', $dataset->getDatasetSubmission()->getModificationTimeStamp()->format('Y'));
                 $document->set('updatedDateTime', $dataset->getDatasetSubmission()->getModificationTimeStamp()->format('Ymd\THis\Z'));
             }
+            // Populate file size and format values
+            $document->set('fileSize', $dataset->getDatasetSubmission()->getDatasetFileSize());
+            if ($dataset->getDatasetSubmission()->isDatasetFileInColdStorage()) {
```
fightingtexasaggie, line 93: :+1:

praneethpr:
> tested. Works and meets ticket, however, I'd like to suggest an improvement of not making the entire card clickable. As stands, it is not possible to copy/paste a DOI or UDI from the card, and it is likely users will want to.

Yes, I agree. Probably we should bring it up to the PO

mickel1138:
> > tested. Works and meets ticket, however, I'd like to suggest an improvement of not making the entire card clickable. As stands, it is not possible to copy/paste a DOI or UDI from the card, and it is likely users will want to.
> 
> Yes, I agree. Probably we should bring it up to the PO

Can we just not make it an A (anchor), and get a click event on it?

assets/js/vue/components/search/ResultRow.vue
```diff
@@ -1,25 +1,38 @@
...
+            <div>
+                <span class="badge badge-secondary" v-if="resultRowData.availabilityStatus === 0">Identified</span>
+                <span class="badge badge-primary" v-else-if="resultRowData.availabilityStatus === 2 || resultRowData.availabilityStatus === 4">Submitted</span>
+                <span class="badge badge-danger" v-else-if="resultRowData.availabilityStatus === 5 || resultRowData.availabilityStatus === 8">Restricted</span>
+                <span class="badge badge-success" v-else-if="resultRowData.availabilityStatus === 7 || resultRowData.availabilityStatus === 10">Available</span>
+            </div>
+            <b-card-title>{{ resultRowData.title }}</b-card-title>
+            <b-card-text class="d-flex justify-content-between" >
+                <div v-if="Object.keys(resultRowData.datasetSubmission).length > 0">
+                    <div v-if="resultRowData.datasetSubmission.authors">
```
mickel1138, line 14: I would wrap these in a div maybe, with a style. Font could be a bit smaller for more real estate, but that's just my opinion. No change needed.

PR #598, opened 2020-05-05
Feature/pelagos 4047 improvements on menu bar

praneethpr:


assets/scss/nas-grp.scss
```diff
@@ -96,60 +95,51 @@
     content: "";
     width: 100%;
     height: 100%;
-    background-color: #fff;
+    background-color: var(--color-white);
     z-index: -1;
     opacity: .9;
   }
 
   .navbar {
```
mickel1138, line 103: I think it's more convinient is we put all the menu stuff in it's own (s)css.

PR #600, opened 2020-05-06
Citation text change

praneethpr:


src/Util/DatasetCitationUtil.php
```diff
@@ -33,7 +33,7 @@ public static function getCitation(Dataset $dataset) :string
         $citationString .= (!empty($author) ? "$author. " : '');
         $citationString .= (!empty($year) ? "$year. " : '');
         $citationString .= "$title. ";
-        $citationString .= 'Distributed by: GRIIDC, Harte Research Institute, Texas A&M University-Corpus Christi. ';
+        $citationString .= 'Distributed by: Gulf of Mexico Research Initiative Information and Data Cooperative (GRIIDC), Harte Research Institute, Texas A&M University – Corpus Christi.';
```
fightingtexasaggie, line 36: Verified with Nicole Rothney from marketing/communications a bit ago, there is not supposed to be a space around the endash (not emdash) character between University and Corpus.

Texas A&M University–Corpus Christi

praneethpr: Made the change

PR #602, opened 2020-05-06
Removed reference date and date type fields

praneethpr:


mickel1138:
Dataset entity also has Reference Year.

mickel1138:
Also reference date in function clearDatasetSubmission in DatasetSubmissionController.php

praneethpr:
> Also reference date in function clearDatasetSubmission in DatasetSubmissionController.php

Done

src/Entity/Dataset.php
```diff
@@ -388,22 +388,6 @@ public function getAuthors()
         }
     }
 
-    /**
-     * Get reference date shortcut, to get Authors from DatasetSubmission.
-     *
-     * @return string|null
-     */
-    public function getReferenceDateYear()
```
mickel1138, line 396: Removing this now may break other things until other tickets are merged.

praneethpr: Yes, this pull request can be merged at the end

PR #603, opened 2020-05-07
Added space

praneethpr:


praneethpr:
> Not needed, fix this in whoever needs the space instead.

I was wrong about Dataland, it's in the Citation Util itself. So, its in the right place

mickel1138:
> > Not needed, fix this in whoever needs the space instead.
> 
> I was wrong about Dataland, it's in the Citation Util itself. So, its in the right place

You are right, I suck.

PR #608, opened 2020-05-08
Bug/pelagos 4066 fix duplicate numbers

mickel1138:


templates/DatasetMonitoring/projects.html.twig
```diff
@@ -35,17 +35,21 @@
                 <tr width="100%">
                     <td>
                         <div class="dotchart_legend" style="position: relative;">
-                            <span id="a"><strong>Dataset with status:</strong>
+                            <span><strong>Dataset with status:</strong>
                                 <img src="{{ asset('build/images/green-circle.png') }}"> Available:
+                                <span id="a"></span>
```
praneethpr, line 40: Should we rename the id's

assets/static/js/dotgraph.js
```diff
@@ -30,10 +30,10 @@ function graphDatasetStatus(){
 }
 
 function createGraph(data, divClass, divSelAttr, divSelAttrVal, cntA, cntR, cntI, myDiv){
-    $('.dotchart_legend #a', myDiv).append(" "+cntA);
-    $('.dotchart_legend #r', myDiv).append(" "+cntR);
-    $('.dotchart_legend #i', myDiv).append(" "+cntI);
-    $('.dotchart_legend #tot', myDiv).append(" " + (+cntA + +cntR + +cntI) + "");
+    $('.dotchart_legend #a', myDiv).text(" "+cntA);
```
praneethpr, line 33: Maybe we should've used always used .text()

templates/DatasetMonitoring/projects.html.twig
fightingtexasaggie: I have to agree. 'Accepted', 'Registered', 'Identified' was some of the old nomenclature.

PR #616, opened 2020-05-13
Empty string DOI.doi in index.

fightingtexasaggie:
If there's a better way to explicitly set to null, please chime in. $document->set('somenestedindex', null) doesn't work. I'm not terribly happy about this hackish emptystring in a new DOI object, but it works.


src/Event/DatasetIndexSubscriber.php
```diff
@@ -119,6 +120,12 @@ public function populateCalculatedFields(TransformEvent $event)
                 $document->set('collectionEndDate', $collectionEndDate->format('Y-m-d H:i:s'));
             }
         }
+
+        // For DIF only, set DOI to a DOI with empty string as the doi value.
+        if (false === $dataset->hasDatasetSubmission()) {
```
praneethpr, line 125: We usually use instanceOf to check if its an object and if its not null, and we also need to check if the dataset has been submitted

src/Event/DatasetIndexSubscriber.php
```diff
@@ -119,6 +120,12 @@ public function populateCalculatedFields(TransformEvent $event)
...
+                $doi = new DOI('');
```
praneethpr, line 126: You don't need to do this, you can get the DOI from the dataset ($dataset->getDoi())

mickel1138:
I'm not sure this is the right approach. This should probably be done in the front end?

src/Event/DatasetIndexSubscriber.php
fightingtexasaggie: hasDatasetSubmission returns a bool, so used it over $dataset->getDatasetSubmission(). I forgot about the unsubmitted...will fix.

src/Event/DatasetIndexSubscriber.php
fightingtexasaggie: indeed correct, changed.

mickel1138:
> Doi doesn't show on any of them now

Yes, I verified this too. I didn't turn of the Identified filter.

fightingtexasaggie:
Closing PR. We agreed fixing in the view is a better solution, instead of custom indexing.

PR #617, opened 2020-05-14
Removed tooltips from result row and added tool tip to dataset status…

praneethpr:
… facets


assets/js/vue/components/search/Facet.vue
```diff
@@ -79,6 +88,24 @@
 
                     }
                 }
+            },
+            statusTooltip: function (datasetStatus) {
+                let datasetStatusTooltip = "";
+                switch (true) {
+                    case (datasetStatus === "Available"):
+                        datasetStatusTooltip = "This dataset is available for download.";
```
mickel1138, line 96: These tooltip texts are used in a previous PR. (Without periods btw). Is there a way we can unifiy these, make in a constant, or include file? ¯\_(ツ)_/¯

PR #619, opened 2020-05-15
Renamed infoTab to overviewtab and added required attributes on the tab

praneethpr:


assets/js/vue/components/research-group/OverviewTab.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+    <div>
+        <h4>{{ overview.name }}</h4>
+        <hr>
+        <h6 class="overview-label">Project Director: </h6>
```
fightingtexasaggie, line 5: want to make it plural if there's more than one? (a nice touch)

PR #620, opened 2020-05-18
Reused the result row component in vue and made changes to the resear…

praneethpr:
…ch group API


assets/js/vue/components/research-group/DatasetsTab.vue
```diff
@@ -1,25 +1,15 @@
...
-                        <article class="col-lg-12">
-                            <b-card-text>
-                                {{ dataset.citation }}
-                            </b-card-text>
-                        </article>
-                    </div>
-                </b-card-body>
-            </b-link>
-        </b-card>
+        <ResultRow :resultRowData="dataset" v-for="dataset in datasets" v-bind:key="dataset.udi"/>
```
praneethpr, line 3: Do we want to change the name of this Vue component to DatasetRow in this Pull request? Now, that we are reusing Vue components I thought DatasetRow would make more sense

fightingtexasaggie: I think we should, if possible, otherwise we might never get around to it later.

assets/js/vue/components/search/DatasetRow.vue
```diff
@@ -0,0 +1,90 @@
+<template>
```
praneethpr, line 1: Rename issue

assets/js/vue/components/search/DatasetRow.vue
```diff
@@ -0,0 +1,90 @@
...
+        data: function () {
+            return {
+                url: Routing.generate("pelagos_app_ui_dataland_default", {'udi' : this.datasetRowData.udi } )
+            }
+        }
+    }
+</script>
+
+<style scoped>
+    div.bootstrap p {
```
mickel1138, line 57: Where is the P?

assets/js/vue/components/search/DatasetRow.vue
```diff
@@ -0,0 +1,90 @@
...
+        margin-bottom: 0 !important;
+    }
+
+    div.bootstrap h4 {
```
mickel1138, line 61: Where is the H4? Also, you don't want to do size with CSS is you are using Bootstrap. That can be done in markup.

assets/js/vue/components/search/DatasetRow.vue
```diff
@@ -0,0 +1,90 @@
...
+    .card-product {
+        margin-bottom: 1rem;
+    }
+
+    .card-product:hover .btn-overlay {
+        opacity: 1;
+    }
+
+    .card-product:hover {
+        -webkit-box-shadow: 0 4px 15px rgba(153, 153, 153, 0.3);
```
mickel1138, line 74: Webkit specific is not longer needed due to PostCSS. (I think that should also work here.)

assets/js/vue/components/search/DatasetRow.vue
praneethpr: Old code when this row had citation text, good catch!

assets/js/vue/components/search/DatasetRow.vue
praneethpr: Old code when this row had citation text, good catch!

assets/js/vue/components/search/DatasetRow.vue
praneethpr: Yes, will remove it

src/Entity/ResearchGroup.php
```diff
@@ -335,18 +336,35 @@ public function serializeDatasets()
...
                 $datasetArray['dif'] = null;
             }
+            if ($dataset->hasDatasetSubmission()) {
+                $datasetArray['datasetSubmission'] = array(
+                    'authors' =>  $dataset->getDatasetSubmission()->getAuthors(),
+                    'themeKeywords' => $dataset->getDatasetSubmission()->getThemeKeywords()
+                );
+                $datasetArray['fileFormat'] = $dataset->getDatasetSubmission()->getDistributionFormatName();
+                if ($dataset->getDatasetSubmission()->isDatasetFileInColdStorage()) {
+                    $datasetArray['fileSize'] = TwigExtentions::formatBytes($dataset->getDatasetSubmission()->getDatasetFileColdStorageArchiveSize(), 2);
```
mickel1138, line 361: We should probably move formatBytes to a utility.

fightingtexasaggie: We certainly could, but it kind of makes sense out on the edge, just like applying any other localization.

PR #621, opened 2020-05-18
Switched help and about us menu item

praneethpr:


praneethpr:
> looks better imho. Should we consider the same for the GoMRI template?

We can ask the PO

PR #622, opened 2020-05-19
Feature/pelagos 4076 add rg land fields

praneethpr:


assets/js/vue/components/research-group/OverviewTab.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+    <div>
+        <h4>{{ overview.name }}</h4>
+        <hr>
+        <h6 class="overview-label">Project Director: </h6>
```
mickel1138, line 5: Use class (add it) `font-weight-bold` (if you want Bootstrap to control bold weight) or just add <b> tags.
Instead of styling later in the file, I fill that way you can see right away this needs to be bold.

assets/js/vue/components/research-group/OverviewTab.vue
```diff
@@ -0,0 +1,51 @@
...
+                        return person;
+                    }
+                });
+            }
+        }
+    }
+</script>
+
+<style scoped>
+    .overview-label {
```
mickel1138, line 48: Remove this, see comment above.

assets/js/vue/components/research-group/OverviewTab.vue
praneethpr: Forgot about the bootstrap class, good catch!

assets/js/vue/components/research-group/PeopleTab.vue
```diff
@@ -1,12 +1,18 @@
...
-        <b-card class="card-product" no-body v-for="person in personResearchGroups" v-bind:key="person.id">
-            <b-card-body>
-                <p><strong>Name: </strong>{{ person.person.firstName + " " + person.person.lastName }}</p>
-                <p><strong>Label: </strong>{{ person.label }}</p>
-            </b-card-body>
-        </b-card>
-    </b-card-text>
+    <b-card-group columns>
+        <b-card-text>
+            <b-card class="card-product" no-body v-for="person in personResearchGroups" v-bind:key="person.id">
```
fightingtexasaggie, line 4: Do we need to consider some sort of ordering?

PR #625, opened 2020-05-21
Changed begin/end date to date (not with time and timezone)

mickel1138:
Also needed migration.


mickel1138:
I gave this another good look, and test. I removed setting the timezones. This should work fine. Even Doctrine date type is a php DateTime, so it will always have a timezone etc. If this is too confusing, we could consider just using a text type.

src/Migrations/Version20200521165709.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE dataset_submission ALTER temporal_extent_begin_position TYPE TIMESTAMP(0) WITH TIME ZONE');
+        $this->addSql('ALTER TABLE dataset_submission ALTER temporal_extent_begin_position DROP DEFAULT');
+        $this->addSql('ALTER TABLE dataset_submission ALTER temporal_extent_end_position TYPE TIMESTAMP(0) WITH TIME ZONE');
+        $this->addSql('ALTER TABLE dataset_submission ALTER temporal_extent_end_position DROP DEFAULT');
+
+        $this->addSql("UPDATE dataset_submission SET temporal_extent_begin_position = temporal_extent_begin_position AT TIME ZONE '+10'");
```
fightingtexasaggie, line 38: I thought you didn't want to just target the middle of the day ;-) j/k. cool.

mickel1138: 10 hours is 5 hours from migrating, plus 5 hours to bring it to UTC. The timezone is not stored as a name but in ISO format.

praneethpr:
Do you want to merge develop into this? It still throws the babel error

PR #629, opened 2020-05-22
Added publications tab and publication row vue components

praneethpr:


assets/js/vue/components/research-group/PublicationsTab.vue
```diff
@@ -0,0 +1,35 @@
...
+        components: { PublicationRow },
+        props: {
+            datasets: {}
+        },
+        methods: {
+            getPublications: function () {
+                let publications = [];
+                let publicationId = [];
+                this.datasets.forEach(dataset => {
+                    dataset.publications.forEach(publication => {
```
praneethpr, line 20: Not a fan of using nested FOR loops, couldn't find a better way. Please let me know if there is an efficient way to do this

mickel1138: Maybe you could use `array_unique` ?

fightingtexasaggie: The overhead of the inefficiency of the nested loops is negligible with the relative small numbers we're ever likely to deal with, and the inner loop isn't doing that much anyway. I'm ok with it as is, but Mike's suggestion would be fine too.

praneethpr: Isn't array_unique a Php function?

PR #634, opened 2020-05-28
Changed people card column count to 2

praneethpr:


mickel1138:
> Tested, works. Looks good, maybe less padding/margin on the cards.

The padding can be set later in the Bootstrap variables (in another PR)

PR #635, opened 2020-05-28
Added URI to search results.

mickel1138:
Because the URL comes from the FOS router, and the FOS router has a base path set that can not be manipulated without causing other issues, I added the URI to the search result. This is always a good Rest thing to do, to show the resource for your result.

### DIFF
--- a/assets/js/vue/components/search/DatasetRow.vue
+++ b/assets/js/vue/components/search/DatasetRow.vue
@@ -44,7 +44,7 @@
         },
         data: function () {
             return {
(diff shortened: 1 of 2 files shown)


assets/js/vue/components/search/DatasetRow.vue
```diff
@@ -44,7 +44,7 @@
         },
         data: function () {
             return {
-                url: Routing.generate("pelagos_app_ui_dataland_default", {'udi' : this.datasetRowData.udi } )
+                url: this.datasetRowData.uri
```
mickel1138, line 47: I guess we could have just named it the new search field url, and then skipped this part. Or call URI instead or URL in the window.open.

PR #636, opened 2020-05-29
Added remotely hosted badge for dataset row component

praneethpr:


assets/js/vue/components/search/DatasetRow.vue
```diff
@@ -5,6 +5,7 @@
                 <span class="badge badge-primary" v-else-if="datasetRowData.availabilityStatus === 2 || datasetRowData.availabilityStatus === 4">Submitted</span>
                 <span class="badge badge-danger" v-else-if="datasetRowData.availabilityStatus === 5 || datasetRowData.availabilityStatus === 8">Restricted</span>
                 <span class="badge badge-success" v-else-if="datasetRowData.availabilityStatus === 7 || datasetRowData.availabilityStatus === 10">Available</span>
+                <span class="badge badge-warning" v-if="datasetRowData.availabilityStatus === 7">Remotely Hosted</span>
```
mickel1138, line 8: Aren't there two number that mean remotely hosted? Even if... we should cover them both?

praneethpr: I thought we used it, but as the POs mentioned it such a state shouldn't exist in the workflow. Probably we should get rid of that state

mickel1138: should !== could

PR #637, opened 2020-06-01
Feature/pelagos 4103 create bootstrap color config file

mickel1138:
Some cleanup is needed and styling needs to be done. And colors configured.


mickel1138:
The GRIIDC theme looks meh, but will be tweaked with PO help.

PR #642, opened 2020-06-03
Feature/pelagos 4108 filter tree used in monitoring and discovery

mickel1138:
This won't work without the new filter util, which is in another PR.

### DIFF
--- a/src/Controller/Api/TreeController.php
+++ b/src/Controller/Api/TreeController.php
@@ -21,6 +21,7 @@ use App\Entity\ResearchGroup;
 use App\Entity\Person;
 
 use App\Util\DatasetIndex;
+use App\Util\FundingOrgFilter;
 
 /**
  * The Tree API controller.
@@ -115,7 +116,7 @@ class TreeController extends EntityController
      *
      * @return string
      */
-    public function getFundingOrganizationsAction(Request $request, DatasetIndex $datasetIndex)
+    public function getFundingOrganizationsAction(Request $request, DatasetIndex $datasetIndex, FundingOrgFilter $fundingOrgFilter)
     {
         $tree = $this->buildTreeConfig($request);
         $filter = false;
@@ -141,8 +142,8 @@ class TreeController extends EntityController
             $criteria['id'] = array_keys($fundingOrganizations);
         }
 
-        if (true === true) { // To be replaced with check if needs to be filtered.
-            $criteria['id'] = array(0 => 15); // To be replaced by function getIdArray or something.
+        if ($fundingOrgFilter->isActive()) {
+            $criteria['id'] = $fundingOrgFilter->getFilterIdArray();
         }
 
         return $this->render(
--- a/src/Controller/Api/TreeController.php
+++ b/src/Controller/Api/TreeController.php
@@ -140,6 +140,11 @@ class TreeController extends EntityController
             }
             $criteria['id'] = array_keys($fundingOrganizations);
         }
+
+        if (true === true) { // To be replaced with check if needs to be filtered.
+            $criteria['id'] = array(0 => 15); // To be replaced by function getIdArray or something.
+        }
+
         return $this->render(
             'Api/Tree/research_awards.json.twig',
             array(


mickel1138:
Merged in the switch util, it won't work without it. Merge the other PR first.

src/Controller/Api/TreeController.php
```diff
@@ -115,7 +116,7 @@ public function __construct(EntityManagerInterface $doctrineOrmEntityManager, En
      *
      * @return string
      */
-    public function getFundingOrganizationsAction(Request $request, DatasetIndex $datasetIndex)
+    public function getFundingOrganizationsAction(Request $request, DatasetIndex $datasetIndex, FundingOrgFilter $fundingOrgFilter)
```
mickel1138, line 119: Add FundingOrgFilter to @param  above.

src/Controller/Api/TreeController.php
```diff
@@ -264,7 +270,7 @@ public function getResearchGroupsByFundingCycleAction(Request $request, int $fun
      *
      * @return string
      */
-    public function getLettersAction(Request $request)
+    public function getLettersAction(Request $request, FundingOrgFilter $fundingOrgFilter)
```
mickel1138, line 273: Add FundingOrgFilter to @param  above.

src/Controller/Api/TreeController.php
```diff
@@ -330,13 +344,13 @@ public function getLettersAction(Request $request)
      *
      * @return string
      */
-    public function getPeopleAction(Request $request, string $letter)
+    public function getPeopleAction(Request $request, string $letter, FundingOrgFilter $fundingOrgFilter)
```
mickel1138, line 347: Add FundingOrgFilter to @param  above.

src/Controller/Api/TreeController.php
```diff
@@ -347,8 +361,16 @@ public function getPeopleAction(Request $request, string $letter)
...
             ->addOrderBy('person.firstName')
-            ->setParameter('letter', "$letter")
-            ->getQuery();
+            ->setParameter('letter', "$letter");
+
+        if ($fundingOrgFilter->isActive()) {
+            $qb->innerJoin('person.personResearchGroups', 'prg');
+            $qb->innerJoin('prg.researchGroup', 'rg');
+            $qb->andWhere('rg.id IN (:fos)');
+            $qb->setParameter('fos', $fundingOrgFilter->getResearchGroupsIdArray());
```
mickel1138, line 370: fos should rgs or something.

src/Controller/Api/TreeController.php
```diff
@@ -278,11 +284,19 @@ public function getLettersAction(Request $request)
...
             ->distinct()
-            ->orderBy($firstLetterLastName)
-            ->getQuery();
+            ->orderBy($firstLetterLastName);
+
+        if ($fundingOrgFilter->isActive()) {
+            $qb->innerJoin('person.personResearchGroups', 'prg');
+            $qb->innerJoin('prg.researchGroup', 'rg');
+            $qb->where('rg.id IN (:fos)');
+            $qb->setParameter('fos', $fundingOrgFilter->getResearchGroupsIdArray());
```
praneethpr, line 296: Instead of using this triple nested foreach method to get research group ids, we can just use two more inner joins to get the result. Something like this
`            $qb->innerJoin('person.personResearchGroups', 'prg');
            $qb->innerJoin('prg.researchGroup', 'rg');
            $qb->innerJoin('rg.fundingCycle', 'fundCyc');
            $qb->innerJoin('fundCyc.fundingOrganization','fundOrg');
            $qb->andWhere('fundOrg.id IN (:fundingOrgIds)');
            $qb->setParameter('fundingOrgIds', $fundingOrgFilter->getFilterIdArray());`

src/Util/FundingOrgFilter.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return $ids;
+    }
+    
+    /**
+     * Returns true is you need to filter by Funding Organization.
+     *
+     * @return array List of ID's to filter on.
+     */
+    public function getResearchGroupsIdArray() :array
```
praneethpr, line 87: Can remove this method if we replace it with the doctrine query

src/Controller/Api/TreeController.php
praneethpr: Same thing here

src/Controller/Api/TreeController.php
praneethpr: FundingOrgFilter could be moved to the constructor as it is being used at more than one place

src/Util/FundingOrgFilter.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+     /**
+     * Set array of FOs to filter by.
+     *
+     * @param array $fundingOrgs List of Funding Organizations Short Names.
+     *
+     * @return void
+     */
+    private function shortNamesToIdArray(array $fundingOrgs)
```
praneethpr, line 50: This method is being called only once (from the constructor) and its a one liner, it could stay in the constructor itself

src/Controller/Api/TreeController.php
mickel1138: I think that's kinda the beauty of local dependency injection. You only get an instance where it's needed, not for the entire controller. The constructor is how we HAD to do it  before this.

src/Controller/Api/TreeController.php
mickel1138: Why not go straight to shortnames?  We may have to this 2 or 3 times, and the list of researchgroups is already available from the utility, which only had to run the query once. It also a little more efficient, but it's more about decentralizing in my mind.

src/Controller/Api/TreeController.php
praneethpr: Are you talking about before the FLEX upgrade? I think we used to get it from the container(). After flex upgrade, if a dependency injection is required for a Controller method, we use the local dependency injection or else if it is required more than once, we move it to the constructor. Just like entityManager or entityHandler

src/Controller/Api/TreeController.php
praneethpr: yes, we can use funding org short names too instead of Ids. But i think using inner joins to filter is more efficient that using the getResearchGroupsIdArray() method

PR #644, opened 2020-06-04
Enable basic homepage for nasgrp.

fightingtexasaggie:
We need something now because we don't have Drupal handling our default route for nasgrp. What is in the controller now results in a re-direct loop, and I think it simply never gets there in griidc mode because Drupal handles it first.

### DIFF
--- a/src/Controller/DefaultController.php
+++ b/src/Controller/DefaultController.php
@@ -24,13 +24,13 @@ class DefaultController extends AbstractController
      */
     public function index()
     {
-        if ($this->getParameter('kernel.debug')) {
+        if ($this->getParameter('kernel.debug') or (!empty($this->getParameter('custom_template')))) {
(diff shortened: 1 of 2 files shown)


src/Controller/DefaultController.php
```diff
@@ -24,13 +24,13 @@ class DefaultController extends AbstractController
      */
     public function index()
     {
-        if ($this->getParameter('kernel.debug')) {
+        if ($this->getParameter('kernel.debug') or (!empty($this->getParameter('custom_template')))) {
             return $this->render('Default/index.html.twig');
         } else {
             return $this->redirect('/', 302);
```
fightingtexasaggie, line 30: In practice, I think this re-directs site/pelagos-symfony/ back to / which is handled by Drupal.

PR #645, opened 2020-06-04
Feature/pelagos 4094 adjust search to not display facets with 0 results

fightingtexasaggie:


src/Util/Search.php
```diff
@@ -62,7 +62,7 @@ class Search
      * Elastic index mapping for dataset DOI.
      */
     const ELASTIC_INDEX_MAPPING_DOI = 'doi.doi';
-    
+
```
praneethpr, line 65: I don't know why these whitespaces get changed sometimes

fightingtexasaggie: My editor is set to remove them. That looks like a soft tab (4 whitspaces) that was left on a line by itself, probably from a previous edit.

PR #647, opened 2020-06-05
Added tooltip and ellipsis for the overflowing text on organization n…

praneethpr:
…ame in the people's tab

### DIFF
--- a/assets/js/vue/components/research-group/PeopleTab.vue
+++ b/assets/js/vue/components/research-group/PeopleTab.vue
@@ -11,7 +11,7 @@
                     <div>
                         {{ person.label }}
                     </div>
-                    <div class="text-truncate text-width"
+                    <div class="text-truncate org-name-width"
                          v-tooltip="{
                                     content: person.person.organization,
                                     placement:'top'
@@ -47,7 +47,7 @@
         }
     }
 
-    .text-width {
+    .org-name-width {
         max-width: 250px;
     }
 </style>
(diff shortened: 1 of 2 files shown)


assets/js/vue/components/research-group/PeopleTab.vue
```diff
@@ -3,12 +3,26 @@
...
-                    </p>
+                    <div>
+                        <p>
+                            <strong>{{ person.person.firstName + " " + person.person.lastName }}</strong> <br> <br>
+                        </p>
+                    </div>
+                    <div>
+                        {{ person.label }}
+                    </div>
+                    <div class="text-overflow"
```
mickel1138, line 14: Use: `text-truncate` (a bootstrap class) and also `mw-50` (or whatever is needed). Instead of the CSS below.
Remember, you can do _nearly_ all CSS with Bootstrap.

praneethpr: For this we can't use the bootstrap width class (w-50), I tried it, it changes the whole tab layout. But i dint know about text-truncate, will use that 👍

assets/js/vue/components/research-group/PeopleTab.vue
```diff
@@ -32,4 +46,8 @@
             height: 200px;
         }
     }
+
+    .text-width {
```
mickel1138, line 50: This may not be the best class name, it's kinda ambiguous, not very descriptive of what it goes with.
Maybe it should be `organization-max-width` ?

praneethpr: yes, you are right. Changed it

PR #652, opened 2020-06-11
Added Yargs, and addition argument to webpack...js

mickel1138:
Use parameter --publicpath /your/path/here for production.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,5 @@
 var Encore = require('@symfony/webpack-encore');
+var arguments = require('yargs').argv;
 
 // Manually configure the runtime environment if not already configured yet by the "encore" command.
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -11,7 +12,7 @@ Encore
     .setOutputPath('public/build/')
 
     // public path used by the web server to access the output path
-    .setPublicPath(Encore.isProduction() ? '/pelagos-symfony/build' : '/build')
+    .setPublicPath(arguments.publicpath ? arguments.publicpath : '/build')
```
fightingtexasaggie, line 15: Maybe we should add a "dev" environment too, just to keep the usage clean. IDK.

mickel1138: Default is dev, so always "/build". `yarn dev` will use default path.

PR #657, opened 2020-06-15
Removed user associated research groups

praneethpr:


src/Controller/DefaultController.php
```diff
@@ -21,20 +22,20 @@ class DefaultController extends AbstractController
...
         $researchGroups = array();
-        if ($this->isGranted('IS_AUTHENTICATED_FULLY')) {
-            if ($this->isGranted('ROLE_DATA_REPOSITORY_MANAGER')) {
-                $researchGroups = $this->container->get('doctrine')->getRepository(ResearchGroup::class)->findAll();
-            } elseif ($this->tokenStorage->getToken()->getUser() instanceof Account) {
-                $researchGroups = $this->tokenStorage->getToken()->getUser()->getPerson()->getResearchGroups();
-            }
+
+        if ($fundingOrgFilter->isActive()) {
+            $researchGroups = $this->get('doctrine')->getRepository(ResearchGroup::class)->findBy(array('id' => $fundingOrgFilter->getResearchGroupsIdArray() ));
```
mickel1138, line 36: Maybe we can start using the new ResearchGroupReposity soon ;-)

templates/Default/nas-grp-index.html.twig
```diff
@@ -19,25 +19,23 @@
...
-                        </a>
-
-                        <div class="dropdown-menu scrollable-menu overflow-auto" aria-labelledby="dropdownMenuLink">
-                            {% for researchGroup in researchGroups %}
-                                <a class="dropdown-item" href="{{ path('pelagos_app_ui_researchgroup_about', { id: researchGroup.getId}) }}" target="_blank">{{ researchGroup.name }}</a>
-                            {% endfor %}
-                        </div>
+                    <div class="dropdown-menu scrollable-menu overflow-auto" aria-labelledby="dropdownMenuLink">
+                        {% for researchGroup in researchGroups %}
+                            <a class="dropdown-item" href="{{ path('pelagos_app_ui_researchgroup_about', { id: researchGroup.getId}) }}" target="_blank">{{ researchGroup.name }}</a>
```
mickel1138, line 34: This may be just me, but I think the target should be _self.

src/Controller/DefaultController.php
mickel1138: Also, should they be ordered?

templates/Default/nas-grp-index.html.twig
praneethpr: We don't want it to open in a new tab?

src/Controller/DefaultController.php
praneethpr: Not specified in the ticket, but we can do it

PR #661, opened 2020-06-16
Feature/pelagos 4132 add result size variable to the query

praneethpr:


src/Util/Search.php
```diff
@@ -216,7 +218,8 @@ private function getResearchGroupsInfo(array $aggregations): array
         }
 
         //Sorting based on highest count
-        array_multisort(array_column($researchGroupsInfo, 'count'), SORT_DESC, $researchGroupsInfo);
```
praneethpr, line 219: We need to pass only variables by reference, throws a warning now. Fixed it, but it doesn't have to be in this branch. I'm okay with either way

PR #664, opened 2020-06-18
Added cold storage property to elastic index and added the badge to r…

praneethpr:
…epresent it in the dataset row component, changed color to iceblue

### DIFF
--- a/config/packages/fos_elastica.yaml
+++ b/config/packages/fos_elastica.yaml
@@ -124,6 +124,9 @@ fos_elastica:
                                     type: 'text'
                         identifiedStatus:
                             type: 'integer'
+                        coldStorage:
+                            property_path: false
+                            type: 'boolean'
                     persistence:
                         driver: orm
                         model: App\Entity\Dataset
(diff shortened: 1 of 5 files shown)


praneethpr:
To test:
Can use R4.x258.000:0014 this dataset

config/packages/fos_elastica.yaml
```diff
@@ -124,6 +124,9 @@ fos_elastica:
                                     type: 'text'
                         identifiedStatus:
                             type: 'integer'
+                        coldStorage:
```
mickel1138, line 127: One of these days, maybe today, we should make properties for these states.
(Also isRestricted, isAvailable, isRemotelyhosted)

mickel1138: Also because this exists in submission, couldn't we copy the property instead? Having this property at the dataset level is better maybe?

praneethpr: You mean at the entity level? I think we have those already. But we should start using them

PR #665, opened 2020-06-18
Added paging controls at the bottom of the page

praneethpr:


assets/js/vue/components/search/ResultSet.vue
```diff
@@ -42,13 +42,26 @@
...
         </section>
-        <b-pagination
-                v-model="currentPage"
-                :total-rows="rows"
-                :per-page="formValues.perPage"
-                class="bg justify-content-center"
-                style="margin-bottom: 100px;">
-        </b-pagination>
+
+        <div class="container">
```
mickel1138, line 46: Maybe this should be a component?

mickel1138, line 46: If you move this div in the `<section>` above, it looks better. There is a stray gray box in the front.

praneethpr: I don't see the benefit of making this pagination to be a component, its more work to get events propagated from the child back to the parent, then submit the form in the grandparent component

praneethpr: Moved it

PR #667, opened 2020-06-18
Do not html encode.

mickel1138:


mickel1138:
This was needed or else the dreaded `&amp;` would show up.

PR #676, opened 2020-06-26
Removed text block.

mickel1138:
This may be a bit aggressive, but it works.

### DIFF
--- a/assets/js/vue/components/search/SearchForm.vue
+++ b/assets/js/vue/components/search/SearchForm.vue
@@ -76,7 +76,7 @@
                 @noOfResults="changeNoOfResults"
                 :formValues="form"/>
         <section class="section-content pt-3 bg" v-else>
-            <div v-show="!hideSearchTextBlock" class="container">
+            <div v-show="!displayTextBlock" class="container">
                 <article class="card">
                     <div class="card-body text-center">
                         <h5 class="card-title">
--- a/assets/js/vue/components/search/SearchForm.vue
+++ b/assets/js/vue/components/search/SearchForm.vue
@@ -118,7 +118,7 @@
             }
         },
         computed: {
-            hideSearchTextBlock: function () {
+            displayTextBlock: function () {
                 if (typeof window.PELAGOS_TEMPLATE_PROPS !== 'undefined') {
                     return window.PELAGOS_TEMPLATE_PROPS.hideSearchTextBlock;
                 } else {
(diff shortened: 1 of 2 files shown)


assets/js/vue/components/search/SearchForm.vue
```diff
@@ -76,20 +76,7 @@
                 @noOfResults="changeNoOfResults"
                 :formValues="form"/>
         <section class="section-content pt-3 bg" v-else>
-            <div class="container">
-                <article class="card">
-                    <div class="card-body text-center">
-                        <h5 class="card-title">
-                            GRIIDC Datasets
```
praneethpr, line 83: Removing this would remove the text on GRIIDC search as well, we need a flag variable to switch

mickel1138: Made it switchable.

templates/nas-grp-base.html.twig
```diff
@@ -159,6 +159,10 @@
         });
     </script>
 
+    <script>
+        window.PELAGOS_TEMPLATE_PROPS = {"showSearchTextBlock": false};
```
praneethpr, line 163: Needs to be set to TRUE in the griidc template

fightingtexasaggie: yes, currently the search page doesn't show at all under the griidc template.

mickel1138: I'm just gonna flip it around.

assets/js/vue/components/search/SearchForm.vue
```diff
@@ -76,7 +76,7 @@
                 @noOfResults="changeNoOfResults"
                 :formValues="form"/>
         <section class="section-content pt-3 bg" v-else>
-            <div class="container">
+            <div v-show="!hideSearchTextBlock" class="container">
```
praneethpr, line 79: using the same name as the window variable key is kind of confusing. Should we name the function something else?

praneethpr:
```suggestion
            <div v-show="!displayTextBlock" class="container">
```

assets/js/vue/components/search/SearchForm.vue
```diff
@@ -117,6 +117,15 @@
                 submitted: false,
             }
         },
+        computed: {
+            hideSearchTextBlock: function () {
```
praneethpr, line 121:
```suggestion
            displayTextBlock: function () {
```

PR #677, opened 2020-06-29
Feature/pelagos 4163 add fc aggregation in the search app

praneethpr:
Merged feature/PELAGOS-4164 into this, because both were using the same file and line to add global window variables


assets/js/vue/components/search/ResultSet.vue
```diff
@@ -32,7 +32,8 @@
                     <aside class="col-lg-3">
                         <div class="card card-filter">
                             <Facet :facet-info="results.facetInfo.statusInfo" :facet-name="datasetStatus" v-on="$listeners" :formValues="formValues"/>
-                            <Facet :facet-info="results.facetInfo.fundingOrgInfo" :facet-name="fundingOrg" v-on="$listeners" :formValues="formValues"/>
+                            <Facet :facet-info="results.facetInfo.fundingCycleInfo" :facet-name="fundingCycle" v-on="$listeners" :formValues="formValues" v-if="showFundingCycleFacet()"/>
```
praneethpr, line 35: Send facet title using facet-name prop

assets/js/vue/components/search/SearchForm.vue
```diff
@@ -117,6 +117,15 @@
...
         },
+        computed: {
+            displayTextBlock: function () {
+                if (typeof window.PELAGOS_TEMPLATE_PROPS !== 'undefined') {
+                    return window.PELAGOS_TEMPLATE_PROPS.hideSearchTextBlock;
+                } else {
+                    return false;
+                }
+            }
+        },
```
mickel1138, line 128: Isn't this from another PR, it should show up right?

praneethpr: Yes, that's weird

praneethpr: Did you commit anything after?

mickel1138: This may be some Github weirdness, because this is on `develop`

PR #678, opened 2020-06-30
Used v-bind class to add scrollable class when condition is met

praneethpr:


assets/js/vue/components/search/Facet.vue
```diff
@@ -118,6 +118,12 @@
                 } else {
                     return this.facetInfo;
                 }
+            },
+            facetScrollable: function () {
+                const scrollableClass = 'scrollable-facet';
+                if (this.facetInfo.length > 10) {
```
praneethpr, line 124: Add constant

assets/js/vue/components/search/Facet.vue
```diff
@@ -132,5 +139,9 @@
 </script>
 
 <style scoped>
-
-</style>
\ No newline at end of file
+    .scrollable-facet {
+        height: auto;
+        max-height: 20rem;
```
mickel1138, line 144: Maybe this should be em, but then again, maybe I don't really undestand these unites.

praneethpr: Im not sure either

PR #679, opened 2020-06-30
Feature/pelagos 4165 add funding cycle dropdown

mickel1138:


assets/js/vue/components/FundingCyclesList.vue
```diff
@@ -0,0 +1,58 @@
...
+                this.researchGroupDisabled = true;
+                this.fundingCycles.forEach(fundingCycle => {
+                    if (fundingCycle.id == event.target.value) {
+                        this.researchGroups = fundingCycle.researchGroups;
+                        this.researchGroupDisabled = false;
+                    }
+                })
+            },
+            selectResearchGroup: function(event) {
+                if (event.target.value != "") {
```
praneethpr, line 48: We could just do this to check if value exists
```suggestion
                if (event.target.value) {
```

src/Controller/DefaultController.php
```diff
@@ -30,14 +31,31 @@ class DefaultController extends AbstractController
...
+
+        foreach ($results as $fundingCycle) {
+            $data = array();
+            $data['id'] = $fundingCycle->getId();
+            $data['name'] = $fundingCycle->getName();
+
+            foreach ($fundingCycle->getResearchGroups() as $researchGroup) {
+                $rg['id'] = $researchGroup->getId();
+                $rg['name'] = $researchGroup->getName();
+                $data['researchGroups'][] = $rg;
```
praneethpr, line 51: We could use array_push

src/Controller/DefaultController.php
```diff
@@ -30,14 +31,31 @@ class DefaultController extends AbstractController
...
-            $researchGroups = $this->get('doctrine')->getRepository(ResearchGroup::class)->findBy(array('id' => $fundingOrgFilter->getResearchGroupsIdArray() ));
+            $filter = array('fundingOrganization' => $fundingOrgFilter->getFilterIdArray());
+        }
+
+        $results = $this->get('doctrine')->getRepository(FundingCycle::class)->findBy($filter, array('name' => 'ASC'));
+
+        $fundingCycles = array();
+
+        foreach ($results as $fundingCycle) {
+            $data = array();
```
praneethpr, line 44: I think we could use some other variable name

src/Controller/DefaultController.php
```diff
@@ -30,14 +31,31 @@ class DefaultController extends AbstractController
...
+            $data['id'] = $fundingCycle->getId();
+            $data['name'] = $fundingCycle->getName();
+
+            foreach ($fundingCycle->getResearchGroups() as $researchGroup) {
+                $rg['id'] = $researchGroup->getId();
```
praneethpr, line 49: Same here, expanded form of $rg would be good

assets/css/grp-home.css
```diff
@@ -0,0 +1,19 @@
+.card-product {
```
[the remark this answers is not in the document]
mickel1138: gotta delete it yes, the file I mean.

assets/js/vue/components/FundingCyclesList.vue
```diff
@@ -0,0 +1,58 @@
...
+        },
+        data() {
+            return {
+                researchGroups: [],
+                researchGroupDisabled: true,
+            }
+        },
+        methods: {
+            populateResearchGroups: function(event) {
+                this.researchGroups = [];
```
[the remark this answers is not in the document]
mickel1138: I'm erasing it here. On purpose.

src/Controller/DefaultController.php
praneethpr: Yes, It looks like using $array[] seems to be faster than using array_push(). Didn't know that

assets/js/vue/components/FundingCyclesList.vue
```diff
@@ -0,0 +1,58 @@
...
+        <select @change="populateResearchGroups" class="form-control">
+            <option value="" selected>[Please select a Funding Cycle]</option>
+            <option v-for="fundingCycle in fundingCycles" :value="fundingCycle.id" :key="fundingCycle.id">{{ fundingCycle.name }}</option>
+        </select>
+
+        <label class="col-form-label col-form-label-lg">
+            Research Groups
+        </label>
+
+        <select @change="selectResearchGroup" class="form-control" :disabled="researchGroupDisabled">
```
praneethpr, line 16: Doesn't work if the user wants the 1st research group in the list

PR #680, opened 2020-07-02
Text only changes from NASGRP to GRP

mickel1138:
I think that's all there is.


fightingtexasaggie:
I grepped for nasgrp on the templates...no other instances, just the tagline change Praneeth mentioned

PR #681, opened 2020-07-03
Feature/pelagos 4168 add project director dropdown on grp home page

praneethpr:


assets/js/vue/components/FundingCyclesList.vue
```diff
@@ -30,30 +39,77 @@
...
+        created() {
+            this.populateProjectDirectors();
+        },
+        watch: {
+            selectedProjectDirector: function () {
+               this.openResearchGroupLandingPage(this.selectedProjectDirector);
+            }
+        },
+        filters: {
+            truncateResearchGroupName: function (text) {
```
mickel1138, line 102: Missing length parameter here.

src/Controller/DefaultController.php
```diff
@@ -31,33 +33,70 @@ class DefaultController extends AbstractController
...
+                    $fundingCycleList[] = array(
+                        'id' => $fundingCycle->getId(),
+                        'name' => $fundingCycle->getName(),
+                        'researchGroups' => $this->getResearchGroupsArray($fundingCycle)
+                    );
+                }
+            }
         }
 
-        $fundingCycles = $this->get('doctrine')->getRepository(FundingCycle::class)->findBy($filter, array('name' => 'ASC'));
```
mickel1138, line 39: Not sure how these even worked before, I used FO id's to filter FCs. ¯\_(ツ)_/¯

src/Controller/DefaultController.php
```diff
@@ -31,33 +33,70 @@ class DefaultController extends AbstractController
      */
     public function nasIndex(FundingOrgFilter $fundingOrgFilter)
     {
-        $filter = array();
+        $fundingCycleList = array();
         if ($fundingOrgFilter->isActive()) {
```
mickel1138, line 37: This works great, but what do you do if the filter is NOT active?

assets/js/vue/components/FundingCyclesList.vue
mickel1138, line 102:
```suggestion
            truncateResearchGroupName: function (text, truncateLength) {
```

praneethpr: It is not a required parameter, by default the value is sent as the parameter in the mustache interpolations

src/Controller/DefaultController.php
praneethpr: So, by default should i always get NAS/GRP funding org data? because its anyways inside the grp-home default route

src/Controller/DefaultController.php
```diff
@@ -31,33 +33,70 @@ class DefaultController extends AbstractController
      */
     public function nasIndex(FundingOrgFilter $fundingOrgFilter)
     {
-        $filter = array();
```
mickel1138, line 34: Put the filter back in place the way it was, and move the foreaches below it. The filter did work because of line 36 (deleted 36)

src/Controller/DefaultController.php
mickel1138: I don't think we can make that assumption. Things may change somewhere else, and it should work either way, even tho this will only be used here.

assets/js/vue/components/FundingCyclesList.vue
mickel1138: Yes, but what if I want to truncate each call with a different length, maybe set a default instead then.

praneethpr: Yes, if you look at line 32, I've added a constant for default length

PR #683, opened 2020-07-06
Added person research group repository, changed the usage of options …

praneethpr:
…in the vue component, added buttons to navigate to research group landing page

### DIFF
--- a/assets/js/vue/components/FundingCyclesList.vue
+++ b/assets/js/vue/components/FundingCyclesList.vue
@@ -4,13 +4,21 @@
             <label class="col-form-label-lg">
                 Funding Cycles
             </label>
-            <b-form-select v-model="selectedFundingCycle" :options="fundingCycleOptions" class="w-75"></b-form-select>
+            <b-form-select v-model="selectedFundingCycle" :options="fundingCycleOptions" class="w-75">
+                <template v-slot:first>
+                    <b-form-select-option :value="null" disabled>-- Please select a Funding Cycle --</b-form-select-option>
+                </template>
+            </b-form-select>
 
             <label class="col-form-label-lg">
                 Research Groups
             </label>
             <div class="form-inline">
-                <b-form-select v-model="selectedResearchGroup" :options="researchGroupOptions" :disabled="disableResearchGroups" class="w-75"></b-form-select>
+                <b-form-select v-model="selectedResearchGroup" :options="researchGroupOptions" :disabled="disableResearchGroups" class="w-75">
+                    <template v-slot:first>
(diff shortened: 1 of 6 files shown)


assets/js/vue/components/FundingCyclesList.vue
```diff
@@ -34,67 +33,80 @@
...
                 })
             }
         },
         created() {
+            this.populateFundingCycles();
             this.populateProjectDirectors();
         },
         watch: {
+            selectedFundingCycle: function () {
+                this.populateResearchGroups(this.selectedFundingCycle);
```
mickel1138, line 103: Also disable researchgroup (and maybe select placeholder)

praneethpr: I didn't quite understand what you meant

mickel1138: When you select a different Funding Cycle, or select the placeholder, the research group should be disabled again.

assets/js/vue/components/FundingCyclesList.vue
```diff
@@ -34,67 +45,81 @@
...
-                            } else {
-                                this.projectDirectorIds.push(projectDirector.id);
-                                this.makeProjectDirectorOption(
-                                    researchGroup.id,
-                                    projectDirector.name + ' - ' + researchGroup.shortName
-                                );
-                            }
-                        })
+                    this.fundingCycleOptions.push({
+                        value: fundingCycle.id,
```
mickel1138, line 93: Here one could have used `text-field` and `value-field` in the corresponding b-form-select.

praneethpr: I don't know whether it makes any difference, instead of using value and text, they have used custom names for them. we still have to push them to the options object

src/Controller/DefaultController.php
```diff
@@ -40,6 +41,15 @@ public function nasIndex(FundingOrgFilter $fundingOrgFilter)
 
         $fundingCycles = $this->get('doctrine')->getRepository(FundingCycle::class)->findBy($filter, array('name' => 'ASC'));
 
+        $leadershipPersons = $this->get('doctrine')->getRepository(PersonResearchGroup::class)->getLeadershipPeople();
```
mickel1138, line 44: I technically doesn't matter in what custom repository you write a certain function, in the case: `getLeadershipPeople`.
It just may be unexpected to find this in `PersonResearchGroup`, maybe this should be a function of `ResearchGroup` or `People` (probably research group). It doesn't matter that much, but it will be easier to find, and maybe more logical?

praneethpr: Yeah, we can refactor as an improvement

PR #685, opened 2020-07-08
Feature/pelagos 4140 add pub elements to datacite metadata

praneethpr:


mickel1138:
I think you should have branched this from Develop, not from the release. Or else what good is Develop?

PR #687, opened 2020-07-13
Feature/pelagos 4175 redesign layout for people cards

praneethpr:


assets/js/grp-home.js
```diff
@@ -2,9 +2,12 @@ import Vue from "vue";
 import '../css/grp-home.css';
 import { BootstrapVue, IconsPlugin } from 'bootstrap-vue';
 import FundingCyclesList from "./vue/components/FundingCyclesList";
+import filters from "./vue/utils/filters";
 
 Vue.use(BootstrapVue);
-
+Vue.filter('truncate', (text, length) => {
```
mickel1138, line 8: Is there no way to direct reference this?

assets/js/vue/components/research-group/PeopleTab.vue
```diff
@@ -43,11 +40,25 @@
         column-gap: 1rem !important;
         column-count: 2 !important;
         .card {
-            height: 200px;
+            height: 12em;
         }
     }
+    .card-product {
+        margin-bottom: 1rem;
```
mickel1138, line 47:
```suggestion
        margin-bottom: 1rem;
        transition: .5s;
```
More a more natural transition, also add the transition out.

assets/js/grp-home.js
praneethpr: You mean directly include it in the child component?

mickel1138: When applying the filter, maybe like this: {{ bla bla | filters.truncate(10) }}
Don't know if that's possible. Maybe that's not the way to do it, because then the child needs to know about the filter name.
Or maybe you can make an alias. Either way, it's the same effect. So this works.

assets/js/vue/components/research-group/PeopleTab.vue
```diff
@@ -1,31 +1,20 @@
...
+        <b-card class="card-product"
+                v-for="person in sortedPeople"
+                :key="person.id"
+                :title="person.person.firstName + ' ' + person.person.lastName"
+                :sub-title="person.label">
+            <b-card-text v-tooltip="{
+                            content: person.person.organization,
+                            placement:'top'
+                            }">
+                {{ person.person.organization | truncate(75) }}
```
praneethpr, line 12: @mickel1138: I'm applying the filter like you mentioned

assets/js/grp-home.js
```diff
@@ -2,9 +2,12 @@ import Vue from "vue";
 import '../css/grp-home.css';
 import { BootstrapVue, IconsPlugin } from 'bootstrap-vue';
 import FundingCyclesList from "./vue/components/FundingCyclesList";
+import filters from "./vue/utils/filters";
```
praneethpr, line 5: I moved the truncate filter into a util directory, so that other apps/components can use it, So thats why we need that Vue.filter() in each of the respective JS app files

PR #688, opened 2020-07-17
Doi update fixed when publink doi is deleted

praneethpr:


src/Controller/Api/DatasetPublicationController.php
```diff
@@ -258,19 +259,22 @@ public function linkAction(int $id, Request $request, ObjectPersister $objectPer
...
+     *     methods={"DELETE"},
+     *     defaults={"_format"="json"}
+     *     )
+     *
+     */
+    public function deleteAction(int $id, RabbitPublisher $publisher)
     {
+        $datasetPublication = $this->handleGetOne(DatasetPublication::class, $id);
+        $dataset = $datasetPublication->getDataset();
         $this->handleDelete(DatasetPublication::class, $id);
```
fightingtexasaggie, line 275: What happens if handleDelete fails? Should we still publish the update?

mickel1138, line 275: handleDelete returns the entity which is just deleted.

fightingtexasaggie: Should this instead be catching the possible "not deletable" BadRequestHttpException in its own try/catch block so it won't attempt to publish if the delete didn't happen?

praneethpr: Even if the delete doesn't happen, the publish wouldnt unlink the publication doi from the datacite doi

mickel1138, line 275:
```suggestion
        $datasetPublication = $this->handleDelete(DatasetPublication::class, $id);
        $dataset = $datasetPublication->getDataset();
```

mickel1138: Just a suggestion, only saved one line.
Because handleDelete already return the to be deleted entity.

PR #692, opened 2020-07-22
Feature/pelagos 4182 create research group report

praneethpr:


src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -77,52 +78,152 @@ public function defaultAction(Request $request, EntityHandler $entityHandler, in
...
+     *     name="pelagos_app_ui_reportresearchgroupdatasetstatus_datasetmonitoringreport",
+     *     methods={"GET"}
+     *     )
+     *
+     * @return Response|StreamedResponse A Symfony Response instance.
+     */
+    public function datasetMonitoringReportAction(Request $request, EntityHandler $entityHandler)
+    {
+        // Checks authorization of users
+        if (!$this->isGranted('ROLE_DATA_REPOSITORY_MANAGER')) {
```
mickel1138, line 98: I think the report should be accessible for everyone.

mickel1138: I guess this is the admin version?

praneethpr: Yes, the post action is called by the button on dataset-monitoring which is accessible by everyone

templates/Menus/admin.html.twig
```diff
@@ -37,6 +37,11 @@
                     Report Research Group Dataset Status
                 </a>
             </li>
+            <li class="first odd">
+                <a href="{{ path('pelagos_app_ui_reportresearchgroupdatasetstatus_datasetmonitoringreport') }}">
+                    Dataset Report for Research Group
+                </a>
+            </li>
             <li class="first even"><a href="{{ path('pelagos_app_ui_gomrireport_default') }}">GOMRI Dataset Report(v1)</a></li>
```
mickel1138, line 45: Even tho these classes don't do anything, but this would have be called middle instead of first.

src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -77,52 +78,152 @@ public function defaultAction(Request $request, EntityHandler $entityHandler, in
...
+            array(parent::BLANK_LINE));
 
         //prepare label array
         $labels = array('labels' => array('DATASET UDI',
-          'TITLE',
-          'PRIMARY POINT OF CONTACT',
-          'STATUS',
-          'DATE APPROVED',
-          'DATE REGISTERED'));
+            'TITLE',
```
mickel1138, line 205: Fields missing: DOI, Restricted Status.

praneethpr: This is the version 1 report, which was already there, i just had to refactor it, because we were using almost the same stuff. If you look at Line 292, you can find the labels for the other report which is requested in this ticket

mickel1138: I missed that. sry.

PR #695, opened 2020-07-23
Added csv icon and link to it, changed pdf icon to font awesome icon

praneethpr:


assets/static/js/dataset-monitoring.js
```diff
@@ -144,9 +144,40 @@ function showProjects(by,id) {
...
+}
+
+const addCsvUrl = () => {
+    const csvElementIds = document.getElementsByClassName('awesome-icon-csv');
+    for (let i = 0; i < csvElementIds.length ; i++) {
+        csvElementIds[i].addEventListener("click", function (event) {
+                axios.create({})
+                    .post(Routing.generate("pelagos_app_ui_reportresearchgroupdatasetstatus_post"), {ResearchGroupSelector: csvElementIds[i].dataset.researchGroupId})
+                    .then(response => {
+                        console.log(response)
```
mickel1138, line 171: Do you wanna keep these, or maybe Throw something. Can they go away without breaking?

praneethpr: Yes, i should remove them. It was for testing

templates/DatasetMonitoring/projects.html.twig
```diff
@@ -4,9 +4,13 @@
         <span class="header-inst-name">({{instName}})</span>
     {% endif %}
     {% if not pdf | default %}
-        <a href="/html2pdf?url={{ url(app.request.get('_route'), {'id': id, 'renderer': 'html2pdf'}) }}&format=Letter&orientation=landscape&margin=1cm&download=true&filename={{pdfFilename}}" class="download-pdf" title="save as pdf">
-            <img src="{{ asset('build/images/pdf_icon.png') }}">
-        </a>
+        <span class="awesome-icon awesome-icon-pdf" data-url="/html2pdf?url={{ url(app.request.get('_route'), {'id': id, 'renderer': 'html2pdf'}) }}&format=Letter&orientation=landscape&margin=1cm&download=true&filename={{pdfFilename}}">
```
mickel1138, line 7: Why aren't these just Anchors?

praneethpr: One is a POST request, and I wanted both icons to be consistent. Plus for anchor tags, had to remove the default styling that was getting applied to the font awesome icons

mickel1138: It could be a GET request as well I guess? It took me a second the figure out what the javascript was doing, I feel it's better if we could do without. If that's possible?

mickel1138: Actually, it's probably fine the way it is.

praneethpr:
> Assuming the pelagos_app_ui_reportresearchgroupdatasetstatus_post route is added in another ticket, all seems to be fine. I too am curious why it can be a GET route, but it's fine.

Yes, it can be a GET route, the old research group report and this report in the admin menu uses the same form, need to use extra html form tags.

PR #696, opened 2020-07-23
Feature/pelagos 4183 modernize stats page

mickel1138:


assets/js/stats.js
```diff
@@ -0,0 +1,64 @@
...
+var overviewSections;
+var style = getComputedStyle(document.body);
+var theme = {};
+
+theme.main = style.getPropertyValue('--color-main');
+theme.secondary = style.getPropertyValue('--color-menu');
+theme.dark = style.getPropertyValue('--color-headerMiddle');
+theme.light = style.getPropertyValue('--color-headerTop');
+
+$(document).ready(function() {
```
praneethpr, line 18: Because its a rewrite should we use regular javascript instead of jQuery?

assets/css/stats.css
```diff
@@ -0,0 +1,142 @@
+#stats-tabs {
```
praneethpr, line 1: Should this file be a git move?

assets/js/stats.js
mickel1138: These section is going away in the next ticket.

assets/js/stats.js
```diff
@@ -0,0 +1,64 @@
+const $ = require('jquery');
+global.$ = global.jQuery = $;
```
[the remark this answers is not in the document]
mickel1138: I still not sure how webpack manages globals. I think we should have used "app" as a common app in the base template.

assets/css/stats.css
mickel1138: There is a minor change, but I didn't want to odd thing to happen again. I can delete the other one.

praneethpr: If there are no changes in the file, i think git move should work just fine

assets/js/stats.js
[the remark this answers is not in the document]
praneethpr: Because we use $ in the static/js/ files we need the global, because client side javascript doesn't recognize "require"

PR #697, opened 2020-07-24
Added argument to update doi method

praneethpr:


src/Util/DOIutil.php
```diff
@@ -306,6 +306,7 @@ public function deleteDOI(string $doi)
                 'inactive',
                 '2019',
                 'none supplied',
+                [],
```
mickel1138, line 309: Are these the empty options/publications?

praneethpr: Yes, it expects publications, because the dataset is deleted sending a empty array

PR #699, opened 2020-07-27
Added additional width to containers and breakpoints

praneethpr:


assets/scss/griidc-colors.scss
```diff
@@ -103,3 +103,23 @@ $navbar-light-active-color: color('black');
 $navbar-light-hover-color: theme-color('primary');
 
 $dropdown-link-hover-color: theme-color('primary');
+
```
mickel1138, line 106: Maybe one day we should move the common things at a higher level?

assets/js/vue/ResearchGroupApp.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="container-xl" ref="formContainer">
+    <div class="container" ref="formContainer">
```
mickel1138, line 2: This makes the background narrower than before. The container is also really tight around the content. For example the cards, and the results per page selector. Maybe it should be wider, or need padding?

assets/scss/griidc-colors.scss
praneethpr: Yes maybe we can have a common file for both themes

assets/js/vue/ResearchGroupApp.vue
praneethpr: Sorry, i dint notice the background color thing. I fixed it, not sure about the content being tight. I haven't changed any of that, and one of the things in the ticket was to reduce padding

fightingtexasaggie: I'm not seeing where this is really tight around content. It looks fine in both Firefox and Chrome to me on the search page.

assets/js/vue/components/search/SearchForm.vue
```diff
@@ -66,31 +66,31 @@
                         </div>
                     </b-form>
                 </div>
-            </div>
```
fightingtexasaggie, line 69: Where's this div open?

assets/js/vue/ResearchGroupApp.vue
mickel1138: > I'm not seeing where this is really tight around content. It looks fine in both Firefox and Chrome to me on the search page.

That's been fixed now.

assets/js/vue/components/search/SearchForm.vue
praneethpr: I cross checked, i don't see any orphan div

fightingtexasaggie: just odd how github displayed it I guess

PR #700, opened 2020-07-27
Feature/pelagos 4182 create research group report

praneethpr:


src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -323,4 +327,35 @@ private function getVersionTwoReport(Collection $datasets, array $options): arra
...
     }
+
+    /**
+     * Get custom dataset status string for the version two report.
+     *
+     * @param Dataset $dataset An instance of dataset entity.
+     *
+     * @return string
+     */
+    private function getDatasetStatus(Dataset $dataset): string
```
mickel1138, line 338: This is a necessary evil, the getStatus from the Dataset entity if there only to support another report, and not used for anything else.

One day we shall refactor all the status things.

src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -155,13 +156,16 @@ protected function getData(array $options): array
     {
         $datasets = $options['researchGroup']->getDatasets();
         $reportData = array();
-
+        $defaultHeaders = $this->getDefaultHeaders();
         if (isset($options['version'])) {
             $reportData = $this->getVersionTwoReport($datasets, $options);
+            $defaultHeaders = $this->getDefaultHeaders();
```
fightingtexasaggie, line 162: already have $defaultHeaders above

PR #706, opened 2020-07-31
Made changes to research group controller to accept GET and POST requ…

praneethpr:
…ests

### DIFF
--- a/src/Controller/UI/ReportResearchGroupDatasetStatusController.php
+++ b/src/Controller/UI/ReportResearchGroupDatasetStatusController.php
@@ -35,13 +35,12 @@ class ReportResearchGroupDatasetStatusController extends ReportController
      *
      * @param Request       $request         Message information for this Request.
      * @param EntityHandler $entityHandler   The entity handler.
-     * @param integer       $researchGroupId The identifier for the Research Group subject of the report.
      *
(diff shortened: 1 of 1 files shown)


src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -114,28 +113,31 @@ public function datasetMonitoringReportAction(Request $request, EntityHandler $e
...
+        if ($form->isSubmitted()) {
+            if ($form->isValid()) {
+                $researchGroupId = $request->get('ResearchGroupSelector');
+                return $this->getReport($researchGroupId);
+            }
+        }
+
+        if ($id) {
+            return $this->getReport($id);
+        }
```
mickel1138, line 125: Can we make this into a big statement?
```suggestion
        if (($form->isSubmitted() and if ($form->isValid()) or $researchGroupId ) {
            ) {
                $researchGroupId = $request->get('ResearchGroupSelector');
                return $this->getReport($researchGroupId);
            }
        }
```

mickel1138: You also have to change id in the route above, to researchGroupId. FYI, don't commit this, it may not work.

praneethpr: I think it is good as it is, or else we need to change the variable name in the route config and add an additional check for it

PR #707, opened 2020-08-03
Spike/pelagos 4190 make demo file manager app

praneethpr:


src/Entity/DatasetSubmission.php
```diff
@@ -1062,11 +1062,18 @@ public function __construct(Entity $entity, PersonDatasetSubmissionDatasetContac
 
                 $this->addDatasetLink($newDatasetLink);
             }
-
             // Copy the fileSet
             $newFileset = new Fileset();
             foreach ($entity->getFileset()->getFiles() as $file) {
-                $newFileset->addFile($file);
+                $newFile = new File();
```
mickel1138, line 1068: There is a difference, I think we can just relink a file, and we don't have to copy the entire file entity properties.
So I don't we need this, we can think of the pros and cons. The file doesn't change, just the container for it.

Were there problems by doing it the other way?

praneethpr: Yes, we need to remove the File entity object from the old Fileset entity's object and we need to persist that as well and it is not consistent to things we usually do. If we are going to do it the other way then we should also unlink Fileset object from the old DatasetSubmission object and relink it to the newly constructed one

mickel1138: I remember now. It's kinda messy. But this is how it works.

PR #709, opened 2020-08-03
Feature/pelagos 4189 use new chart component

mickel1138:


assets/js/stats.js
```diff
@@ -0,0 +1,21 @@
...
+import '../css/stats.css';
+
+import 'bootstrap';
+import Vue from 'vue';
+import StatsApp from './vue/Stats.vue';
+
+var style = getComputedStyle(document.body);
+var theme = {};
+
+theme.primary = style.getPropertyValue('--primary');
```
mickel1138, line 13: can go.

assets/js/vue/components/stats/DatasetSizeRanges.vue
```diff
@@ -0,0 +1,74 @@
...
+    DxTitle,
+    DxSubtitle,
+    DxTick,
+    DxLegend
+  }
+};
+</script>
+<style>
+#chart {
+    height: 440px;
```
praneethpr, line 72: We should use %s instead, so that it is responsive

assets/js/stats.js
```diff
@@ -0,0 +1,21 @@
+const $ = require('jquery');
+global.$ = global.jQuery = $;
+
+import '../css/stats.css';
+
+import 'bootstrap';
```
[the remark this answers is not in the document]
mickel1138: I'm just following what was said about bootstrap, you add the stuff you need, regardless if the parent already did it. Because you may not know what was included, and if it changes you know at least that this file needed bootstrap.

assets/js/vue/components/stats/RepositorySummary.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+    <div id="program-resources">
+        <div style="display:table-row;">
+            <div class="summaryLabel" style="display:table-cell;">Repository Summary</div>
+        </div>
+        <div style="display:table-row;">
```
mickel1138, line 6: Table to bootstrap grid.

assets/js/stats.js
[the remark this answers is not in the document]
praneethpr: Yes, webpack will include it only once, but we have created a custom file bootstrap.scss which includes bootstrap and our custom colors, etc which is included in both templates (GRP and GRIIDC). I don't think we need to include it again to show that this page uses it, its redundant

assets/js/vue/Stats.vue
```diff
@@ -0,0 +1,34 @@
...
+    const axios = require('axios');
+    import DatasetOverTime from "./components/stats/DatasetOverTime";
+    import DatasetSizeRanges from "./components/stats/DatasetSizeRanges";
+    import RepositorySummary from "./components/stats/RepositorySummary";
+
+    export default {
+        name: "StatsApp",
+        components: { DatasetOverTime, DatasetSizeRanges, RepositorySummary },
+          methods: {
+            getThemeColor: function(color) {
```
praneethpr, line 25: This method is never used in the parent component, Instead of this I think we could use Vue mixins
https://vuejs.org/v2/guide/mixins.html, which can have a method that gets the theme color

assets/js/vue/components/stats/RepositorySummary.vue
```diff
@@ -0,0 +1,130 @@
...
+        </div>
+    </div>
+</template>
+
+<script>
+    export default {
+        name: "RepositorySummary",
+        data() {
+            return {
+                datasets: '-',
```
praneethpr, line 47: This primarily defines the type of the variable, should we just use 0 here and use v-show=false till the data loads

assets/scss/griidc-colors.scss
```diff
@@ -54,10 +54,10 @@ $gradientStart: #022f3a;
 $gradientEnd:   #4f6e74;
 
 $theme-colors: (
-    "primary":       $boston,
-    "secondary":     $blueBird,
+    "primary":       $dark-cyan,
```
praneethpr, line 57: Is there a reason why we are changing the main template colors in this PR, because it would affect other apps as well

assets/scss/griidc.scss
```diff
@@ -1,53 +1,53 @@
-@import url(https://fonts.googleapis.com/css?family=Lato:400,700,400italic,700italic);
```
praneethpr, line 1: Should be reverted, no changes

assets/scss/nas-colors.scss
```diff
@@ -1,102 +1,102 @@
-@import '~bootstrap/scss/functions';
```
praneethpr, line 1: Same here, no changes. Needs to be reverted

assets/js/vue/Stats.vue
mickel1138: I have to change that a bit anyway, because the color will not be the same "name" for each template.

assets/scss/griidc-colors.scss
mickel1138: Yes, these are GRIIDC colors, and they don't use Boston or Blue Bird, this was a bad copy and paste from the NAS scss.

praneethpr: I think the POs didn't want these colors on Search app and Research group landing page apps, Changing these would change the colors on them.

mickel1138: I see now, that was maybe maybe not the best thing to do, because I already got confused.

assets/scss/griidc-colors.scss
```diff
@@ -1,125 +1,129 @@
...
+    "available":     $green,
+    "submitted":     $blue,
+    "identified":    $gray-600,
+    "restricted":    $red,
+    "remotlyhosted": $yellow,
+    "coldstorage":   $iceBlue,
+    "chart-primary":   $teal,
+    "chart-secondary":   $light-cyan,
+    "chart-alternate":   $dark-green,
+
```
mickel1138, line 70: New colors here, the rest of the file had CRLF change to LF.

assets/js/vue/components/stats/DatasetSizeRanges.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+    <DxChart
+        id="dsr-chart"
+        data-source="/stats/data/dataset-size-ranges"
```
praneethpr, line 4: We could add Routing.generate here i guess, and remove the hard coded routes

assets/js/vue/components/stats/DatasetSizeRanges.vue
```diff
@@ -0,0 +1,88 @@
...
+import {themeFunctions} from '@/vue/mixins/themeFunctions';
+
+import {
+    DxChart,
+    DxSeries,
+    DxLabel,
+    DxConnector,
+    DxValueAxis,
+    DxTitle,
+    DxSubtitle,
```
praneethpr, line 58: Not used

assets/js/vue/components/stats/DatasetSizeRanges.vue
mickel1138: How would you implements that?

praneethpr: Use a data variable here, and define the data variable as 
`data() {
        return {
            datasetSizeUrl: Routing.generate("pelagos_app_ui_stats_getdatasetsizeranges),
        };
    }, `

PR #710, opened 2020-08-04
Removed authorization check

praneethpr:


src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -94,10 +94,6 @@ public function defaultAction(Request $request, EntityHandler $entityHandler)
      */
     public function datasetMonitoringReportAction(Request $request, EntityHandler $entityHandler, int $id = null)
     {
-        // Checks authorization of users
-        if (!$this->isGranted('ROLE_DATA_REPOSITORY_MANAGER')) {
```
mickel1138, line 98: It will still work if this is moved down to the render of the form. And make it a possitive, (don't use !) and an else, to should the adminonly template.

praneethpr: It should work now, i changed it

PR #714, opened 2020-08-05
Feature/pelagos 4202 change facet labels

fightingtexasaggie:
I figured we'll use this "template name" variable again from time to time, so set it for both GRIIDC and GRP.

### DIFF
--- a/assets/js/vue/components/search/ResultSet.vue
+++ b/assets/js/vue/components/search/ResultSet.vue
@@ -70,6 +70,16 @@
 <script>
     import Facet from "./Facet";
     import DatasetRow from "./DatasetRow";
+
+    // Conditional, GRP-specific labels
+    if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName === 'GRP') {
+        var fundingCycleAlias = 'Grant Awards';
+        var researchGroupAlias = 'Projects';
+    } else {
+        var fundingCycleAlias = 'Funding Cycles';
+        var researchGroupAlias = 'Research Groups';
+    }
+
     export default {
         name: "ResultSet",
         components: { DatasetRow, Facet },
@@ -89,7 +99,7 @@
                         queryParam: 'status'
                     },
                     fundingCycle: {
-                        label: 'Funding Cycles',
+                        label: fundingCycleAlias,
                         queryParam: 'fundingCycle'
                     },
                     fundingOrg: {
@@ -97,7 +107,7 @@
                         queryParam: 'fundingOrg'
                     },
                     researchGroup: {
-                        label: 'Research Groups',
+                        label: researchGroupAlias,
                         queryParam: 'researchGroup'
                     }
                 },
--- a/templates/griidc-base.html.twig
+++ b/templates/griidc-base.html.twig
@@ -145,6 +145,11 @@
             };
         });
     </script>
-{% endblock %}
+
+    <script type="text/javascript">
+        window.PELAGOS_TEMPLATE_PROPS = {"BaseTemplateName": "GRIIDC"};
+    </script>
+
+    {% endblock %}
 </body>
 </html>
--- a/templates/nas-grp-base.html.twig
+++ b/templates/nas-grp-base.html.twig
@@ -159,8 +159,8 @@
         });
     </script>
 
-    <script>
-        window.PELAGOS_TEMPLATE_PROPS = {"hideSearchTextBlock": true, "fundingCycleFacet": true};
+    <script type="text/javascript">
+        window.PELAGOS_TEMPLATE_PROPS = {"BaseTemplateName": "GRP", "hideSearchTextBlock": true, "fundingCycleFacet": true};
     </script>
 
 {% endblock %}


templates/griidc-base.html.twig
```diff
@@ -145,6 +145,11 @@
             };
         });
     </script>
-{% endblock %}
+
+    <script type="text/javascript">
```
mickel1138, line 149: This doesn't really do anything.

assets/js/vue/components/search/ResultSet.vue
```diff
@@ -89,15 +90,15 @@
                         queryParam: 'status'
                     },
                     fundingCycle: {
-                        label: 'Funding Cycles',
+                        label: (PELAGOS_TEMPLATE_PROPS.BaseTemplateName === 'GRP' ? "Grant Awards" : "Funding Cycles"),
```
mickel1138, line 93: It's better to move this to methods (line 134), that way you can only check if `PELAGOS_TEMPLATE_PROPS` is defined.
Currently this only works because `BaseTemplateName` is defined in the GRIIDC template, where is has no function at all.

templates/nas-grp-base.html.twig
```diff
@@ -159,8 +159,8 @@
         });
     </script>
 
-    <script>
-        window.PELAGOS_TEMPLATE_PROPS = {"hideSearchTextBlock": true, "fundingCycleFacet": true};
+    <script type="text/javascript">
+        window.PELAGOS_TEMPLATE_PROPS = {"BaseTemplateName": "GRP", "hideSearchTextBlock": true, "fundingCycleFacet": true};
```
mickel1138, line 163: Because `fundingCycleFacet` kinda does the same thing, you could key of that. Maybe even rename it to something common.

praneethpr, line 163: I think we should get rid other variables for individual rendering, and just have BaseTemplateName and have a util js file which gets griidc text or grp text

mickel1138: Each "app" could import some type of language/verbiage file. Or we could use different templates with properties in it.
It's only a few now, but if there will be a lot more we should change it.

fightingtexasaggie: this is only a start

praneethpr: I've created a dummy PR, where I've committed the code of what might be a possible way to do this
https://github.com/griidc/pelagos/pull/717

fightingtexasaggie:
PELAGOS-4206 was combined into this PR as it also needed the template-switch.js and it was suggested combining was preferable than branching from 4202, especially considering how github displays things stale.

assets/js/vue/components/search/ResultSet.vue
```diff
@@ -89,31 +92,31 @@
...
                 currentPage: 1,
                 perPage: this.formValues.perPage,
                 perPageOptions: [
                     { value: 10, text: '10' },
                     { value: 25, text: '25' },
                     { value: 50, text: '50' },
                     { value: 100, text: '100' }
-                ]
+                    ],
+                    showFundingCycleFacet: templateSwitch.isGrpTemplate(),
```
mickel1138, line 119: I'm not a big fan of this method name. It's should probably be called showFundingCycles or something. What is the next template or another template also wants to this, then the GRP part of the method name has no meaning.

assets/js/vue/utils/template-switch.js
```diff
@@ -0,0 +1,26 @@
...
+const griidc = {
+    status: 'Dataset Status',
+    fundingCycle: 'Funding Cycles',
+    researchGroup:  'Research Groups',
+    fundingOrg: 'Funding Organizations'
+}
+
+export default {
+    getLabel: (name) => {
+        if (PELAGOS_TEMPLATE_PROPS === undefined || PELAGOS_TEMPLATE_PROPS.BaseTemplateName === undefined || PELAGOS_TEMPLATE_PROPS.BaseTemplateName === 'GRIIDC') {
```
mickel1138, line 17: undefined doesn't mean GRIIDC. It's means not GRP. What if have another BaseTemplateName of something else, in this case it would assume it's GRP.

assets/js/vue/utils/template-switch.js
```diff
@@ -0,0 +1,26 @@
...
+            return griidc[name];
+        } else {
+            return grp[name];
+        }
+    },
+    isGrpTemplate: () => {
+        return PELAGOS_TEMPLATE_PROPS.BaseTemplateName === 'GRP';
```
mickel1138, line 24: This will fail is `PELAGOS_TEMPLATE_PROPS` is undefined.

assets/js/vue/components/search/SearchForm.vue
```diff
@@ -227,7 +220,8 @@
                 fundingOrg: '',
                 researchGroup: '',
                 fundingCycle: '',
-                perPage: 10
+                perPage: 10,
+                projectDirector: ''
```
mickel1138, line 224: This is not used anywhere.

assets/js/vue/utils/template-switch.js
```diff
@@ -0,0 +1,26 @@
+const grp = {
+    status: 'Dataset Status',
+    fundingCycle: 'Grant Awards',
+    researchGroup:  'Projects',
+    fundingOrg: 'Funding Organizations'
```
mickel1138, line 5: Here add another variable like: showFundingCycles, then you can use that function instead rather than checking if it's a template with a certain name.

assets/js/vue/utils/template-switch.js
mickel1138: Also see comment above.

config/packages/fos_elastica.yaml
```diff
@@ -134,6 +134,11 @@ fos_elastica:
                             properties:
                                 doi:
                                     type: 'text'
+                        projectDirectors:
```
mickel1138, line 137: Why is this diff showing here? This was done in another PR.

mickel1138:
I keep getting `showFundingCycleFacet not a function` error on the console. Search if not working.

assets/js/vue/utils/template-switch.js
fightingtexasaggie: So if it's not GRP, and we don't know what it is, how should we handle it? I thought GRIIDC was always the fallback.

assets/js/vue/utils/template-switch.js
fightingtexasaggie: Do we really need to insure it is set? The templates for griidc and grp set it, and they are not external parameters. How far should we go?

config/packages/fos_elastica.yaml
fightingtexasaggie: probably just a github artifact. I didn't touch this file. I feel it's best to compare outside of github's gui at this point.

assets/js/vue/utils/template-switch.js
mickel1138: Yes, GRIIDC is default, maybe it should be called that. Maybe up front say Template is GRIIDC until I know otherwise.

assets/js/vue/utils/template-switch.js
mickel1138: Yes, we will probably have more templates, these template properties should be optional in my mind. So one doesn't have to code for every scenario.

fightingtexasaggie: I'm getting rid of this and instead created a flag section I added to the template-switch.js utility. I think it'll be cleaner, but we always have to handle the undefined cases.

assets/js/vue/utils/template-switch.js
fightingtexasaggie: I added a section for flags, separate of the labels section and a method to get the flag value in whatever template that is in effect.

assets/js/vue/utils/template-switch.js
```diff
@@ -0,0 +1,60 @@
...
+
+export default {
+    getLabel: (name) => {
+        if (PELAGOS_TEMPLATE_PROPS === undefined) {
+            return undefined;
+        } else if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName === undefined) {
+            return undefined;
+        } else if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName === 'GRP') {
+            return grpLabels[name];
+        } else if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName = 'GRIIDC') {
```
praneethpr, line 30: '===' is missing

assets/js/vue/utils/template-switch.js
```diff
@@ -0,0 +1,60 @@
...
+        }
+    },
+    getFlag: (name) => {
+        if (PELAGOS_TEMPLATE_PROPS === undefined) {
+            return undefined;
+        } else if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName === undefined) {
+            return undefined;
+        } else if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName === 'GRP') {
+            return grpFlags[name];
+        } else if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName = 'GRIIDC') {
```
praneethpr, line 43: '===' is missing

assets/js/vue/utils/template-switch.js
```diff
@@ -0,0 +1,60 @@
...
+    researchGroup:  'Research Groups',
+    fundingOrg: 'Funding Organizations',
+};
+const griidcFlags = {
+    showFundingCycles: true,
+    showProjectDirector: false
+};
+
+export default {
+    getLabel: (name) => {
```
praneethpr, line 23: Do we need these functions which all do the same thing?, Could we just have one function which checks if it is GRIIDC or GRP and return the variable value/text based on it

PR #715, opened 2020-08-05
Added number of download to stats page.

mickel1138:


src/Repository/LogActionItemRepository.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $qb = $this->createQueryBuilder('log')
+            ->select('COUNT(log.id)')
+            ->where('log.subjectEntityName = ?1')
+            ->andWhere('log.actionName = ?2')
+            ->setParameter(1, 'Pelagos\Entity\Dataset')
+            ->setParameter(2, 'File Download');
+
+        if ($this->fundingOrgFilter->isActive()) {
+            $researchGroupIds = $this->fundingOrgFilter->getResearchGroupsIdArray();
```
praneethpr, line 53: So this could mean that, a GRP dataset that is downloaded from the GRIIDC site would also show up on GRP stats page

mickel1138: That's true, not sure what the desired behavior is. We don't track from which "site" it was downloaded.

wwnick: That is ok, the GRP dataset is the same as the GRIIDC dataset, we just want the total counts.  Same with views if we implement them.  I could see adding something to the download report that indicates which data landing page it was downloaded from but that doesn't affect this.

mickel1138: > That is ok, the GRP dataset is the same as the GRIIDC dataset, we just want the total counts. Same with views if we implement them. I could see adding something to the download report that indicates which data landing page it was downloaded from but that doesn't affect this.

Who this, new phone!?

wwnichols: I don't know why I have multiple github accounts but you can be sure I will always be logged into the wrong one

fightingtexasaggie: If we ever wanted counts per "site" in a report, we'd need to add it to the log action item payload.

PR #716, opened 2020-08-05
Feature/pelagos 4206 conditionally use grp labels

fightingtexasaggie:
Review 4202 first as this uses changes I made there.

### DIFF
--- a/assets/js/vue/components/research-group/OverviewTab.vue
+++ b/assets/js/vue/components/research-group/OverviewTab.vue
@@ -13,7 +13,7 @@
                 {{ overview.fundingCycle.fundingOrganization.name }}
             </a>
         </p>
-        <h6 class="font-weight-bold">Funding Cycle</h6>
+        <h6 class="font-weight-bold">{{ getFcLabel() }}</h6>
         <p>
             <a :href="overview.fundingCycle.url">
                 {{ overview.fundingCycle.name }}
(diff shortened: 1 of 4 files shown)


assets/js/vue/components/search/ResultSet.vue
```diff
@@ -89,15 +99,15 @@
                         queryParam: 'status'
                     },
                     fundingCycle: {
-                        label: 'Funding Cycles',
+                        label: fundingCycleAlias,
```
mickel1138, line 102: This changed code in another pull request. Remove these here, and redo them in the other pull request. Also see noted in that pull request.

assets/js/vue/components/research-group/OverviewTab.vue
```diff
@@ -13,7 +13,7 @@
                 {{ overview.fundingCycle.fundingOrganization.name }}
             </a>
         </p>
-        <h6 class="font-weight-bold">Funding Cycle</h6>
+        <h6 class="font-weight-bold">{{ getFcLabel() }}</h6>
```
praneethpr, line 16: I don't think GRIIDC users can get to this page, even if we want a conditional rendering i think it could all go into a JS util file

fightingtexasaggie:
Not needed. Work done in 4202 instead.

PR #717, opened 2020-08-06
A possible way to have the text in a separate file

praneethpr:


assets/js/vue/utils/template-switch.js
```diff
@@ -0,0 +1,26 @@
...
+const griidc = {
+    status: 'Dataset Status',
+    fundingCycle: 'Funding Cycles',
+    researchGroup:  'Research Groups',
+    fundingOrg: 'Funding Organizations'
+}
+
+export default {
+    getLabel: (name) => {
+        if (PELAGOS_TEMPLATE_PROPS.BaseTemplateName === 'GRP') {
```
mickel1138, line 17: On these I would just make sure it's not `undefined`.

PR #719, opened 2020-08-07
Feature/pelagos 4205 add project director aggregation to elastic search query

praneethpr:


praneethpr:
> Tested and reviewed code, looks good, but we need to sort project directors by count, then last name per ticket

Made that change

PR #722, opened 2020-08-12
Feature/pelagos 4218 change route parameter to string

praneethpr:


praneethpr:
> couple of missed annotations in Controller/Api/EntityController.php and Controller/Api/PersonResearchGroupController.php
> 
> Also, should the download controller be modified the same way, so it won't break if someone throws a non-int at it? (Controller/DownloadController.php)

Fixed those

mickel1138:
I am very unclear why all these changes had to be made?

mickel1138:
Please only make changes needed to satisfy this ticket, and do the other changes in another ticket. Some of these changes will "break" (behavior), so we have to fix them later.

src/Controller/UI/ResearchGroupController.php
```diff
@@ -84,15 +87,15 @@ public function defaultAction(EntityHandler $entityHandler, int $id = null)
     /**
      * The Research Group Ladning page action.
      *
-     * @param integer       $id            The id of the entity to retrieve.
+     * @param string $id The id of the entity to retrieve.
      *
      * @throws NotFoundHttpException When the research group was not found.
      *
      * @Route("/research-group/about/{id}", name="pelagos_app_ui_researchgroup_about")
```
mickel1138, line 94: Use requirment annotation instead, then type changes or not needed.
https://symfony.com/doc/4.4/routing.html#parameters-validation

PR #723, opened 2020-08-14
Feature/pelagos 4219 add new research group report

praneethpr:


src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -175,19 +257,28 @@ protected function getData(array $options): array
...
+    private function createCsvReportFileName(string $researchGroupName, string $researchGroupId, int $version)
     {
         $nowDateTimeString = date(self::REPORTFILENAMEDATETIMEFORMAT);
-        $researchGroupNameSubstring = substr($researchGroupName, 0, self::MAXRESEARCHGROUPLENGTH);
-        $tempFileName = $researchGroupNameSubstring . '_' . $researchGroupId
-            . '_'
+
+        if ($version === self::REPORT_VERSION_THREE) {
+            $tempFileName = $researchGroupName . '_'
             . $nowDateTimeString
```
fightingtexasaggie, line 272: PO wanted a yyyy-mm-dd HH:mm format, this results in a dd-mm-yyyy HH:mm format. Check with PO if this is ok as is though.

praneethpr: No, i think it results in Y-m-d.

praneethpr: This line is for the filename timestamp, i get it you were talking about timestamp in one of the date fields and you were right, it is not in the right format. will change it

PR #729, opened 2020-08-18
Bugfix/pelagos 4222 add missing parm in getstatistics method

fightingtexasaggie:


mickel1138:
I wonder how it got in this state, maybe a merge conflict?

praneethpr:
> I wonder how it got in this state, maybe a merge conflict?

I don't think we ever tested this part

mickel1138:
> > I wonder how it got in this state, maybe a merge conflict?
> 
> I don't think we ever tested this part

I see now, that's the JSON function, I didn't do that one. Who uses it? I guess that drupal front thing.

src/Controller/UI/StatsController.php
```diff
@@ -105,13 +105,14 @@ private function getStatistics(?int &$totalDatasets, ?string &$totalSize, ?int &
      */
     public function getStatisticsJson()
     {
-        $this->getStatistics($totalDatasets, $totalSize, $peopleCount, $researchGroupCount);
+        $this->getStatistics($totalDatasets, $totalSize, $peopleCount, $researchGroupCount, $totalDownloadCount);
```
mickel1138, line 108: This function has ? in front of the parameters, but also & (byref), the last parameters should be optional right?
So I looked into this, and the ? type is only for the type, so if the function above was called with the last parameter set to null, it would have worked. To truly make the function parameters optional you must add `=null` so the question marks here do kinda nothing.

praneethpr: The question mark just accepts null variable to be passed, This would work
$this->getStatistics($totalDatasets, $totalSize, $peopleCount, $researchGroupCount, null);
but if you didn't have the "?" operator in the parameter, then it would break

fightingtexasaggie: I'm not sure why we implemented this thing by reference to begin with. Also, I believe the only thing that currently uses this json route is drupal for griidc.

mickel1138: > The question mark just accepts null variable to be passed, This would work
> $this->getStatistics($totalDatasets, $totalSize, $peopleCount, $researchGroupCount, null);
> but if you didn't have the "?" operator in the parameter, then it would break

Have you tried this, it doesn't work.

praneethpr: Yes, you are right. It always need to be a variable. 
`$blah = null;
$this->getStatistics($totalDatasets, $totalSize, $peopleCount, $researchGroupCount, $blah);
`

fightingtexasaggie: The annotation ends up leaving a bit to be desired, as the |null in this case _only_ refers to the value of a still-mandatory argument being set to null, because in this situation it is being set by the function itself to a type. I guess this is a side-effect of an untyped language.

PR #734, opened 2020-08-24
Feature/pelagos 4223 upload file api crud

praneethpr:


composer.json
```diff
@@ -9,13 +9,15 @@
...
         "ext-iconv": "*",
+        "ext-json": "*",
         "doctrine/doctrine-bundle": "^1.6",
         "doctrine/doctrine-fixtures-bundle": "^2.3",
         "doctrine/doctrine-migrations-bundle": "^1.2",
         "doctrine/orm": "^2.5",
         "friendsofsymfony/elastica-bundle": "^5.1.1",
         "friendsofsymfony/jsrouting-bundle": "^2.1",
         "friendsofsymfony/rest-bundle": "^2.5",
+        "gedmo/doctrine-extensions": "^2.4",
```
mickel1138, line 20: This is an interesting library, but do we need it?

praneethpr: Used to normalize file names, saw this on symfonycasts

config/services.yaml
```diff
@@ -216,3 +216,7 @@ services:
     App\Util\FundingOrgFilter:
         bind:
             array $fundingOrgs: "%env(json:FILTER_BY_FUNDING_ORG)%"
+
+    App\Util\FileUploader:
+        bind:
+            $homedirPrefix: "%env(HOMEDIR_PREFIX)%"
```
mickel1138, line 222: This is ok for now, but the have to make a good set of variables that are named relevant to what they do.

src/Controller/Api/FilesController.php
```diff
@@ -0,0 +1,112 @@
...
+     *     name="pelagos_api_post_files_dataset_submission",
+     *     methods={"POST"},
+     *     defaults={"_format"="json"}
+     *     )
+     *
+     * @return Response The result of the post.
+     */
+    public function postFiles(Request $request, FileUploader $fileUploader, EntityManagerInterface $entityManager, string $id)
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
```
mickel1138, line 84: This just means you are logged in, we may want to consider a voter that allowed uploading for certain DS's for certain people.

src/Util/FileUploader.php
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * File uploader service.
+     *
+     * @param Request $request Symfony request instance.
+     *
+     * @return array
+     */
+    public function upload(Request $request): array
+    {
+        dump($request);
```
mickel1138, line 41: debug code.

src/Util/FileUploader.php
```diff
@@ -0,0 +1,122 @@
...
+        $totalChunks = $request->get('dztotalchunkcount');
+        $chunkIndex = $request->get('dzchunkindex');
+        /** @var UploadedFile $uploadedFile */
+        $uploadedFile = $request->files->get('file');
```
mickel1138, line 45: Can also use filebag?

src/Util/FileUploader.php
```diff
@@ -0,0 +1,122 @@
...
+        if ($totalChunks > 1) {
+            $chunksFolder = $this->chunksDirectory . DIRECTORY_SEPARATOR . $uuid;
+            $this->isFolder($chunksFolder);
+            $isChunk = true;
+            $uploadedFile->move(
+                $chunksFolder,
+                $chunkIndex
+            );
+
+            if ((int)$totalChunks === ($chunkIndex + 1)) {
```
mickel1138, line 62: You can't assume that the chunkIndex will be in order. It is possible that the last chunk is very small and will finish first.

src/Util/FileUploader.php
praneethpr: I'll try it, should work i guess

src/Util/FileUploader.php
praneethpr: So, we should check the size of the file uploaded with the total size of the actual file? or something like that along with this?

mickel1138: We should probably check the number of chunks in the UUID folder, and if the last part completes, then assemble the chunks.

src/Util/FileUploader.php
praneethpr: Not sure, if there is getFileBag function. I think this is the right way to get it

mickel1138: filebag is in the request, but it's an option to use.

templates/DatasetReview/submission-form.html.twig
```diff
@@ -83,8 +83,11 @@
             </div>
 
             <div id="ds-submit" class="ds-metadata">
+                <div id="dropzone-uploader" class="file-dropzone-area">
```
mickel1138, line 86: Move this to the filemanager component, and use the same ID.
Also disable previewTemplate.

assets/js/file-manager.js
```diff
@@ -29,4 +35,35 @@ if (fileManagerElement) {
...
     });
+
+    let myDropzone = new Dropzone("div#dropzone-uploader", {
+        url: Routing.generate('pelagos_api_post_files_dataset_submission') + "/" + datasetSubmissionId,
+        chunking: true,
+        chunkSize: 1024*1024,
+        forceChunking: false,
+        parallelChunkUploads: true,
+        retryChunks: true,
+        retryChunksLimit: 3,
```
mickel1138, line 46: Also add maxFileSize.

mickel1138 (approved):

fightingtexasaggie (approved):
discussed

PR #736, opened 2020-08-25
Misc/grp aware deploy

fightingtexasaggie:
I updated this script to deploy to both sites from a single run. This will make it harder to ever get grp and griidc out of sync versionwise. It also now checks for invalid branches and local changes, and won't deploy to either site if found.


mickel1138:
How is PR ever going to get to Master?

bin/deploy
```diff
@@ -1,42 +1,79 @@
...
+# check for local changes
+cd /opt/pelagos
+checkdiff
+cd /opt/grp
+checkdiff
+
+# PELAGOS
+cd /opt/pelagos
+git fetch
+gitresult=`git checkout $branch 2>&1`
```
praneethpr, line 57: Can we make the Git Tag to be a argument that is being passed into the command? just like PELAGOS/NAS

PR #737, opened 2020-08-26
Feature/pelagos 4214 update organizational names for grp site

fightingtexasaggie:
I figured this was the simplest solution for these legacy twig products, so just matched the existing template name we were already importing and made it fail safe if not set. Too brutish?


templates/Dataland/summary.html.twig
```diff
@@ -90,7 +90,11 @@
             <tr>
                 <td valign="top">
                     {% if dataset.researchGroup.fundingCycle %}
-                        <b>Funding Source:</b><br/>
+                        {% if custom_base_template|default == 'nas-grp-base.html.twig' %}
```
praneethpr, line 93: Do we want to use this function doesTwigFileExist() from Twig/Extensions class?

praneethpr:
Develop also needs to be merged from old release branch

templates/Dataland/summary.html.twig
fightingtexasaggie: that file will exist either way, but this is what the variable gets set to, so I'm just checking for that. In the event that it does not get set, the |default == 'string' won't match, which will still be the default case.

praneethpr: Yes, should we create a twig extension filter to do something similar to what we did with the Vue JS

assets/static/js/tree.js
```diff
@@ -50,7 +50,13 @@ function insertTree(tree) {
         treeHTML += '            <select id="treetype-selector" onchange="' + on_filter_by_change + 'trees[\'' + tree.name + '\'].selected=null;trees[\'' + tree.name + '\'].type=this.value;updateTree(trees[\'' + tree.name + '\']);">';
         treeHTML += '                <option value="ra"';
         if (tree.type == "ra") treeHTML += ' selected';
-        treeHTML += '>Research Award</option>';
+        if (typeof PELAGOS_TEMPLATE_PROPS !== 'undefined'
```
praneethpr, line 53: We can still use the utils/template-switch.js, we need to add it to the 'assets/js/app.js' and 'assets/js/nas-app.js'

`import templateSwitch from 'vue/utils/template-switch';
global.templateSwitch = templateSwitch;`


and then we can use the method `templateSwitch.getProperty(name)`

praneethpr (dismissed):
Tested and works fine

praneethpr (dismissed):

PR #739, opened 2020-08-27
Yarn upgrade to newer package versions

praneethpr:


mickel1138:
This also fixed the serializer-javascript and lodash security issues.

mickel1138:
I got `  WARNING   Webpack Encore requires version ^9.0.1 of sass-loader, but your version (7.3.1) is too old. The related feature will probably *not* work correctly.`
I think we can upgrade to 9, Webpack is the only one that uses it.

praneethpr:
> Tested key components, and everything works fine.
> Would recommend upgrading to sass-loader 9, I upgraded and test this, and there are not issues.

Should I do it in this PR itself?

mickel1138:
> > Tested key components, and everything works fine.
> > Would recommend upgrading to sass-loader 9, I upgraded and test this, and there are not issues.
> 
> Should I do it in this PR itself?

Yes, we can do it here.

praneethpr:
Updated the sass-loader, didn't know why we were using such an old version.

mickel1138:
> Updated the sass-loader, didn't know why we were using such an old version.

That's the version that was suggested at the time we started using this. I don't remember the source.

PR #740, opened 2020-09-01
Feature/pelagos 4189 use new chart component

mickel1138:


mickel1138:
> Tested and works fine. Shouldn't the image files be moved instead of create and delete?

That's my mistake, and a weird artifact of what I did.

PR #741, opened 2020-09-01
Misc/grp consumer autostart

fightingtexasaggie:
take a look, then we can merge into develop and master. We need the grp consumers to be able to come up by themselves.


mickel1138 (approved):
Reviewed with developer.

praneethpr (approved):

PR #748, opened 2020-09-15
Misc/spike flysystem

praneethpr:


assets/js/file-manager.js
```diff
@@ -37,10 +37,10 @@ if (fileManagerElement.dataset.id) {
     });
 
     let myDropzone = new Dropzone("div#dropzone-uploader", {
-        url: Routing.generate('pelagos_api_post_files_dataset_submission') + "/" + datasetSubmissionId,
+        url: Routing.generate('pelagos_api_post_chunks'),
         chunking: true,
         chunkSize: 1024*1024,
```
praneethpr, line 42: Change it to 10000000

src/Controller/Api/FilesController.php
```diff
@@ -61,32 +61,28 @@ public function getFiles(int $id)
     }
```
mickel1138, line 61: Maybe this should be the uploadController?

mickel1138 (approved):
Looks good, flysystem works.

src/Util/FileUploader.php
```diff
@@ -2,96 +2,78 @@
...
+
+    /**
+     * Relative path for files folder.
+     */
+    const FILES_DIRECTORY = 'upload/files';
+
+    /**
+     * FileUploader constructor.
+     *
+     * @param string              $homedirPrefix    Home directory path prefix.
```
mickel1138, line 38: Not really using these anymore right?

src/Controller/Api/FilesController.php
praneethpr: Yes, Makes sense to rename it

src/Util/FileUploader.php
praneethpr: But we need some kind of directory prefix

PR #749, opened 2020-09-16
Feature/pelagos 4240 add sort filter to the main query

praneethpr:


config/packages/fos_elastica.yaml
```diff
@@ -139,6 +139,10 @@ fos_elastica:
                             properties:
                                 id:
                                     type: 'keyword'
+                        sortingDateForDisplay:
```
praneethpr, line 142: Couldn't think of a better name for this variable, let me know if you guys think of something, will change it

mickel1138: Maybe it's time to bring ApprovalDate to the Dataset entity?

praneethpr: What do you mean by ApprovalDate?

mickel1138: I just wanted to add "some" date up to the dataset level, but it doesn't seem pratical.

mickel1138 (approved):
Works. Good change.

fightingtexasaggie:
Per the ticket, should all the accepted datasets group first? I know there was some discussion about this in the ticket, but I'm not  sure it was fully resolved.

praneethpr:
> Per the ticket, should all the accepted datasets group first? I know there was some discussion about this in the ticket, but I'm not sure it was fully resolved.

No grouping of datasets, just display based on sort order

PR #750, opened 2020-09-17
Added HashHandler stuff.

mickel1138:


src/MessageHandler/HashFileHandler.php
```diff
@@ -2,14 +2,75 @@
...
+    }
+
+    /**
+     * Invoke function to hash a file.
+     */
+    public function __invoke(HashFile $hashFile)
     {
-        //
+        $fileId = $hashFile->getFileId();
+        $this->logger->info(sprintf('Processing File with ID: %d', $fileId));
```
praneethpr, line 60: I think you need double quotes if you are having `%variables`

src/MessageHandler/HashFileHandler.php
```diff
@@ -2,14 +2,75 @@
...
+        $file = $this->fileRepository->find($fileId);
+        if ($file instanceof File) {
+            $filePath = $file->getFilePath();
+            try {
+                $fileHash = hash_file('sha256', 'bla');
```
praneethpr, line 65: You forgot to replace 'bla' with $filePath

src/MessageHandler/HashFileHandler.php
mickel1138: Not for sprintf. Only when using `$variables`

src/MessageHandler/HashFileHandler.php
mickel1138: Yes, that was for testing! :-D

src/MessageHandler/HashFileHandler.php
```diff
@@ -2,14 +2,75 @@
...
+     *
+     * @param EntityHandler   $entityHandler           The entity handler.
+     * @param FileRepository  $fileRepository          The file Repository.
+     * @param LoggerInterface $datasetFileHasherLogger Name hinted dataset_file_hasher logger.
+     */
+    public function __construct(EntityManagerInterface $entityManager, FileRepository $fileRepository, LoggerInterface $datasetFileHasherLogger)
+    {
+        $this->entityManager = $entityManager;
+        $this->fileRepository = $fileRepository;
+        $this->logger = $datasetFileHasherLogger;
```
praneethpr, line 51: I think this needs a bind in the servics.yaml 
`        bind:
            Symfony\Bridge\Monolog\Logger $logger: '@monolog.logger.dataset_file_hasher'`

mickel1138: No, this is a new shortcut you can use. see: https://symfony.com/doc/current/logging/channels_handlers.html#how-to-autowire-logger-channels

praneethpr: That's pretty cool 👍

src/MessageHandler/HashFileHandler.php
```diff
@@ -2,14 +2,75 @@
...
+    }
+
+    /**
+     * Invoke function to hash a file.
+     */
+    public function __invoke(HashFile $hashFile)
```
praneethpr, line 57: Can update the PhpDocblock for this

src/MessageHandler/HashFileHandler.php
```diff
@@ -2,14 +2,75 @@
 
 namespace App\MessageHandler;
 
-use App\Message\HashFile;
-
+use Doctrine\ORM\EntityManagerInterface;
+use League\Flysystem\FilesystemInterface;
```
praneethpr, line 6: This is not used

src/MessageHandler/HashFileHandler.php
```diff
@@ -2,14 +2,75 @@
...
+        $fileId = $hashFile->getFileId();
+        $this->logger->info(sprintf('Processing File with ID: %d', $fileId));
+        $file = $this->fileRepository->find($fileId);
+        if ($file instanceof File) {
+            $filePath = $file->getFilePath();
+            try {
+                $fileHash = hash_file('sha256', $filePath);
+            } catch (\Exception $e) {
+                $this->logger->error(sprintf('Unable to hash file. Message: %s', $e->getMessage()));
+                return;
```
mickel1138, line 68: I decided to do a return on error, because if the file can't be hashed for whatever reason, there is no need to retry? Whadda you guys think?

praneethpr: Do we want to create an Exception class, so that based on the type of error we can either put it in the queue or return void?

praneethpr (approved):
Looks good.

src/MessageHandler/HashFileHandler.php
mickel1138: It just returns a general exception, and creating another class won't do much good. Maybe we should see how it goes and then see if these exceptions need to be retried.

PR #754, opened 2020-09-21
Feature/pelagos 4243 add delete file worker and delete api

praneethpr:


src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -352,6 +353,45 @@ public function getUploadedFilesAction(int $id)
         return array($fileInfo);
     }
 
+    /**
```
praneethpr, line 356: This whole block isn't anything new, just moved it here. Because it uses DatasetSubmission Entity to get all the related files of a Fileset

src/Controller/Api/UploadController.php
```diff
@@ -2,98 +2,113 @@
```
praneethpr, line 2: Moved the existing FilesController on top of the old UploadController, didn't change anything

config/packages/monolog.yaml
```diff
@@ -1,5 +1,5 @@
 monolog:
-    channels: ['mailer', 'filer', 'retriever', 'dataset_file_hasher','doi_issue', 'create_homedir']
+    channels: ['mailer', 'filer', 'retriever', 'dataset_file_hasher','doi_issue', 'create_homedir', 'delete_file']
```
praneethpr, line 2: Because its an async process, I thought we should have a logger for it

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,44 @@
...
+use App\Repository\FileRepository;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\Messenger\MessageBusInterface;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * CRUD API for File Entity.
+ */
+class FileController extends EntityController
```
mickel1138, line 16: I don't think we should be extending EntityController anymore. It's based around entityHandler, we should take care of our own security and permissions. You can extend the AbstractFOSRestController if so desired.

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -352,6 +353,45 @@ public function getUploadedFilesAction(int $id)
...
+     * @View()
+     *
+     * @return array The list of uploaded files.
+     */
+    public function getFiles(int $id)
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
+
+        $fileData = array();
+        $datasetSubmission = $this->handleGetOne(DatasetSubmission::class, $id);
```
mickel1138, line 378: Don't use entityHandler anymore, use a/the repository instead.

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -352,6 +353,45 @@ public function getUploadedFilesAction(int $id)
...
+
+        if ($datasetSubmission->getFileset() instanceof Fileset) {
```
mickel1138, line 380: Also this logic can go to the repository.

src/Controller/Api/FileController.php
praneethpr: Changed that to use AbstractFOSRestController

src/Controller/Api/DatasetSubmissionController.php
praneethpr: This whole block was just moved over from the other File, didn't add it in this pull request, we can do it but i feel it shouldn't be part of this PR

src/Controller/Api/UploadController.php
```diff
@@ -2,98 +2,113 @@
...
+     * @return Response The result of the post.
      */
-    public function postAction(Request $request)
+    public function combineChunks(Request $request, FileUploader $fileUploader, EntityManagerInterface $entityManager, string $id)
     {
         $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
-        return $this->uploadHandler->handleUpload($request);
+        try {
+            $fileMetadata = $fileUploader->combineChunks($request);
+            $this->updateFileEntity($fileMetadata, $entityManager, $id);
```
mickel1138, line 78: I still don't think combineChunks should be responsible for updating the entity, it should just return the path where the file was uploaded. Then the combineChunks event should tell the submission api to add a file. We could do this in a later PR.

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function deleteFile(string $id, FileRepository $fileRepository, MessageBusInterface $messageBus)
+    {
+        $file = $fileRepository->find((int)$id);
+
+        if ($file instanceof File) {
+            $filePath = $file->getFilePath();
+            $deleteFileMessage = new DeleteFile($filePath);
+            $messageBus->dispatch($deleteFileMessage);
+            $this->handleDelete(File::class, (int)$id);
```
mickel1138, line 40: handleDelete is from entityHandlerControler, we should stop using it. That means we are going to need a voter here.

praneethpr: Should we create a ticket for that

mickel1138: Yes, good idea. We should think of which voters we need.

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,56 @@
...
+    public function deleteFile(string $id, FileRepository $fileRepository, MessageBusInterface $messageBus, EntityManagerInterface $entityManager)
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
+        $file = $fileRepository->find((int)$id);
+
+        if ($file instanceof File) {
+            $filePath = $file->getFilePath();
+            $deleteFileMessage = new DeleteFile($filePath);
+            $messageBus->dispatch($deleteFileMessage);
+            $entityManager->remove($file);
```
mickel1138, line 42: Don't just remove enity. Use datasetSubmission->fileSet->removeFIle();

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,56 @@
...
+     * @param FileRepository         $fileRepository File entity Repository instance.
+     * @param MessageBusInterface    $messageBus     Symfony messenger bus interface instance.
+     * @param EntityManagerInterface $entityManager  Entity manager interface instance.
+     * @Route("/api/file/{id}", name="pelagos_api_datasets_delete", methods={"DELETE"}, defaults={"_format"="json"})
+     *
+     * @return Response A response object with an empty body and a "no content" status code.
+     */
+    public function deleteFile(string $id, FileRepository $fileRepository, MessageBusInterface $messageBus, EntityManagerInterface $entityManager)
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
```
mickel1138, line 35: Try using @granted ? Does it work?

src/Controller/Api/FileController.php
praneethpr: Changes made

src/Controller/Api/FileController.php
praneethpr: Yes changed that

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param FileRepository         $fileRepository File entity Repository instance.
+     * @param MessageBusInterface    $messageBus     Symfony messenger bus interface instance.
+     * @param EntityManagerInterface $entityManager  Entity manager interface instance.
+     * @Route("/api/file/{id}", name="pelagos_api_datasets_delete", methods={"DELETE"}, defaults={"_format"="json"})
+     *
+     * @IsGranted("IS_AUTHENTICATED_FULLY")
+     *
+     * @return Response A response object with an empty body and a "no content" status code.
+     */
+    public function deleteFile(string $id, FileRepository $fileRepository, MessageBusInterface $messageBus, EntityManagerInterface $entityManager)
```
mickel1138, line 37: Instead `string $id, FileRepository $fileRepository` use 'File $file' this way it will check the entity for you, get it if it exists, or throw an error if it is not there.

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $file = $fileRepository->find((int)$id);
+
+        if ($file instanceof File) {
```
mickel1138, line 41: This logic would also not be needed then.

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Delete a Dataset and associated Metadata and Difs.
+     *
+     * @param string                 $id             The id of the File to delete.
+     * @param FileRepository         $fileRepository File entity Repository instance.
+     * @param MessageBusInterface    $messageBus     Symfony messenger bus interface instance.
+     * @param EntityManagerInterface $entityManager  Entity manager interface instance.
+     * @Route("/api/file/{id}", name="pelagos_api_datasets_delete", methods={"DELETE"}, defaults={"_format"="json"})
+     *
+     * @IsGranted("IS_AUTHENTICATED_FULLY")
```
mickel1138, line 33: Here you can you add the subject `file` to check for permissions of the file. Can they actually delete it?

src/Controller/Api/FileController.php
```diff
@@ -0,0 +1,54 @@
...
+     * Delete a file.
+     *
+     * @param File                   $file           File entity instance.
+     * @param MessageBusInterface    $messageBus     Symfony messenger bus interface instance.
+     * @param EntityManagerInterface $entityManager  Entity manager interface instance.
+     *
+     * @Route("/api/file/{id}", name="pelagos_api_datasets_delete", methods={"DELETE"}, defaults={"_format"="json"})
+     *
+     * @IsGranted("IS_AUTHENTICATED_FULLY")
+     * @IsGranted("CAN_DELETE", subject="file")
```
mickel1138, line 32: There is not voter who supports the file entity, so no one will vote on this, which is good for now.

mickel1138 (approved):
Everything looks in order.  Needs some integration.

PR #755, opened 2020-09-22
Feature/pelagos 4242 virus scanner

fightingtexasaggie:
This can be tested by placing the virus test file in the location of one of the datafiles. It’s easiest to just run it the pelagos command (which will que a virus scan message similarly to the  filer will in the future), let it fail, then put the file to be scanned in the expected location.

For now, all this scanner does is log to var/log/virus_scan.log to meet ticket criteria but in the future I expect we’ll probably record the last scan results on the file entity itself, send emails, etc.

This requires that clamd be running on the machine used. I’ll leave it running on griidc-dev. clamd runs as user ‘clamscan’ which has been added to the pelagos group, so it should be able to see all the datafiles once filer does its thing. It can also see the incoming files, which is how I’m currently testing it.


src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,79 @@
...
+     * @param VirusScan $virusScan The VirusScan message to be handled.
+     */
+    public function __invoke(VirusScan $virusScan)
+    {
+        $fileId = $virusScan->getFileId();
+        $file = $this->fileRepository->find($fileId);
+        if ($file instanceof File) {
+            $filePath = $file->getFilePath();
+            try {
+                $socket = (new \Socket\Raw\Factory())->createClient('unix:///run/clamd.scan/clamd.sock');
```
fightingtexasaggie, line 62: This shouldn't be hard-coded as it can change per installation.

src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,79 @@
...
+                $quahog = new \Xenolope\Quahog\Client($socket);
```
praneethpr, line 63: Should we use "use" statements here, instead of fully specifying

src/MessageHandler/VirusScanHandler.php
mickel1138: You can put this in the .env file.

src/Command/ScanFileByUDICommand.php
```diff
@@ -0,0 +1,102 @@
...
+use Symfony\Component\Messenger\MessageBusInterface;
+
+use App\Entity\DatasetSubmission;
+use App\Entity\Dataset;
+use App\Message\VirusScan;
+
+/**
+ * This command triggers a file virus scan, by UDI.
+ */
+class ScanFileByUDICommand extends Command
```
mickel1138, line 19: I think we should remove this command. Manual scans can just be done by other means.

src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,89 @@
...
+     * @param VirusScan $virusScan The VirusScan message to be handled.
+     */
+    public function __invoke(VirusScan $virusScan)
+    {
+        $fileId = $virusScan->getFileId();
+        $file = $this->fileRepository->find($fileId);
+        if ($file instanceof File) {
+            $filePath = $file->getFilePath();
+            try {
+                $socket = (new SocketFactory())->createClient($this->clamdSock);
```
mickel1138, line 72: I would like to see this moved to a utility, so we can call the scanning utility at other times (synchronously).

src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,89 @@
...
+                $quahog = new QuahogClient($socket);
+                $result = $quahog->scanFile($filePath);
```
mickel1138, line 74: Could we use scanStream here instead. That way we don't have to deal with file permissions.

src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,89 @@
...
+                $this->logger->info(sprintf('Virus scanned file id:%s filename: %s. Status: %s. Reason: %s.', $fileId, $result['filename'], $result['status'], $result['reason']));
+            } catch (\Exception $e) {
+                $this->logger->error(sprintf('Unable to scan file. Message: %s', $e->getMessage()));
+                // On catch, furter execution prevented, so attributes will not be updated.
+                return;
+            }
+            // In the future, we'll likely set an attribute of the scan results, or an email for identified viruses, etc.
+            //$file->setVirusLastScanDate();
+            //$file->setVirusLastScanResult($result['status']);
+            //$this->entityManager->flush();
```
mickel1138, line 84: Don't think we need this.

src/Message/VirusScan.php
```diff
@@ -0,0 +1,36 @@
...
+    protected $fileId;
+
+    /**
+     * Constructor.
+     *
+     * @param int $fileId The file ID of the file to be scanned.
+     */
+    public function __construct(int $fileId)
+    {
+        $this->fileId = $fileId;
```
mickel1138, line 24: I think we should change this to filepath as a constructor, this way it can be independent if a file entity exists.

src/Command/ScanFileByUDICommand.php
fightingtexasaggie: If we remove it, we need a way to get a listing of file paths and names to all files associated with an UDI. Mass-scanning the storage directory would also require a means to back-resolve a file to an UDI.

src/MessageHandler/VirusScanHandler.php
fightingtexasaggie: removing

src/Command/ScanFileByUDICommand.php
mickel1138: There a way to look it up I guess. I just don't see this command ever being used. Scan once in ingress, maybe again on egress. If period scans and required, we can deal with it then.

fightingtexasaggie:
To test (on griidc-dev):
Load fresh data dump.
Replace datafile for an UDI's dataset for these test cases, I used E2.x842.000:0001 in this example, which has file id 9191 in the dump.

Test case 1: Normal clean file.
pelagos_mwilliamson=> update file set file_path = '/home/users/mwilliamson/1GB-of-random-uncompressible-data.dat' where id = 9291;
run test command (will be removed later...handy for now)
bin/console pelagos:virusscan-file-by-udi E2.x842.000:0001
Verify var/log/dev.log virus_scan.INFO line shows OK.

Test case 2: Oversize file
pelagos_mwilliamson=> update file set file_path = '/home/users/mwilliamson/2GB-of-random-uncompressible-data.dat' where id = 9291;
bin/console pelagos:virusscan-file-by-udi E2.x842.000:0001
Verify var/log/dev.log virus_scan.WARNING says skipped because oversize. (command also warns)

Test case 3: missing file
pelagos_mwilliamson=> update file set file_path = '/home/users/mwilliamson/missing.dat' where id = 9291;
run command and verify log shows virus_scan.WARNING Unable to scan missing file

Test case 4: virus
pelagos_mwilliamson=> update file set file_path = '/home/users/mwilliamson/eicar-testfile.txt' where id = 9291;
run command and verify log shows virus_scan.WARNING Virus found message.

Test case 5: clamd not running (and not oversize or missing)
sudo killall clamd
run command with file from step 4 and verify log shows virus_scan.WARNING "Unable to scan fille. Message: Socket operation failed..."
The quohog library connects to the clamav engine via a socket, which is only present when the clamd process is running.

Re-enable clamd:
sudo clamd (uses configuration from /etc/clamd.d/scan.conf, configured to allow up to 1GB files to be scanned)

src/Command/ScanFileByUDICommand.php
fightingtexasaggie: I'd like to leave this in for ease of testing until we hook it into something else. It's otherwise hard to actually test a run of it.

src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Invoke function to scan a file for viruses.
+     *
+     * @param VirusScan $virusScan The VirusScan message to be handled.
+     */
+    public function __invoke(VirusScan $virusScan)
+    {
+        $fileId = $virusScan->getFileId();
+        $file = $this->fileRepository->find($fileId);
```
mickel1138, line 70: Once again, I think we should use use file path as a property in VirusScan, way less stuff to inject and things to look up, also less code. Plus extra bonus, we can use this scanner for anything that has a file path. Also, didn't we discuss that this scanner would be stream based? In that case the property would be a stream, and the virusscanner doesn't have to care about permissions.

src/Message/VirusScan.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace App\Message;
+
+/**
+ * Message for virus-scanning files.
+ */
+class VirusScan
```
mickel1138, line 8: So the message are of type "command" and should be names a command. Or else it will be very confusing. Something like: 'ScanFileForVirus' or 'doVirusScanOnFile' or 'VirusScanFile' even maybe?

src/MessageHandler/VirusScanHandler.php
fightingtexasaggie: The utility is stream based, the message handler is not. Something has to translate a file to a stream.

fightingtexasaggie: I can certainly load filepath into the message though.

fightingtexasaggie:
There are tuning issuse still that cause clamd to sometimes fail on the 1GB test files. For now, we can test up to 100MB reliably.  We can tune it accordingly later. (timeouts likely)

src/Message/ScanFileForVirus.php
```diff
@@ -0,0 +1,74 @@
...
+    protected $id;
+
+    /**
+     * Constructor.
+     *
+     * @param array    The stream and attributes describing the associated file for who's stream is to be scanned.
+     * @param string   The UDI associted with the stream.
+     * @param int      The fileId associated with the stream.
+     */
+    public function __construct(array $stream, string $udi, int $id)
```
mickel1138, line 38: Add :void for the PSR12 gods...

src/Message/ScanFileForVirus.php
```diff
@@ -0,0 +1,74 @@
...
+ * Message for virus-scanning files.
+ */
+class ScanFileForVirus
+{
+    /**
+     * The stream to be scanned.
+     *
+     * @var resource
+     */
+    protected $stream;
```
mickel1138, line 15: These can be private, you're not sharing this class with anyone like an entity.

fightingtexasaggie:
To test, just submit some files. I've removed the Pelagos command and instead tied it into the dataset submission handler. Submit an eicar file to simulate a real virus, if your desktop will allow. Verify logfile output in var/log/virus_scan.log file.

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -166,17 +187,25 @@ private function processFile(File $file): void
...
 
         // File virus Scan
-        // TODO implement method
-        /*
-        $scanFile = new VirusScan($fileId);
-        $this->messageBus->dispatch($scanFile);
-        */
+        if ($file instanceof File) {
+            $filename = $this->datastoreDirectory . '/' . $file->getFilePath();
+            $id = $file->getId();
```
fightingtexasaggie, line 197: not used, remove

src/Message/ScanFileForVirus.php
fightingtexasaggie: Constructors, destructors, and the clone operator cannot declare return types though.
https://wiki.php.net/rfc/return_types

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -166,17 +187,24 @@ private function processFile(File $file): void
...
         $this->messageBus->dispatch($hashFile);
 
         // File virus Scan
-        // TODO implement method
-        /*
-        $scanFile = new VirusScan($fileId);
-        $this->messageBus->dispatch($scanFile);
-        */
+        if ($file instanceof File) {
+            $filename = $this->datastoreDirectory . '/' . $file->getFilePath();
```
praneethpr, line 196: We don't need to do this, we can inject Datastore util class, and get the FileStream. Use getFile() to get the stream

src/Message/ScanFileForVirus.php
```diff
@@ -0,0 +1,74 @@
...
+     * The ID of the file assocated with the stream, for identification purposes.
+     *
+     * @var int id
+     */
+    private $id;
+
+    /**
+     * Constructor.
+     *
+     * @param array    The stream and attributes describing the associated file for who's stream is to be scanned.
```
praneethpr, line 34: variable names are missing

src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @param VirusScan $virusScan The VirusScan message to be handled.
+     */
+    public function __invoke(ScanFileForVirus $scanFileForVirusMessage)
+    {
+        $udi = $scanFileForVirusMessage->getUdi();
+        $id = $scanFileForVirusMessage->getId();
+        $streamArray = $scanFileForVirusMessage->getStream();
+        $result = $this->scanner->scanResourceStream($streamArray['fileStream']);
+        if ($result['status'] == 'FOUND') {
```
praneethpr, line 51: FOUND can be changed to a constant, and can use '==='

src/MessageHandler/VirusScanHandler.php
```diff
@@ -0,0 +1,63 @@
...
+            $this->logger->alert(sprintf('Virus found in file ID: %s for UDI: %s, VIRUS ID: %s.', $id, $udi, $result['reason']));
+        } elseif ($result['status'] !== 'failed') {
+            $this->logger->info(sprintf('Virus scanned file ID: %s for UDI: %s. Status: %s.', $id, $udi, $result['status']));
+        } else {
+            if ($result['reason'] == 'oversize') {
```
praneethpr, line 56: Missing type comparison, and can use constants

src/Util/VirusScanUtil.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->clamdSock = $clamdSock;
+    }
+
+    /**
+     * Scan a filestream for viruses.
+     *
+     * @param mixed $fileHandle A filesystem resource to scan.
+     * @return array
+     */
+    public function scanResourceStream($fileHandle)
```
praneethpr, line 33: return type missing, and use array for the method parameter

src/Util/VirusScanUtil.php
```diff
@@ -0,0 +1,55 @@
...
+    public function __construct(string $clamdSock)
+    {
+        $this->clamdSock = $clamdSock;
+    }
+
+    /**
+     * Scan a filestream for viruses.
+     *
+     * @param mixed $fileHandle A filesystem resource to scan.
+     * @return array
```
praneethpr, line 31: Missing blank line

src/Util/VirusScanUtil.php
```diff
@@ -0,0 +1,55 @@
...
+                    $socket = (new SocketFactory())->createClient($this->clamdSock);
+                    $quahog = new QuahogClient($socket);
+                    $result = $quahog->scanResourceStream($fileHandle, 1024000);
+                } catch (\Exception $e) {
+                    $result['status'] = 'failed';
+                    $result['reason'] = $e->getMessage();
+                }
+            }
+        } else {
+            throw new \Exception('stream not a resource');
```
praneethpr, line 51: Missing throws tag in the annotation

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -70,30 +77,43 @@ class DatasetSubmissionFilerHandler implements MessageHandlerInterface
...
+     * @var string
+     */
+    private $datastoreDirectory;
+
     /**
      * DatasetSubmissionFilerHandler constructor.
      *
      * @param DatasetSubmissionRepository $datasetSubmissionRepository Dataset Submission Repository.
      * @param LoggerInterface             $filerLogger                 Name hinted filer logger.
+     * @param LoggerInterface             $virusScanLogger             Name hinted virus-scan logger.
```
mickel1138, line 92: This is is submissions handler, virus scan logger should go to it's own handler.

mickel1138 (approved):

PR #757, opened 2020-09-25
Added dataset submission message and message handler

praneethpr:


src/Message/DatasetSubmissionFiler.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $this->datasetSubmissionId = $datasetSubmissionId;
+    }
+
+    /**
+     * The dataset submission ID getter.
+     *
+     * @return integer The dataset submission ID.
+     */
+    public function getSubmissionId(): int
```
mickel1138, line 32: Even tho it might look silly, I would put the entire name here. getDatasetSubmissionId.

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -0,0 +1,157 @@
...
+            foreach ($fileset->getFiles() as $file) {
+                $this->processFile($file, $loggingContext);
+            }
+
+            $datasetSubmission->setDatasetFileTransferStatus(
+                DatasetSubmission::TRANSFER_STATUS_COMPLETED
+            );
+            $dataset->updateAvailabilityStatus();
+            $this->entityManager->persist($datasetSubmission);
+            $this->entityManager->persist($dataset);
```
mickel1138, line 113: Do these really have to be persisted? Doesn't that mean they have recreated?

praneethpr: Yes, because we update the availability status of the dataset

src/Message/DatasetSubmissionFiler.php
praneethpr: Changed it

src/MessageHandler/DatasetSubmissionFilerHandler.php
mickel1138: I think you only persist when you want to start tracking an object, this object is already being tracked. A flush should do. I'm have a bit of a hard time testing this.

praneethpr: I'm not sure how it is already being tracked?

mickel1138 (approved):
Looks good. Tested as much as I could.

PR #758, opened 2020-09-29
Feature/pelagos 4251 filter dataset for sitemap

mickel1138:


src/Controller/DefaultController.php
```diff
@@ -136,24 +136,32 @@ public function admin()
...
-                )
-            );
+        if ($fundingOrgFilter->isActive()) {
+            $criteria = array_merge($criteria, array(
+                'researchGroup' =>
+                $fundingOrgFilter->getResearchGroupsIdArray()
+            ));
+        }
+
+        $datasets = $this->getDoctrine()->getRepository(Dataset::class)->findBy($criteria);
```
praneethpr, line 162: I don't which is the best practice, autowiring DatasetRepository here or using it like this

src/Controller/DefaultController.php
```diff
@@ -136,24 +136,32 @@ public function admin()
...
 
+        $response = new StreamedResponse(function () use (&$datasets) {
```
praneethpr, line 164: Do we need to pass it by reference here?

mickel1138: No you don't have to, but otherwise a copy will be made. Not that is matters really.

src/Controller/DefaultController.php
mickel1138: Autowiring is proably better. I'll look into it.

src/Controller/DefaultController.php
praneethpr: I think objects are passed by references by default, I don't think we need the "&" here

praneethpr (approved):

PR #760, opened 2020-10-01
Feature/pelagos 4252 create data store utility class

praneethpr:


src/MessageHandler/DeleteFileHandler.php
```diff
@@ -25,31 +26,38 @@ class DeleteFileHandler implements MessageHandlerInterface
...
         $this->logger = $deleteFileLogger;
         $this->fileRepository = $fileRepository;
+        $this->datastore = $datastore;
     }
 
     public function __invoke(DeleteFile $deleteFile)
     {
-        $fileId = $deleteFile->getFileId();
+        $fileId = $deleteFile->getFilePath();
         $this->logger->info(sprintf('Processing File with ID: %d', $fileId));
```
mickel1138, line 53: This won't work now, because %d is decimal, and the path is string.

src/Util/Datastore.php
```diff
@@ -0,0 +1,92 @@
...
+     * Deletes a file from the disk.
+     *
+     * @param string $filePath
+     *
+     * @return void
+     */
+    public function deleteFile(string $filePath): void
+    {
+        $this->datastoreFlysystem->delete($filePath);
+
```
mickel1138, line 81: extra line

mickel1138:
I'm missing a write stream function?

src/MessageHandler/DeleteFileHandler.php
praneethpr: Fixed it

src/Util/Datastore.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Retrieves all files in a dataset.
+     */
+    public function getAllFiles()
```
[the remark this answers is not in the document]
praneethpr: Done

praneethpr:
> I'm missing a write stream function?

Why would we need a write stream function?

praneethpr:
> > I'm missing a write stream function?
> 
> Why would we need a write stream function? I'm using a copy function instead

Looked into the v2 docs, copy is not going to be supported. Changed it to writeStream

src/Util/Datastore.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Deletes a file from the disk.
+     *
+     * @param string $filePath
+     *
+     * @return void
+     */
+    public function deleteFile(string $filePath): void
```
mickel1138, line 83: This should probably return success?

src/MessageHandler/DeleteFileHandler.php
```diff
@@ -19,39 +20,33 @@ class DeleteFileHandler implements MessageHandlerInterface
...
-                // TODO implement method
-                //$this->fileManager->deleteFile($filePath);
-            } catch (\Exception $e) {
-                $this->logger->error(sprintf('Unable to hash file. Message: %s', $e->getMessage()));
-                return;
-            }
+        $filePath = $deleteFile->getFilePath();
+        $this->logger->info(sprintf('Processing File with ID: "%s"', $filePath));
+        try {
+            $this->datastore->deleteFile($filePath);
```
mickel1138, line 46: This is returning a void.

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -102,7 +114,11 @@ public function __invoke(DatasetSubmissionFiler $datasetSubmissionFiler)
             // Log processing complete.
             $this->logger->info('Dataset submission process started', $loggingContext);
             foreach ($fileset->getFiles() as $file) {
-                $this->processFile($file, $loggingContext);
```
mickel1138, line 105: I don't know if we need to shove this logging context around. It'll be the same information every log event. The only thing that's new in `processFile` it an generate it's own logging context there, it just need the file ID, the other information would we seen a few lines over it? Opinions? Do we need all the context for every log line?

praneethpr: Maybe its not required, I think it would be safe to assume that those files are under the same submission. We can remove passing the logging context into the processFile method

src/Util/Datastore.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Setting the logger interface.
+     *
+     * @param LoggerInterface $logger Monolog logger interface instance.
+     *
+     * @return void
+     */
+    public function setLogger(LoggerInterface $logger): void
```
mickel1138, line 52: This is not used. Should this be in the contructor?

praneethpr: We discussed this on Friday, we decided we will use the default monologger https://symfonycasts.com/screencast/symfony4-fundamentals/logger-trait

mickel1138: Ah ok. Isn't this the old method? Instead you want to implement loggerInterface or something? Or just add it to the constructor, it should use the default logger.

praneethpr: LoggerTrait was creating another Helper class etc, and the other method of using it in the constructor was the usual way and we thought we will do it like this because we don't have to specify any specific channel and it goes to the default

praneethpr: Moved it to the constructor, but I thought this was the way to inject default channel

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -12,6 +12,8 @@
 use App\Message\VirusScan;
 use App\Repository\DatasetSubmissionRepository;
 
+use App\Repository\FileRepository;
```
mickel1138, line 15: Sorting of use statements.

mickel1138 (approved):
Tested where possible, approved.

PR #764, opened 2020-10-07
Feature/pelagos 4253 replace consumer with new handler s

praneethpr:


src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -153,7 +153,7 @@ private function processFile(File $file): void
         // TODO implement method
 
         try {
-            $newFileDestination = $this->datastore->addFile(fopen($filepath, 'r'));
+            $newFileDestination = $this->datastore->addFile(['fileStream' => fopen($filepath, 'r')]);
```
mickel1138, line 156: Should we turn this into a helper class so it's a little easier?

praneethpr: Even if its a helper class, the object of the helper class should return resource type inside an array

mickel1138 (approved):
This part works.

fightingtexasaggie (approved):
reviewed, tested...seems to work

PR #765, opened 2020-10-08
filter only non deleted files

praneethpr:


src/Entity/Fileset.php
```diff
@@ -31,7 +31,7 @@ public function __construct()
     }
 
     /**
-     * Getter for files.
+     * Getter for all files.
      *
      * @return Collection
      */
```
mickel1138, line 37: Maybe call this function getAllFiles?

tests/unit/Entity/FilesetTest.php
```diff
@@ -41,8 +41,13 @@ class FilesetTest extends TestCase
     protected function setUp()
     {
         $this->fileset = new Fileset;
-        $this->mockFile = \Mockery::mock(File::class, array('setFileset' => null));
-        $this->defaultFile = $this->fileset->getFiles()->first();
+        $this->mockFile = \Mockery::mock(
+            File::class,
+            array(
+                'setFileset' => null,
+                'getStatus' => 'new'
```
mickel1138, line 48: Use constant instead: `File::FILE_NEW`

tests/unit/Entity/FilesetTest.php
```diff
@@ -52,17 +57,17 @@ protected function setUp()
      */
     public function testAddFileToFileset()
```
mickel1138, line 58: Do we want test for all the other functions? If you look at my PR, maybe we can merge some of the functionality?

praneethpr: I added those tests for the other getters

mickel1138 (approved):
Looks good.

PR #766, opened 2020-10-08
Added isDone method to fileset.

mickel1138:


mickel1138:
Needs test...

src/Controller/UI/DatalandController.php
```diff
@@ -90,6 +90,11 @@ public function defaultAction(string $udi)
                 $wkt = null;
             }
         }
+        
+        $fileset = $dataset->getDatasetSubmission()->getFileset();
+        
+        dd($fileset->isDone());
```
fightingtexasaggie, line 96: debug code

mickel1138: That file shouldn't be there...

src/Entity/Fileset.php
```diff
@@ -64,4 +66,26 @@ public function removeFile(File $file)
...
+    public function isDone() :bool
+    {
+        $criteria = Criteria::create()
+        ->where(
+            new Comparison(
+                'status',
+                Comparison::IN,
+                array(
+                    File::FILE_NEW,
+                    File::FILE_IN_PROGRESS
```
praneethpr, line 84: What about error? should we add that too?

mickel1138: I think error means processed. We can add it later if FILE_ERROR has meaning. How does that sound?

praneethpr (approved):

fightingtexasaggie (approved):
looks pretty straightforward using criteria, test ran clean

PR #768, opened 2020-10-13
Added file process listener

praneethpr:


src/Event/FileProcessListener.php
```diff
@@ -0,0 +1,32 @@
...
+     * @return void
+     */
+    public function preUpdate(File $file, LifecycleEventArgs $args): void
+    {
+        $file = $args->getObject();
+        $entityManager = $args->getEntityManager();
+        $unitOfWork = $entityManager->getUnitOfWork();
+        $metadata = $entityManager->getClassMetadata(get_class($file));
+        if ($file instanceof File) {
+            if ($file->getFileSha256Hash()) {
```
mickel1138, line 26: We may also want to check if the status is "In Progress". Also should we check other fields for completeness?

praneethpr: Fixed it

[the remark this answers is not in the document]
praneethpr: Changed it to empty

mickel1138 (approved):
Works well!

PR #769, opened 2020-10-14
Added getFilesize to fileset.

mickel1138:


src/Entity/Fileset.php
```diff
@@ -124,4 +124,21 @@ public function isDone() :bool
...
+
+    /**
+     * Returns the total filesize for this fileset.
+     *
+     * @return integer
+     */
+    public function getFileSize() :int
+    {
+        $fileSize = 0;
+        $filesArray = $this->files->toArray();
```
praneethpr, line 136: I think we can remove this, if we are going to use foreach

fightingtexasaggie: indeed not needed...test passes without it to verify

fightingtexasaggie (approved):

src/Entity/Fileset.php
mickel1138: I was lead to believe that I had to use an array, but ArrayCollection is iterable, I guess Collection is not. Fixed it.

praneethpr (approved):

fightingtexasaggie (approved):
retested, working

PR #770, opened 2020-10-14
Feature/pelagos 4271 redirect getfileset for datasetsubmission

mickel1138:
Review getFilesize from other PR first!


mickel1138:
Also delete the setter calls.

fightingtexasaggie (dismissed):
I did a quick grep in src and templates and the only thing that would break is the ColdStorageFlagCommand, which we'll have to address later anyway. Everything else just calls the getter. The twig extension formatBytes should handle the null fine...think it already can be null.

praneethpr (approved):

PR #771, opened 2020-10-14
Removed transfer status.

mickel1138:


fightingtexasaggie (commented):
While this tests fine and is straightforward, we are currently using datasetSubmissionTransferStatus all over the place, including to indicate remotely hosted. Should we address these changes on  this PR?

grep -ir datasetfiletransferstatus src templates

src/Entity/DatasetSubmission.php
```diff
@@ -1612,29 +1598,17 @@ public function getDatasetFileUri() : ?string
...
-
     /**
      * Get the dataset file transfer status.
      *
      * @return string|null
      */
     public function getDatasetFileTransferStatus() : ?string
     {
-        return $this->datasetFileTransferStatus;
+        if ($this->fileset instanceof Fileset and $this->fileset->isDone()) {
```
praneethpr, line 1608: We need to figure out a way for remotely hosted datasets, because updateAvailabilityStatus() expects transfer status to be TRANSFER_STATUS_REMOTELY_HOSTED

mickel1138: I think the best thing to do is just to make a whole new property called "Remotely Hosted".

src/Entity/DatasetSubmission.php
```diff
@@ -1612,29 +1598,17 @@ public function getDatasetFileUri() : ?string
...
-     * @param string|null $datasetFileTransferStatus The dataset file transfer status.
-     *
-     * @see TRANSFER_STATUSES class constant for valid values.
-     *
-     * @return void
-     */
-    public function setDatasetFileTransferStatus(?string $datasetFileTransferStatus)
-    {
-        $this->datasetFileTransferStatus = $datasetFileTransferStatus;
-        $this->updateAvailabilityStatus();
```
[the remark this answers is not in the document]
mickel1138: We could, but we should have a good think on this, and see if you improve the status flow.

src/Entity/DatasetSubmission.php
praneethpr: Sure, we could do that. But we need to discuss the current dataset workflow, because the availability status is controlled by this property

PR #772, opened 2020-10-14
Added API to create file entity in dataset submission

praneethpr:


src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -392,6 +396,44 @@ public function getFiles(int $id)
...
+     *     requirements={"id"="\d+"}
+     *     )
+     *
+     * @View()
+     *
+     * @return Response
+     */
+    public function addFile(DatasetSubmission $datasetSubmission, Request $request, EntityManagerInterface $entityManager)
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
```
mickel1138, line 420: If file already exists in filelist, mark other one as FILE_DELETE (unless it's still FILE_NEW, iow, it's not FILE_DONE or FILE_...).

fightingtexasaggie (commented):
I currently see  this error upon submitting a dataset. Prior to that, the upload appears to work fine.

Attempted to call an undefined method named "setDatasetFileSize" of class "App\Entity\DatasetSubmission".

praneethpr:
> I currently see this error upon submitting a dataset. Prior to that, the upload appears to work fine.
> 
> Attempted to call an undefined method named "setDatasetFileSize" of class "App\Entity\DatasetSubmission".

I think I need to merge the epic into this, we fixed removing those setters in the other pull request which I merged this morning

fightingtexasaggie (approved):
it worked after you merged in the fresh epic.

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -2,23 +2,27 @@
...
 use Nelmio\ApiDocBundle\Annotation\Operation;
 use Nelmio\ApiDocBundle\Annotation\Model;
+
 use Swagger\Annotations as SWG;
 
-use App\Util\UrlValidation;
-use App\Entity\DatasetSubmission;
-use App\Entity\Fileset;
-use App\Form\DatasetSubmissionType;
+use Symfony\Component\HttpFoundation\Request;
```
mickel1138, line 22: Cool ordering, but in the past we have always put our own App last, and we should be consistent.

praneethpr: I thought it was alphabetical, will move App stuff to the bottom

PR #773, opened 2020-10-16
Removed setter

praneethpr:


fightingtexasaggie (commented):
I guess this command will need to store the "new" stubfile in the new manner.

praneethpr:
> I guess this command will need to store the "new" stubfile in the new manner.

Yes, do we want to put a ticket for it so that we don't forget about it

src/Controller/UI/DatasetReviewController.php
```diff
@@ -598,7 +598,6 @@ protected function newDatasetFile(DatasetSubmission $datasetSubmission)
     {
         $datasetSubmission->setDatasetFileTransferStatus(DatasetSubmission::TRANSFER_STATUS_NONE);
         $datasetSubmission->setDatasetFileName(null);
```
mickel1138, line 600: We can probably get rid of these too soon.

mickel1138:
> > I guess this command will need to store the "new" stubfile in the new manner.
> 
> Yes, do we want to put a ticket for it so that we don't forget about it

Yes, let's put in a ticket to deal with Cold Storage.

mickel1138 (approved):
Approved.

src/Controller/UI/DatasetSubmissionController.php
```diff
@@ -318,7 +318,6 @@ protected function newDatasetFile(DatasetSubmission $datasetSubmission)
     {
         $datasetSubmission->setDatasetFileTransferStatus(DatasetSubmission::TRANSFER_STATUS_NONE);
```
mickel1138, line 319: This is also broken, this function is gone now?

mickel1138 (changes requested):
Set tranfer type breaks on submission.

src/Controller/UI/DatasetSubmissionController.php
praneethpr: No, we din't merge that branch. let me test again

praneethpr:
> Set tranfer type breaks on submission.

I'm unable to reproduce the error which you are talking about, the regular dataset workflow works fine. Dataset submission works for old ones (i.e Request revisions) also

mickel1138 (approved):
Looks good, did a fill rebuild and cleared old files. Works now.

PR #774, opened 2020-10-20
Added zip file utility class

praneethpr:


src/Util/ZipFiles.php
```diff
@@ -0,0 +1,53 @@
...
+     * @param string          $zipFile Filename of the zipfile.
+     *
+     * @throws \Exception When utility class can not open/write to zip file.
+     *
+     * @return string
+     */
+    public function createZipFile(Collection $files, string $zipFile) : string
+    {
+        $flag = (file_exists($zipFile))? ZIPARCHIVE::OVERWRITE : ZIPARCHIVE::CREATE;
+        if ($this->zip->open($zipFile, $flag) === true) {
```
fightingtexasaggie, line 42: I'm concerned there really isn't a way to control where the zipfile itself is created. I'm assuming it would go into the symfony var folder, but it would be nice if we could contain them to a separate filesystem that could be filled without detriment to the rest of the system. If var itself fills, I don't think the site would be unusable.

praneethpr: I think we should do that in the messageHandler class which calls this utility

src/Util/ZipFiles.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace App\Util;
+
+use App\Entity\File;
+
+use Doctrine\Common\Collections\Collection;
```
mickel1138, line 7: I'm not sure it it's better to use ArrayCollection, which have used more often in other code. But when I look at the source code of this library, they seem nearly identical.

src/Util/ZipFiles.php
```diff
@@ -0,0 +1,53 @@
...
+    public function createZipFile(Collection $files, string $zipFile) : string
+    {
+        $flag = (file_exists($zipFile))? ZIPARCHIVE::OVERWRITE : ZIPARCHIVE::CREATE;
+        if ($this->zip->open($zipFile, $flag) === true) {
+            foreach ($files as $file) {
+                $this->zip->addFile($file->getFilePath(), $file->getFileName());
+            }
+            $this->zip->close();
+        } else {
+            throw new \Exception('Unable to open zip file');
```
mickel1138, line 48: Maybe this should say "unable to create"?

mickel1138:
I would really like to see a test on this, I would write a test to see if this works. I can write it if you like?

src/Util/ZipFiles.php
mickel1138: Ok, collection is the "interface", and arrayCollection is what you can instantiate.

tests/unit/Util/ZipFilesTest.php
```diff
@@ -0,0 +1,74 @@
...
+        $file2->setFileName('zz.txt');
+        
+        $fileset->addFile($file1);
+        $fileset->addFile($file2);
+        
+        $zipFiles = new ZipFiles();
+        
+        $zipFiles->createZipFile($fileset->getAllFiles(), '/tmp/testzip.zip');
+        
+        $zip = new \ZipArchive;
```
praneethpr, line 59: I'm not sure unit tests should contain logic, do we want to look at this library for using it for unit testing
https://github.com/bovigo/vfsStream

src/Util/ZipFiles.php
```diff
@@ -0,0 +1,53 @@
...
+     * @param string          $zipFile Filename of the zipfile.
+     *
+     * @throws \Exception When utility class can not open/write to zip file.
+     *
+     * @return void
+     */
+    public function createZipFile(Collection $files, string $zipFile) : void
+    {
+        $options = new Archive();
+        $options->setSendHttpHeaders(true);
```
mickel1138, line 46: This probably not needed. I think is used if to stream this stream out to HTTP.

src/Util/ZipFiles.php
```diff
@@ -0,0 +1,53 @@
...
+     * Creates a zip file from collection of files.
+     *
+     * @param Collection|File $files   Files that need to be zipped.
+     * @param string          $zipFile Filename of the zipfile.
+     *
+     * @throws \Exception When utility class can not open/write to zip file.
+     *
+     * @return void
+     */
+    public function createZipFile(Collection $files, string $zipFile) : void
```
mickel1138, line 43: I don't like giving this function the files collection, it has to know to do deal with files, and not be used for other things.
I think we should refactor this. Create another function called `addFiles` and just extent the `addFilesFromStream` function. You can create a new Archive and zipStream in the construction. Then finally create a `finish` function with the zip finish.

tests/unit/Util/ZipFilesTest.php
```diff
@@ -0,0 +1,74 @@
...
+use Doctrine\Common\Collections\ArrayCollection;
+
+use App\Entity\File;
+use App\Entity\Fileset;
+use App\Util\ZipFiles;
+
+/**
+ * Unit tests for App\Util\ZipFiles.
+ */
+class ZipFilesTest extends TestCase
```
mickel1138, line 16: This test case it not needed I think. It can be removed. Together with the test files.

src/Util/ZipFiles.php
praneethpr: True

src/Util/ZipFiles.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $options = new Archive();
+        $options->setSendHttpHeaders(true);
+        $zip = new ZipStream($zipFile, $options);
+        foreach ($files as $file) {
+           $zip->addFileFromStream($file->getFileName(), $this->datastoreFlysystem->readStream($file->getFilePath));
+        }
+        $zip->finish();
```
mickel1138, line 51: This doesn't return a stream, if we would like to do that check this example:
https://github.com/maennchen/ZipStream-PHP/wiki/FlySystem-example
But we can use `php://temp` instead of `php://memory` (Memory is limited  to memory, temp used a temp disk file).

mickel1138 (approved):
Approved.

PR #775, opened 2020-10-21
Feature/pelagos 4279 create message handler to generate zip file and trigger event

praneethpr:
Zip file utility branch is merged into this one. Please take a look at this and we can further add events to trigger an email or any other approach of how we are going to serve this zip file to the user

### DIFF
--- a/.env
+++ b/.env
@@ -69,6 +69,7 @@ DATA_DOWNLOAD_BROWSER_GROUP=browser_group
 WEB_SERVER_USER=apache
 ANONYMOUS_FTP_USER=user
 ANONYMOUS_FTP_PASS=pass
+DOWNLOAD_PATH="path/to/zip"
 
 ###> INDEX NAMES ###
 DISCOVERY_INDEX=pelagos_instancename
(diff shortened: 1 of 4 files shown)


src/Message/ZipFile.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace App\Message;
+
+/**
+ * The message for Zip file handler.
+ */
+class ZipFile
+{
+
```
mickel1138, line 10: New line.

src/Message/ZipFile.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace App\Message;
+
+/**
+ * The message for Zip file handler.
+ */
+class ZipFile
```
mickel1138, line 8: The message is very specific, the zip the dataset files, maybe we should call it that?

.env
```diff
@@ -69,6 +69,7 @@ DATA_DOWNLOAD_BROWSER_GROUP=browser_group
 WEB_SERVER_USER=apache
 ANONYMOUS_FTP_USER=user
 ANONYMOUS_FTP_PASS=pass
+DOWNLOAD_PATH="path/to/zip"
```
mickel1138, line 72: Maybe this should be another Flysystem, or the datastore? Discuss...

PR #776, opened 2020-10-28
Misc/library updates object path

mickel1138:
Did composer update, and yarn update. Please test.


fightingtexasaggie (approved):
Tested and all seemed to work, except download, but I'm confident that was an unrelated permissions problem in my dev env.

praneethpr (approved):

PR #777, opened 2020-10-28
Update eslint config to support ECMA6

praneethpr:
We don't have to add it into travis, but I thought we can update the eslint config file
To run,
For example: 
`> eslint assets/js/`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,8 @@
 {
     "env": {
-        "browser": true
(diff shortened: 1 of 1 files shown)


fightingtexasaggie (approved):
I think this is a good idea.

mickel1138:
It's mostly double quote issues, but I like this improvement.

mickel1138 (approved):

PR #778, opened 2020-10-28
Feature/pelagos 4280 add zip file for accepted datasets

praneethpr:


fightingtexasaggie (approved):

src/Message/ZipFile.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace App\Message;
+
+/**
+ * The message for Zip file handler.
+ */
+class ZipFile
```
mickel1138, line 8: I think we should change to name to reflect what it's doing. According to the Symfony message methodology, in the command message bus there is one message (command) and one handler. And this message doesn't just zip all files, or any file, it's very specific. it zips all dataset files. Maybe call it: zipAllDatasetFiles? We might get a generic zipFile message some day.

src/MessageHandler/ZipFileHandler.php
```diff
@@ -0,0 +1,121 @@
...
+
+use App\Repository\FileRepository;
+
+use App\Util\Datastore;
+use App\Util\ZipFiles;
+
+/**
+ * Handler for zipping files for download.
+ */
+class ZipFileHandler implements MessageHandlerInterface
```
mickel1138, line 21: This handler should reflect the changed message name. (from comment above,)

src/MessageHandler/ZipFileHandler.php
```diff
@@ -0,0 +1,121 @@
...
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+use App\Entity\DatasetSubmission;
+
+use App\Message\ZipFile;
+
+use App\Repository\FileRepository;
+
+use App\Util\Datastore;
+use App\Util\ZipFiles;
```
mickel1138, line 16: Maybe start getting this PSR-12 ready?

src/Message/ZipFile.php
praneethpr: Yes, I think so too and it is confusing as the Util class name is also similar

src/MessageHandler/ZipFileHandler.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Invoke function to zip files for download.
+     *
+     * @param ZipFile $zipFile
+     */
+    public function __invoke(ZipFile $zipFile)
+    {
+        $this->logger->info(sprintf('Zipping files started'));
```
mickel1138: Don't really need sprintf if you are not passing in stuff. Also I learned about a new magic syntax in logger:
`$context['something'] = "bla";
`$this->logger->info('Hello {something} , how are you?', $context);`

praneethpr: Copy pasted, Yeah will remove it 👍

src/Repository/FileRepository.php
```diff
@@ -25,4 +27,23 @@ public function __construct(ManagerRegistry $registry)
...
     }
+
+    /**
+     * Query for getting filename and fileapth.
+     *
+     * @param array $fileIds List of fileIds to be queried on.
+     *
+     * @return array
+     */
+    public function getFileNameAndPath(array $fileIds) : array
```
mickel1138, line 38: We don't really need this function? The final result and not needed for something like before. It is just as fast to go trough the about, and then use the functions in the entity instead.

praneethpr: I think we still need it, because we give the list of File ids to the message handler. We need to retrieve the file entities, we can just use findBy(criteria), then we need to filter out to get filename and filepath

src/Util/ZipFiles.php
```diff
@@ -2,48 +2,53 @@
...
-    public function createZipFile(array $fileInfo, string $zipFile) : void
+    public function addFile(string $fileName, array $fileStream): void
     {
-        $zip = new ZipStream($zipFile);
-        foreach ($fileInfo as $file) {
-            $fileStream = $this->datastore->getFile($file['filePath']);
-            if ($fileStream and is_resource($fileStream['fileStream'])) {
-                $zip->addFileFromStream($file['fileName'], $fileStream['fileStream']);
-            }
+        if (!empty($fileStream) and is_resource($fileStream['fileStream'])) {
```
mickel1138, line 40: Maybe this is overkill, but we could write a trait that can check this as a function. In addition it would also check if the fileStream key even existed. What do you think?

src/Repository/FileRepository.php
mickel1138: You're right, we need it. It's not always all files in a dataset.

mickel1138 (approved):
Seen it working, looks good.

PR #779, opened 2020-11-02
Added supervisor config and messenger yaml for prod.

mickel1138:
More details on this can be found here: https://symfony.com/doc/current/messenger.html#supervisor-configuration


bin/deploy
```diff
@@ -20,6 +20,7 @@ esac
 
 # Stop RabbitMQ consumers
 bin/console rabbitmq-supervisor:control stop --wait-for-supervisord
+bin/console messenger:stop-workers
```
mickel1138, line 23: This is actually wrong, this will stop the workers, but then supervisor will restart them right away, but at least the workers will be refreshed if there are code changes.

Do we even have to stop the workers ever at all during deployment? Discuss...

praneethpr: This is what they say on the Symfony messenger documentation

`**Restart Workers on Deploy**

Each time you deploy, you’ll need to restart all your worker processes so that they see the newly deployed code. To do this, run messenger:stop-workers on deploy. This will signal to each worker that it should finish the message it’s currently handling and shut down gracefully. Then, Supervisor will create new worker processes. The command uses the app cache internally - so make sure this is configured to use an adapter you like.`

mickel1138:
Install: 
(Optionally copy .ini and .conf from config/supervisor to var/supervisor, and rename indentifier)
`supervisord --configuration=/opt/pelagos/config/supervisor/supervisord.conf --identifier=pelagos`

To use: 
`supervisorctl --serverurl unix:///opt/pelagos/var/supervisor/supervisor.sock reload`
`supervisorctl --serverurl unix:///opt/pelagos/var/supervisor/supervisor.sock update`
`supervisorctl --serverurl unix:///opt/pelagos/var/supervisor/supervisor.sock start messenger-consume:*`

To deploy and GRP, copy the ini and conf file to var/supervisor and change paths above.

fightingtexasaggie (approved):
went over with mvde, tested

mickel1138:
Also had to update composer to composer 2 (from 1), and while doing that also updated yarn/node packages.

PR #780, opened 2020-11-02
Misc/psr12 compliance

fightingtexasaggie:


src/Twig/Extensions.php
```diff
@@ -252,7 +248,7 @@ function ($personAssociation) use ($roleName) {
      */
     protected static function parseString(Environment $environment, array $context, string $string)
     {
-        $environment->setLoader(new \Twig\Loader\ArrayLoader);
+        $environment->setLoader(new \Twig\Loader\ArrayLoader());
```
fightingtexasaggie, line 251: I know twig is weird. I need to test if this works.

PR #782, opened 2020-11-04
Removed use of various transfer states, also unused force re-download.

mickel1138:
Removed older filer functions.
Simplified "filer" status checks.


praneethpr (approved):

fightingtexasaggie (approved):
seems good to go

PR #784, opened 2020-11-10
Composer upgrade

mickel1138:


fightingtexasaggie (approved):
dif renders a little different, but I'm 99% suspecting it's my test environment. The same happened on 5.50.0, so I think it's unrelated.

fightingtexasaggie (approved):
the fact travis runs now is about as good a test as I can think of.

praneethpr (approved):

PR #786, opened 2020-11-12
Added back dataset filesize.

mickel1138:


praneethpr (approved):
Tested and works, stats also works

src/Entity/Fileset.php
```diff
@@ -157,7 +157,7 @@ public function getFileSize() :int
     {
         $fileSize = 0;
 
-        foreach ($this->files as $file) {
+        foreach ($this->getProcessedFiles() as $file) {
```
fightingtexasaggie, line 160: should we fix this in a different ticket? I guess since this is an epic, no biggie either way...

mickel1138: It's something we forgot to fix before. I think what we want from this is the file size of all the actual files (processed) in a dataset, no the deleted or pending ones.

fightingtexasaggie (approved):
tested, works (after migration)

PR #787, opened 2020-11-12
Removed usages of transfer status and method to figure out if it is m…

praneethpr:
…arked as remotely hosted


src/Entity/DatasetSubmission.php
```diff
@@ -2694,4 +2682,19 @@ public function setRemotelyHostedUrl(?string $remotelyHostedUrl) : void
     {
         $this->remotelyHostedUrl = $remotelyHostedUrl;
     }
+
+    /**
+     * Check if dataset submission is marked as remotely hosted.
+     *
+     * @return bool
+     */
+    public function isMarkedRemotelyHosted(): bool
```
mickel1138, line 2691: "Marked" is a bit ambigious. It could have also said just `isRemotelyHosted' but this seem OK. Opinion?

src/Form/DatasetSubmissionType.php
```diff
@@ -69,37 +69,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'expanded' => true,
                 'multiple' => false,
             ))
-            ->add('datasetFileUri', Type\HiddenType::class, array(
```
mickel1138, line 72: This is being used in `templates/DatasetSubmission/file-submit.html.twig`

src/Entity/DatasetSubmission.php
praneethpr: Dataset entity already has the same method name, I'm ok with **isRemotelyHosted**

src/Form/DatasetSubmissionType.php
praneethpr: I need to do a grep on templates and javascript files

src/Controller/UI/RemotelyHostedDatasetsController.php
```diff
@@ -210,7 +209,7 @@ public function remotelyHostedPrerequsiteCheck(DatasetSubmission $datasetSubmiss
                 }
             }
         }
-        if (!in_array($datasetSubmission->getDatasetFileUri(), $distributionUrls)) {
+        if (!in_array($datasetSubmission->getRemotelyHostedUrl(), $distributionUrls)) {
             $errors[] = 'Dataset File URI does not match distribution URL.';
```
mickel1138, line 213: We have to clean up these errors one day.

src/Entity/DatasetSubmission.php
```diff
@@ -105,25 +105,13 @@ class DatasetSubmission extends Entity
...
     /**
      * Valid values for $datasetFileTransferStatus.
      */
     const TRANSFER_STATUSES = array(
         self::TRANSFER_STATUS_NONE => 'Not Yet Transferred',
         self::TRANSFER_STATUS_COMPLETED => 'Transfer Complete',
-        self::TRANSFER_STATUS_ERROR => 'Transfer Error',
-        self::TRANSFER_STATUS_NEEDS_REVIEW => 'URL Needs Review',
-        self::TRANSFER_STATUS_REMOTELY_HOSTED => 'Remotely Hosted',
+        self::TRANSFER_STATUS_ERROR => 'Transfer Error'
```
mickel1138, line 114: I would delete these just now, they may still be in our database.

src/Entity/DatasetSubmission.php
mickel1138: Was it changed?

mickel1138 (approved):

PR #790, opened 2020-11-18
Feature/pelagos 4290 add udi folder to storage path in datastore

praneethpr:
I've merged feature/PELAGOS-4289-send-full-path-in-dropzone-js-to-api into this branch, we need to merge the other branch first

### DIFF
--- a/src/Message/HashFile.php
+++ b/src/Message/HashFile.php
@@ -10,7 +10,7 @@ class HashFile
     /**
      * The file ID of the to be hashed file.
      *
-     * @var string
+     * @var integer
      */
     protected $fileId;
 
--- a/src/MessageHandler/DatasetSubmissionFilerHandler.php
+++ b/src/MessageHandler/DatasetSubmissionFilerHandler.php
@@ -106,9 +106,10 @@ class DatasetSubmissionFilerHandler implements MessageHandlerInterface
         $datasetSubmissionId = $datasetSubmissionFiler->getDatasetSubmissionId();
         $datasetSubmission = $this->datasetSubmissionRepository->find($datasetSubmissionId);
         $dataset = $datasetSubmission->getDataset();
+        $udi = $datasetSubmission->getDataset()->getUdi();
         $loggingContext = array(
             'dataset_id' => $dataset->getId(),
-            'udi' => $datasetSubmission->getDataset()->getUdi(),
+            'udi' => $udi,
             'dataset_submission_id' => $datasetSubmissionId
         );
         $fileset = $datasetSubmission->getFileset();
@@ -118,7 +119,7 @@ class DatasetSubmissionFilerHandler implements MessageHandlerInterface
             $this->logger->info('Dataset submission process started', $loggingContext);
             foreach ($fileset->getNewFiles() as $file) {
                 if ($file instanceof File) {
-                    $this->processFile($file, $loggingContext);
+                    $this->processFile($file, $loggingContext, $udi);
                     $fileIds[] = $file->getId();
                 } else {
                     $this->logger->alert('File object does not exist');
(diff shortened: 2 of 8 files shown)


src/Util/Datastore.php
```diff
@@ -67,14 +67,16 @@ public function getFile(string $filePath): array
...
      * Moves an uploaded file to datastore disk location.
      *
-     * @param array $fileStream File stream resource object.
+     * @param array  $fileStream      File stream resource object.
+     * @param string $destinationPath File destination path on datastore.
      *
      * @return string
      */
-    public function addFile(array $fileStream): string
+    public function addFile(array $fileStream, string $destinationPath): string
```
mickel1138, line 75: I would just call this filePathName. This should be only the file path name of the file, and not the dictated file location/destination of the file. Of course destinationPath was already used in the code below. I'm just making this comment because it could be confusing.

src/Util/Datastore.php
```diff
@@ -67,14 +67,16 @@ public function getFile(string $filePath): array
...
     {
-        $uuid = Uuid::uuid4();
-        $destinationPath = self::FILES_DIRECTORY . DIRECTORY_SEPARATOR . $uuid->toString();
+        $uuid = Uuid::uuid4()->toString();
+        // add only last 5 bytes of uuid to the destination path
+        $destinationPath = $destinationPath . '_' . substr($uuid, -5);
```
mickel1138, line 79:
```suggestion
        $destinationPath .= '_' . substr($uuid, -5);
```
Here you could use the "Concat Assignment Operator"

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -165,7 +169,10 @@ private function processFile(File $file, array $loggingContext): void
         $file->setStatus(File::FILE_IN_PROGRESS);
 
         try {
-            $newFileDestination = $this->datastore->addFile(['fileStream' => fopen($filepath, 'r')]);
+            $newFileDestination = $this->datastore->addFile(
+                ['fileStream' => fopen($filepath, 'r')],
+                str_replace(':', '.', $loggingContext['udi']) . DIRECTORY_SEPARATOR . $file->getFileName()
```
mickel1138, line 174: The UDI retrieved from the logging context now has significantly more weight than when it was just a logging variable.
Maybe here we should pass in the UDI in the function, and get the function by another means. What we decide to change the logging context, that would break this. So the UDI is an essential piece of information now.

praneethpr: Sure, we could do that

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -153,12 +154,13 @@ public function __invoke(DatasetSubmissionFiler $datasetSubmissionFiler)
...
-     * @param File  $file           The file that is being processed.
-     * @param array $loggingContext The logging context for the related dataset submission.
+     * @param File   $file           The file that is being processed.
+     * @param array  $loggingContext The logging context for the related dataset submission.
+     * @param string $udi            The UDI of the dataset.
      *
      * @return void
      */
-    private function processFile(File $file, array $loggingContext): void
+    private function processFile(File $file, array $loggingContext, string $udi): void
```
mickel1138, line 163: UDI should probably be second. I'm thinking most significant parameters first. Logging context could even be empty.

src/Message/HashFile.php
```diff
@@ -10,7 +10,7 @@ class HashFile
     /**
      * The file ID of the to be hashed file.
      *
-     * @var string
+     * @var integer
```
fightingtexasaggie, line 13: I've always left it a string unless doing math with it. That was a very old rule though, so maybe time to discuss again.

praneethpr: I think its better we use integers for IDs, unless it is like UDI or DOI

fightingtexasaggie (approved):
reviewed, discussed, tested. The files are written to disk as expected now, and the clamd process is getting the correct names too. The whitespace issue has been solved.

PR #791, opened 2020-11-19
Updated migration script.

mickel1138:
I think this should work to move the dataset files.
It is possible, but haven't found any cases that a submission is not yet submitted, this can be checked with:
```select dataset_submission.id, dataset.udi, dataset.dataset_submission_status from dataset_submission join dataset on dataset_submission.dataset_id = dataset.id where dataset_file_transfer_status is null and dataset_file_transfer_type is not null and dataset.dataset_submission_status = 0;```

### DIFF
--- a/src/Migrations/Version20200625131822.php
+++ b/src/Migrations/Version20200625131822.php
@@ -20,14 +20,17 @@ final class Version20200625131822 extends AbstractMigration
     {
         $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
         
-        $statement = "SELECT id FROM dataset_submission WHERE dataset_file_uri LIKE 'file://%';";
+        $statement = "SELECT dataset_submission.dataset_id AS id, dataset.udi AS udi FROM dataset_submission JOIN dataset ON dataset.dataset_submission_id = dataset_submission.id WHERE dataset.dataset_status NOT IN ('None');";
         $datasetSubmissions = $this->connection->fetchAll($statement);
                 
         foreach ($datasetSubmissions as $submission) {
             $id = $submission['id'];
-            $this->addSql("INSERT INTO fileset (id, creator_id, modifier_id, creation_time_stamp, modification_time_stamp) VALUES (nextval('public.fileset_id_seq'), 0, 0, CURRENT_TIMESTAMP(0), CURRENT_TIMESTAMP(0))");
-            $this->addSql("UPDATE dataset_submission SET fileset_id = currval('public.fileset_id_seq') WHERE id = $id");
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Migrations/Version20200625131822.php [hunk 1]
```diff
@@ -20,14 +20,17 @@ public function up(Schema $schema) : void
...
         $datasetSubmissions = $this->connection->fetchAll($statement);
                 
         foreach ($datasetSubmissions as $submission) {
             $id = $submission['id'];
-            $this->addSql("INSERT INTO fileset (id, creator_id, modifier_id, creation_time_stamp, modification_time_stamp) VALUES (nextval('public.fileset_id_seq'), 0, 0, CURRENT_TIMESTAMP(0), CURRENT_TIMESTAMP(0))");
-            $this->addSql("UPDATE dataset_submission SET fileset_id = currval('public.fileset_id_seq') WHERE id = $id");
-            $this->addSql("INSERT INTO file (id, fileset_id, creation_time_stamp, creator_id, modification_time_stamp, modifier_id, file_name, file_size, file_sha256_hash, file_path, status) SELECT nextval('file_id_seq'), currval('public.fileset_id_seq'), CURRENT_TIMESTAMP(0), 0, CURRENT_TIMESTAMP(0), 0, dataset_file_name, dataset_file_size, dataset_file_sha256_hash, SUBSTRING(dataset_file_uri FROM '\/[^\/].*$'), 'done' FROM dataset_submission WHERE id = $id");
+            $udi = $submission['udi'];
+            if (file_exists("/san/data/store/$udi/$udi.dat")) {
```
mickel1138, line 29: This path is based on production.

fightingtexasaggie (approved):
tested and worked on the 800 or so small datasets I had on my dev system.

praneethpr (approved):

src/Migrations/Version20200625131822.php
```diff
@@ -20,14 +20,17 @@ public function up(Schema $schema) : void
...
+                $this->addSql("INSERT INTO fileset (id, creator_id, modifier_id, creation_time_stamp, modification_time_stamp) VALUES (nextval('public.fileset_id_seq'), 0, 0, CURRENT_TIMESTAMP(0), CURRENT_TIMESTAMP(0))");
+                $this->addSql("UPDATE dataset_submission SET fileset_id = currval('public.fileset_id_seq') WHERE id = $id");
+                $this->addSql("INSERT INTO file (id, fileset_id, creation_time_stamp, creator_id, modification_time_stamp, modifier_id, file_name, file_size, file_sha256_hash, file_path, status) SELECT nextval('file_id_seq'), currval('public.fileset_id_seq'), CURRENT_TIMESTAMP(0), 0, CURRENT_TIMESTAMP(0), 0, dataset_file_name, dataset_file_size, dataset_file_sha256_hash, '$udi/$udi.dat', 'done' FROM dataset_submission WHERE id = $id");
```
praneethpr, line 32: Can we also add leading slash for the filenames? For example, as of now the filename field : "RT spectra.zip". It would be helpful to get the directory, if it was "/RT spectra.zip". I could do it in my ticket as well, but thought it would be easier to add in this

praneethpr: Maybe its not required, I added workaround code in my branch

PR #792, opened 2020-11-19
Replaced object file provider with custom file provider

praneethpr:


fightingtexasaggie (approved):
tested and confirmed working. I'm not as familiar in vue, but seems good.

PR #793, opened 2020-11-23
Changed attribute's name

praneethpr:


fightingtexasaggie (commented):
the src/Message/DeleteFile.php still references a getFilePath() call.

src/Migrations/Version20201123210350.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20201123210350 extends AbstractMigration
```
mickel1138, line 11: Rather than making a new migration, should we just change to earlier migration that creates these?

praneethpr: I thought about it too, but then we need to change the data migration sql scripts. I feel this is more cleaner

praneethpr:
> the src/Message/DeleteFile.php still references a getFilePath() call. Also assets/static/js/download.js references data.dataset.filename.

I think those are independent, which have the same method name

fightingtexasaggie (commented):
This is causing a 500 on post.
src/Controller/Api/DatasetSubmissionController.php still calls setFileName

mickel1138:
> still a few calls to either setFileName or getDatasetFileName
> src/Controller/Api/DatasetSubmissionController.php still calls setFileName
> src/Controller/UI/DatalandController.php
> src/Controller/DownloadController.php

Found one here: 
https://github.com/griidc/pelagos/pull/793/files#diff-b3c6f6b7e7d480d10a942df2622e4b5abcebe562290ed084fe4d31d3811b1c1dR381

praneethpr:
> This is causing a 500 on post.
> src/Controller/Api/DatasetSubmissionController.php still calls setFileName

Merged the epic into this branch, will rename them

praneethpr:
> > still a few calls to either setFileName or getDatasetFileName
> > src/Controller/Api/DatasetSubmissionController.php still calls setFileName
> > src/Controller/UI/DatalandController.php
> > src/Controller/DownloadController.php
> 
> Found one here:
> https://github.com/griidc/pelagos/pull/793/files#diff-b3c6f6b7e7d480d10a942df2622e4b5abcebe562290ed084fe4d31d3811b1c1dR381

DownloadController uses legacy getDatasetFileName() from DatasetSubmission Entity, which has pointer to the first File's filename, eventually those methods will be gone. Those are independent from the attributes in this PR

fightingtexasaggie (approved):
Everything seems to be working now after the last commit.

PR #794, opened 2020-11-25
Replaced colons with periods in file names and zip file names

praneethpr:


fightingtexasaggie (approved):
tested and works, however, we should probably slap the user's hand on the front end if they enter undesirable characters too

mickel1138:
> tested and works, however, we should probably slap the user's hand on the front end if they enter undesirable characters too

Add a ticket for it.

mickel1138:
I don't think we should do this, the API should just outright reject files with colons in it. Is this even possible in any filesystem?

src/MessageHandler/ZipDatasetFilesHandler.php
```diff
@@ -96,7 +96,7 @@ public function __invoke(ZipDatasetFiles $zipDatasetFiles)
         $datasetSubmissionId = $zipDatasetFiles->getDatasetSubmissionId();
         $datasetSubmission = $this->entityManager->getRepository(DatasetSubmission::class)->find($datasetSubmissionId);
         $filesInfo = $this->fileRepository->getFileNameAndPath($fileIds);
-        $destinationPath = $this->downloadDirectory . DIRECTORY_SEPARATOR . $datasetSubmission->getDataset()->getUdi() . '.zip';
+        $destinationPath = $this->downloadDirectory . DIRECTORY_SEPARATOR . str_replace(':', '.', $datasetSubmission->getDataset()->getUdi()) . '.zip';
```
mickel1138, line 99: This can stay tho...

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -378,7 +378,7 @@ public function addFile(DatasetSubmission $datasetSubmission, Request $request,
             $datasetSubmission->setFileset($fileset);
         }
         $newFile = new File();
-        $newFile->setFileName(trim($request->get('name')));
+        $newFile->setFileName(trim(str_replace(':', '.', $request->get('name'))));
```
mickel1138, line 381: I wouldn't do this here, or at all. The API should reject files with colons I feel. We can't change expectations.

praneethpr: Most operating systems, don't allow file creation with ":". Atleast on macOS and windows, this is just an additional check William wanted

src/MessageHandler/ZipDatasetFilesHandler.php
praneethpr: William wanted the zipFileName also without the colon

PR #796, opened 2020-12-04
Original datastore migration command.

mickel1138:


praneethpr (approved):
Looks good

fightingtexasaggie (commented):
expected a larger number of files to be changed during testing, considering the number of test datasets on disk.

src/Migrations/Version20200625131822.php
```diff
@@ -29,7 +32,7 @@ public function up(Schema $schema) : void
...
                 $this->addSql("UPDATE dataset_submission SET fileset_id = currval('public.fileset_id_seq') WHERE id = $id");
-                $this->addSql("INSERT INTO file (id, fileset_id, creation_time_stamp, creator_id, modification_time_stamp, modifier_id, file_name, file_size, file_sha256_hash, file_path, status) SELECT nextval('file_id_seq'), currval('public.fileset_id_seq'), CURRENT_TIMESTAMP(0), 0, CURRENT_TIMESTAMP(0), 0, dataset_file_name, dataset_file_size, dataset_file_sha256_hash, '$udi/$udi.dat', 'done' FROM dataset_submission WHERE id = $id");
+                $this->addSql("INSERT INTO file (id, fileset_id, creation_time_stamp, creator_id, modification_time_stamp, modifier_id, file_path_name, file_size, file_sha256_hash, physical_file_path  , status) SELECT nextval('file_id_seq'), currval('public.fileset_id_seq'), CURRENT_TIMESTAMP(0), 0, CURRENT_TIMESTAMP(0), 0, dataset_file_name, dataset_file_size, dataset_file_sha256_hash, '$udi/$udi.dat', 'done' FROM dataset_submission WHERE id = $id");
```
praneethpr, line 35: I think you need to set "uploaded_at" field to "submission_time_stamp"

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,128 @@
...
+            $file->setFilePathName($datasetSubmission->getDatasetFileName());
+            $file->setFileSize($datasetSubmission->getDatasetFileSize());
+            $file->setFileSha256Hash($datasetSubmission->getDatasetFileSha256Hash());
+
+            $file->setUploadedAt($datasetSubmission->getModificationTimeStamp());
+            $file->setUploadedBy($datasetSubmission->getModifier());
+
+            $file->setDescription('Original Dataset File');
+
+            $file->setPhysicalFilePath("/san/data/store/$udi/$udi.dat");
```
fightingtexasaggie, line 118: perhaps get from the DATA_STORE_DIRECTORY env variable

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,128 @@
...
+        return 0;
+    }
+    
+    function setFile($datasetSubmission)
+    {
+        
+        if ($datasetSubmission instanceof DatasetSubmission) {
+            $udi = $datasetSubmission->getDataset()->getUdi();
+            
+            // TODO: Run querybuild to get filename and hash and size...
```
fightingtexasaggie: won't the size and hash be the same anyway?

src/Command/PelagosMigrateDatasetFilesCommand.php
mickel1138: It's a parameter now, you review this just as I pushed an update.

src/Command/PelagosMigrateDatasetFilesCommand.php
mickel1138: No, we neutered the getHash and getName function to get them from the first file name in the files list. So it wasn't possible to get these, besides using a querybuilder.

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+            if ($queueFiler === true) {
+                $this->submissionsListForFiler[] = $datasetSubmission->getId();
+            }
+        }
+    }
+
+    /**
+     * Get the file data from the submission, because the settings have been forwarded.
+     *
+     * @param DatasetSubmission $datasetSubmission The Dataset Submission to process.
```
fightingtexasaggie, line 192: Do we need a return annotation?

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+    protected $submissionsListForFiler;
+
+    /**
+     * Class constructor for dependency injection.
+     *
+     * @param EntityManagerInterface $entityManager A Doctrine EntityManager.
+     */
+    public function __construct(EntityManagerInterface $entityManager, MessageBusInterface $messageBus)
+    {
+        $this->submissionsListForFiler = array();
```
praneethpr, line 54: This can be initialized when it is declared on line 45, as the scope of this variable is global

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+        $io = new SymfonyStyle($input, $output);
+
+        $io->title('This is is dataset Migration Tool');
+
+        $queueFiler = $input->getOption('qf') ? true : false;
+        $checkFileExist = $input->getOption('ifc') ? true : false;
+
+        $datasetRepository = $this->entityManager->getRepository(Dataset::class);
+
+        $datasets = $datasetRepository->findAll();
```
praneethpr, line 91: Can be chained with previous one

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Get the file data from the submission, because the settings have been forwarded.
+     *
+     * @param DatasetSubmission $datasetSubmission The Dataset Submission to process.
+     */
+    protected function getFileData($datasetSubmission)
+    {
+        $datasetSubmissionid = $datasetSubmission->getid();
```
praneethpr, line 196: isn't it getId()?

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+     * The message bus for dispatching the filer message.
+     *
+     * @var MessageBusInterface $messageBus
+     */
+    protected $messageBus;
+
+    /**
+     * This contains a list of submission IDs that need to be queued to filer.
+     *
+     * @var Array $submissionsListForFiler
```
praneethpr, line 43: Need to change it to "array"

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+use App\Entity\Dataset;
+use App\Entity\DatasetSubmission;
+use App\Entity\File;
+use App\Entity\FileSet;
+use App\Message\DatasetSubmissionFiler;
+
+use Doctrine\ORM\EntityManagerInterface;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
```
praneethpr, line 14: Unused use statement

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+            ->addOption('qf', null, InputOption::VALUE_NONE, 'Queue Dataset to Filer')
+            ->addOption('ifc', null, InputOption::VALUE_NONE, 'Do not check if dataset file exist on disk.')
+        ;
+    }
+
+    /**
+     * The symfony command execute function.
+     *
+     * @param InputInterface  $input  The Symfony Console Input.
+     * @param OutputInterface $output The Symfony Console Output.
```
praneethpr, line 78: Missing return annotation

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+
+        $io->success('Done!');
+
+        return 0;
+    }
+
+    /**
+     * Set file data for the dataset submission.
+     *
+     * @param DatasetSubmission  $datasetSubmission The Dataset Submission to process.
```
praneethpr, line 144: null type not included in the annotation

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+class PelagosMigrateDatasetFilesCommand extends Command
+{
+    protected static $defaultName = 'pelagos:migrate-dataset-files';
```
praneethpr, line 24: Missing annotation

praneethpr (commented):
Tested it with /san/data/store on griidc-dev, checked Files entity, 2000+ files got created. I'm assuming that's right

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+
+        return 0;
+    }
+
+    /**
+     * Set file data for the dataset submission.
+     *
+     * @param DatasetSubmission  $datasetSubmission The Dataset Submission to process.
+     * @param String             $dataStore         The path to the datatore.
+     * @param Boolean            $queueFiler        Should this also be queued to the filer.
```
praneethpr, line 146: Return type not specified

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+
+            $file->setUploadedAt($datasetSubmission->getModificationTimeStamp());
+            $file->setUploadedBy($datasetSubmission->getModifier());
+
+            $file->setDescription('Original Dataset File');
+
+            $file->setPhysicalFilePath($fileName);
+            $file->setStatus($queueFiler ? File::FILE_NEW : File::FILE_DONE);
+
+            $fileset = $datasetSubmission->getFileset();
```
praneethpr, line 177: Not used

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+     * @param OutputInterface $output The Symfony Console Output.
+     */
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $io->title('This is is dataset Migration Tool');
+
+        $queueFiler = $input->getOption('qf') ? true : false;
+        $checkFileExist = $input->getOption('ifc') ? true : false;
```
praneethpr, line 87: This option isn't being used anywhere

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+    }
+
+    /**
+     * Get the file data from the submission, because the settings have been forwarded.
+     *
+     * @param DatasetSubmission $datasetSubmission The Dataset Submission to process.
+     */
+    protected function getFileData($datasetSubmission)
```
praneethpr, line 194: Type not declared

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -0,0 +1,209 @@
...
+     * @var Array $submissionsListForFiler
+     */
+    protected $submissionsListForFiler;
+
+    /**
+     * Class constructor for dependency injection.
+     *
+     * @param EntityManagerInterface $entityManager A Doctrine EntityManager.
+     */
+    public function __construct(EntityManagerInterface $entityManager, MessageBusInterface $messageBus)
```
praneethpr, line 52: Missing annotation for $messageBus

fightingtexasaggie (approved):
While I am concerned about using user-generated filenames internally, (unforseen consequences of tools that don't handle weird filenames right, etc) it does work as intended. Our approach of handling files via streams will help mitigate effects of weird filenames on 3rd party products outside of our control that may or may not be handling them correctly. (ClamAV, etc)

src/Command/PelagosMigrateDatasetFilesCommand.php
mickel1138: Yes, but how does it work then? I think it's some kind of "magic method" because id is always part of the entity.
Anyway, it's changed.

src/Command/PelagosMigrateDatasetFilesCommand.php
mickel1138: I was confused with the PSR-12, but the guidance is to use lower case short (if possible, like int, bool) statements. I thought it was the other way around.

src/Command/PelagosMigrateDatasetFilesCommand.php
mickel1138: I can either leave of the type, or remove null. Or can check if it's even a submissions before the call.

src/Command/PelagosMigrateDatasetFilesCommand.php
mickel1138: It doesn't return anything, do we put void?

praneethpr (approved):

PR #797, opened 2020-12-07
Feature/pelagos 4309 make get files api to return json file folder structure

praneethpr:


assets/js/vue/FileManager.vue
```diff
@@ -40,9 +40,10 @@ export default {
 };
 
 function getItems(pathInfo) {
+    console.log(pathInfo)
```
mickel1138, line 43: Console log

src/Entity/File.php
```diff
@@ -319,4 +319,21 @@ public function setStatus(string $status): void
...
     }
+
+    /**
+     * Get file path parts in an array.
+     *
+     * @param string $path The filepath which needs to exploded into parts.
+     *
+     * @return array
+     */
+    public function getFilePathParts(string $path = '') : array
```
mickel1138, line 330: I like the look of this, much cleaner.

src/Util/FolderStructureGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+            $isDir = count($filePathParts) > 1;
+            if (!empty($folderJson) and array_search($filePathParts[0], array_column($folderJson, 'name'))) {
+                continue;
+            } else {
+                if (!$isDir) {
+                    $folderJson[] = array(
+                        'name' => $filePathParts[0],
+                        'isDirectory' => false,
+                        'size' => $file->getFileSize(),
+                        'dateModified' => $file->getUploadedAt()->format('Y/m/d')
```
praneethpr, line 50: This might break now for older files, because this attribute value isn't backfilled in the migration. We should backfill this value to Dataset Submission's SubmissionTimeStamp attribute

src/Util/FolderStructureGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+                    );
+                } else {
+                    $folderJson[] = array(
+                        'name' => $filePathParts[0],
+                        'isDirectory' => true,
+                        'hasSubDirectories' => (count($filePathParts) > 2)
```
mickel1138, line 56: Is a bool cast needed?

praneethpr: Just to be sure, so that its not a string. And i think this doesn't cover other scenarios. Maybe we should remove hasSubDirectories

mickel1138 (approved):
Just minor issue on first load, but will be fixed later.
Date doesn't show right away until refresh button it clicked.

fightingtexasaggie (approved):
tested with 100's of folders/files, tested with file app and with manual api calls. working well as best I can tell.

PR #799, opened 2020-12-09
Added delete functionality to file manager

praneethpr:


assets/js/vue/FileManager.vue
```diff
@@ -37,9 +40,16 @@ export default {
         // Assigning it to the global variable of this class so that functions outside the export can use it
         datasetSubmissionId = this.datasetSubId;
     },
+
+    methods: {
+        onErrorOccurred: function(e) {
+            e.errorText = 'Cannot delete folders';
```
praneethpr, line 46: Placeholder code, not sure how to call this method from the reject() in the Promise methods below when there is an exception

fightingtexasaggie (approved):
tested this out and it works, however, currently nothing cleans up the pre-submission files that have been deleted. Also, post-submission, all files in the HOMEDIR_PREFIX location are left in place, so they'd build up fast if we don't clean them.

mickel1138 (approved):
Looks good for the ticket.

PR #800, opened 2020-12-09
Feature/pelagos 4314 voter for files api

fightingtexasaggie:
not sure how we're going to tie in the API actions just yet. I can tie in DELETE for now. Also, is it too early to add a READ check to the voter?


src/Security/Voter/PelagosEntityVoter.php
```diff
@@ -29,6 +29,7 @@ abstract class PelagosEntityVoter extends Voter
 {
     const CAN_CREATE = 'CAN_CREATE';
     const CAN_EDIT = 'CAN_EDIT';
+    const CAN_READ   = 'CAN_READ';
```
mickel1138, line 32: Everyone can always read, we probably don't have to vote on that.

src/Security/Voter/FileVoter.php
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * The Symfony calling security framework calls supports before calling voteOnAttribute.
+     *
+     * @param string         $attribute The action to be considered.
+     * @param mixed          $subject   A Dataset.
+     * @param TokenInterface $token     A security token containing user authentication information.
+     *
+     * @return boolean True If the user has one of the target roles for any of the subject's DataRepositories.
+     */
+    protected function voteOnAttribute($attribute, $subject, TokenInterface $token) //phpcs:ignore
```
praneethpr, line 54: I think this voter should be the same as DatasetSubmissionVoter

src/Security/Voter/PelagosEntityVoter.php
fightingtexasaggie: I was thinking of restricted mode later down the line, but yeah, we can implement that when we need to.

src/Security/Voter/FileVoter.php
fightingtexasaggie: Do you think it should consider the state of the submission as well?

fightingtexasaggie:
Yesterday we put our heads together and realized that the approach to use a querybuilder within a file voter wasn't the best approach. I added commits to add a RRAR relationship from Fileset back to DatasetSubmission, but discussing this, we decided it would result in a circular reference making either object undeletable. 

By using some of the Doctrine tools, we determined what seemed to be the correct way of establishing these bi-directional OneToOne relationships, without the RRAR construct. What we still have left to determine is whether or not a separate voter is needed, or if the DatasetSubmission voter could be extended with additional attributes to handle File. The Api works on objects almost stand-alone, imho, and I don't think it will be able to relate File/Fileset to DatasetSubmission simply by virtue of it being in the same voter. I'm under the impression we will need a separate File voter that can be referenced in the File API. This voter will need the bi-directional relationship to figure out it's related DatasetSubmission (so that research groups can be found, etc)

src/Entity/File.php
```diff
@@ -120,6 +120,11 @@ class File extends Entity
      */
     protected $status = self::FILE_NEW;
 
+    /**
+     * ORM\OneToOne(targetEntity=DatasetSubmission::class, mappedBy="File", cascade={"persist", "remove"})
+     */
+    private $datasetSubmission;
```
mickel1138, line 126: File should NOT have relationship with DatasetSubmission, it SHOULD have a relationship with FileSet

fightingtexasaggie: This was an overlooked mistake. I'd already updated Fileset but left this in from earlier.

mickel1138 (approved):

praneethpr (approved):

PR #802, opened 2020-12-11
Feature/pelagos 4322 dropzone and dev extreme file manager

praneethpr:


assets/js/vue/FileManager.vue
```diff
@@ -1,7 +1,10 @@
 <template>
     <div ref="dev">
-        <DxFileManager :file-system-provider="customFileProvider" :on-error-occurred="onErrorOccurred">
-            <DxPermissions :delete="true"/>
+        <DxFileManager
```
fightingtexasaggie, line 3: extra whitespace?

fightingtexasaggie (commented):
There's still a situation where deleting a file inside a folder results in the deletion of another folder altogether. I'll put together a test case and see if I can make it repeatable. I first noticed it on the other ticket, but couldn't get it to repeat so blamed my testing.

Empty folders disappear, but I've also confirmed folders with content disappear too if deleting from other folders.

praneethpr:
> There's still a situation where deleting a file inside a folder results in the deletion of another folder altogether. I'll put together a test case and see if I can make it repeatable. I first noticed it on the other ticket, but couldn't get it to repeat so blamed my testing.
> 
> Empty folders disappear, but I've also confirmed folders with content disappear too if deleting from other folders.
> 
> Example: unzip and upload folder from https://staff.tamucc.edu/mwilliamson/testdir3.zip
> Delete files in each of the folders and notice sometimes other folders disappear. It isn't always consistent.

The destination directory path is null sometimes, I guess that's why its causing this. I'm still trying to understand how the promises work, it looks like the first time we upload the variable is set but after that the variable is always empty

fightingtexasaggie (approved):
It looks clean and tested clean, but as a sidenote I could not drag in more than one object at a time, unless I was dragging in a folder. Trying to drag in multiple objects the control didn't seem to allow, but then again, this is likely a browser limitation. Tested up to 10mb file too.

praneethpr:
> It looks clean and tested clean, but as a sidenote I could not drag in more than one object at a time, unless I was dragging in a folder. Trying to drag in multiple objects the control didn't seem to allow, but then again, this is likely a browser limitation. Tested up to 10mb file too.

Can i know which browser you tested on ? It works for me, Chrome on macOS

assets/js/vue/FileManager.vue
```diff
@@ -73,13 +97,86 @@ const deleteItem = (item) => {
...
         } else {
             reject(new Error('Cannot delete folders'));
         }
     })
 }
 
+const uploadFileChunk = (fileData, uploadInfo, destinationDirectory) => {
+    destinationDir = destinationDirectory.path;
+    return new Promise((resolve, reject) => {
+        resolve();
```
mickel1138, line 111: We might be able to use this function instead of the dropZone drop area. But for now we'll move on to test.

mickel1138 (approved):
Approved so it can be tested. I think lots of optimization tickets will come out of testing.

PR #803, opened 2020-12-15
Added missing async message to prod config

praneethpr:


mickel1138:
Ah, that's why I didn't get any zip files.

praneethpr:
> Ah, that's why I didn't get any zip files.

I think we missed it when we were added the other ones

fightingtexasaggie (approved):
confirmed the zipfile is now generated.

PR #806, opened 2021-01-05
Added move and rename for files

praneethpr:


fightingtexasaggie (approved):
This looks pretty solid. Both the rename file in place and move worked, and the resulting submission contained the expected directory structure on disk. This is pretty impressive.

src/Controller/Api/FileController.php
```diff
@@ -51,7 +53,48 @@ public function deleteFile(File $file, MessageBusInterface $messageBus, EntityMa
...
+     * @return Response A response object with an empty body and a "no content" status code.
+     */
+    public function updateFileName(File $file, EntityManagerInterface $entityManager, Request $request, Datastore $datastore) : Response
+    {
+        $newFileName = $request->get('destinationDir');
+        $fileset = $file->getFileset();
+        if (!$fileset->doesFileExist($newFileName)) {
+            // Rename file on disk if it is already processed
+            if ($file->getStatus() === File::FILE_DONE) {
+                $datastore->renameFile($file->getFilePathName(), $newFileName);
```
praneethpr, line 87: First parameter needs to be physicalFilePath, and renameFile needs to make a newFileName based on Datastore standards

src/Controller/Api/FileController.php
```diff
@@ -51,7 +53,48 @@ public function deleteFile(File $file, MessageBusInterface $messageBus, EntityMa
...
+            }
+            $file->setFilePathName($newFileName);
+            $entityManager->flush();
+        } else {
+            throw new BadRequestHttpException('File with same name and folder already exists');
+        }
+
+        return new Response(
+            null,
+            Response::HTTP_OK,
```
praneethpr, line 97: Return HTTP_NO CONTENT

src/Util/Datastore.php
```diff
@@ -111,6 +108,22 @@ public function deleteFile(string $filePath): bool
      */
     public function renameFile(string $oldFilePath, string $newFilePath): bool
     {
-        return $this->datastoreFlysystem->rename($oldFilePath, $newFilePath);
+        return $this->datastoreFlysystem->rename($oldFilePath, $this->makeFileName($newFilePath));
```
fightingtexasaggie, line 111: We also need to inform the database of the filename change.

PR #808, opened 2021-01-06
Feature/pelagos 4336 add download api and functionality in file manager

praneethpr:
I've merged move and rename file branch into this, just to avoid merge conflicts

### DIFF
--- a/src/Controller/Api/FileController.php
+++ b/src/Controller/Api/FileController.php
@@ -104,7 +104,7 @@ class FileController extends AbstractFOSRestController
     }
 
     /**
-     * Update a file entity.
+     * Download a file from disk.
      *
      * @param File                   $file          File entity instance.
      * @param Datastore              $datastore     Datastore to manipulate the file on disk.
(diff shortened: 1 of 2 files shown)


src/Controller/Api/FileController.php
```diff
@@ -51,10 +55,88 @@ public function deleteFile(File $file, MessageBusInterface $messageBus, EntityMa
...
+    /**
+     * Download a file from disk.
+     *
+     * @param File                   $file          File entity instance.
+     * @param Datastore              $datastore     Datastore to manipulate the file on disk.
+     *
+     * @Route("/api/file/download/{id}", name="pelagos_api_file_download", defaults={"_format"="json"})
+     *
+     * @IsGranted("IS_AUTHENTICATED_FULLY")
+     * @IsGranted("CAN_EDIT", subject="file")
```
fightingtexasaggie, line 115: as soon as the voter is ready, we'll have to make such that the user has to be in the same RG as the file, but this is fine for now imho.

fightingtexasaggie (dismissed):
Saw your comment about the move/rename stuff merged in...makes sense. I also tested and DL works as expected from the submission in-progress, so looks good.

fightingtexasaggie (approved):

mickel1138 (approved):
Works. But eventually should work on folder support.

PR #809, opened 2021-01-06
Feature/pelagos 4337 fix file delete api to delete files from disk

praneethpr:


src/Controller/Api/FileController.php
```diff
@@ -38,15 +37,15 @@ public function deleteFile(File $file, MessageBusInterface $messageBus, EntityMa
         $fileset->removeFile($file);
         $filePath = $file->getPhysicalFilePath();
         if ($file->getStatus() === File::FILE_NEW) {
-            $deleteFileMessage = new DeleteFile($filePath);
-            $messageBus->dispatch($deleteFileMessage);
+            unlink($file->getPhysicalFilePath());
+            rmdir(dirname($file->getPhysicalFilePath()));
```
fightingtexasaggie, line 41: Might the directory still have other files in it that were not intended to be deleted? I know rmdir fails and warns if the directory isn't empty, but I just think we shouldn't rely on that.

praneethpr: I'm only deleting the "New" files, which are in the uploads directory. Each file has its own UUID folder, so there wouldn't be anything in this folder, but i could add a check for being extra safe

fightingtexasaggie (approved):
ok, this seems to be working fine now. in the FILE_NEW state, they aren't in the store location, so each has their own folder, so this makes sense every file deletion will have its own directory. I think it's good to go.

PR #810, opened 2021-01-08
Added API to get all file ids in the given path, used it to delete folders

praneethpr:


fightingtexasaggie (approved):
The refresh option is not always present, in cases where I feel it should be. To illustrate, submit a folder, delete something, then try and drag in another folder. It works, but the view is not refreshed nor is the control available to manually refresh it. This isn't really a bug in this particular delete ticket though, so passing, but I wanted to document it somewhere for discussion.

praneethpr:
> The refresh option is not always present, in cases where I feel it should be. To illustrate, submit a folder, delete something, then try and drag in another folder. It works, but the view is not refreshed nor is the control available to manually refresh it. This isn't really a bug in this particular delete ticket though, so passing, but I wanted to document it somewhere for discussion.

Yes, it is an issue with their css layout. As soon as you redraw or resize your browser, the controls appear back in. Even when it doesn't appear on the screen, its actually there but outside the container width

assets/js/vue/FileManager.vue
```diff
@@ -108,7 +101,17 @@ const deleteItem = (item) => {
                     reject(error)
                 })
         } else {
-            reject(new Error('Cannot delete folders'));
+            axiosInstance
+                .get(`${Routing.generate('pelagos_api_get_file_ids_dataset_submission')}/${datasetSubmissionId}?path=${item.path}`)
+                .then(response => {
+                    response.data.forEach(id => {
+                        axiosInstance.delete(`${Routing.generate('pelagos_api_file_delete')}/${id}`);
```
mickel1138, line 108: Why not have an API that deletes files,  rather than this handoff, it will just slow things down.

mickel1138: You can even call a route from the controller itself!

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -376,6 +376,48 @@ public function getFile(DatasetSubmission $datasetSubmission, Request $request):
...
+     *     methods={"GET"},
+     *     defaults={"_format"="json"},
+     *     requirements={"id"="\d+"}
+     *     )
+     *
+     * @View()
+     *
+     * @return Response
+     */
+    public function getFileIds(DatasetSubmission $datasetSubmission, Request $request): Response
```
mickel1138, line 397: If this route is only use to delete files from a folder, then just call this function deleteFilesFromPath and actually delete them here. If it's in the same controller you may have to make the delete function a common function, but you can also call it from here. You can even all a routed controller from another controller.
It's called forwarding: https://symfony.com/doc/current/controller/forwarding.html

assets/js/vue/FileManager.vue
praneethpr: So, you are saying have a delete folder API, which takes in multiple file ids? We definitely would need the getFileIDs API, as we would do the same thing for moving/renaming folders as well

src/Controller/Api/DatasetSubmissionController.php
praneethpr: We would be using this API, in the upcoming ticket for moving/renaming Folders

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -376,6 +376,48 @@ public function getFile(DatasetSubmission $datasetSubmission, Request $request):
...
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
```
mickel1138, line 399: For API we could starting using the `@IsGranted` annotation, because a "fancy" screen isn't usually needed for API's.

src/Controller/Api/DatasetSubmissionController.php
mickel1138: Ok, we should keep this then, but maybe make a new function/route.

mickel1138 (approved):
Approved, but discussed and we'll refactor the for each loop for delete.

PR #817, opened 2021-01-15
Added single delete api for file manager to use

praneethpr:


src/Controller/Api/FileManager.php
```diff
@@ -0,0 +1,98 @@
...
+     * @Route(
+     *     "/api/file_delete/{id}",
+     *     name="pelagos_api_file_delete",
+     *     methods={"DELETE"},
+     *     defaults={"_format"="json"},
+     *     requirements={"id"="\d+"}
+     *     )
+     *
+     * @IsGranted("IS_AUTHENTICATED_FULLY")
+     * @IsGranted("CAN_DELETE", subject="datasetSubmission")
```
mickel1138, line 39: Should this voter subject be changed? I mean this actually works. We may not really need a file voter if the submission Id is passed along.

mickel1138 (approved):
Tested, and reviewed.

PR #819, opened 2021-01-15
Add additional details to remotely hosted failure report.

fightingtexasaggie:
To test, make sure and insert a bad URL in an active remote dataset submission.

i.e.
update dataset_submission set dataset_file_uri = 'https://www.tamucc.edu/data/R4.x257.229:0003'  where id = 11092; # works with current dump, ymmv.


praneethpr (approved):
Tested and works fine

PR #822, opened 2021-01-26
Fixed the owning side of the Fileset DatasetSubmission relationship

praneethpr:


mickel1138 (approved):
This is correct name of the property, and class referencing is more consistent now.

PR #825, opened 2021-01-27
Added additional doctrine event to see if fileset is done, THEN zip f…

mickel1138:
…iles.

### DIFF
--- a/src/Event/FileProcessListener.php
+++ b/src/Event/FileProcessListener.php
@@ -58,7 +58,8 @@ class FileProcessListener
      *
      * @return void
      */
-     public function postUpdate(File $file, LifecycleEventArgs $args): void
+    
+    public function postUpdate(File $file, LifecycleEventArgs $args): void
     {
         if (!$file instanceof File) {
             return;
@@ -80,7 +81,6 @@ class FileProcessListener
             // Dispatch message to zip files
             $zipFiles = new ZipDatasetFiles($fileIds, $datasetSubmissionId);
             $this->messageBus->dispatch($zipFiles);
-
         }
     }
 }
--- a/src/Event/FileProcessListener.php
+++ b/src/Event/FileProcessListener.php
@@ -3,13 +3,32 @@
 namespace App\Event;
 
 use App\Entity\File;
-
+use App\Entity\DatasetSubmission;
+use App\Message\ZipDatasetFiles;
 use Doctrine\ORM\Event\LifecycleEventArgs;
+use Symfony\Component\Messenger\MessageBusInterface;
 
 class FileProcessListener
 {
     /**
-     * Post update method for dataset entity, to publish/update Doi.
+     * The Message Bus.
+     *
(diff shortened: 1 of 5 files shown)


src/MessageHandler/ZipDatasetFilesHandler.php
```diff
@@ -74,14 +82,16 @@ public function __construct(
         string $downloadDirectory,
         FileRepository $fileRepository,
         EntityManagerInterface $entityManager,
-        Datastore $datastore
+        Datastore $datastore,
+        EntityEventDispatcher $entityEventDispatcher
```
praneethpr, line 86: Annotation missing for $entityEventDispatcher

src/MessageHandler/ZipDatasetFilesHandler.php
```diff
@@ -107,6 +117,9 @@ public function __invoke(ZipDatasetFiles $zipDatasetFiles)
             fclose($outputStream['fileStream']);
             $datasetSubmission->getFileset()->setZipFilePath($destinationPath);
             $this->entityManager->flush();
+
+            // Dispatch entity event.
+            $this->entityEventDispatcher->dispatch($datasetSubmission, 'dataset_zipped');
```
praneethpr, line 122: I think we can remove this, and add it when we are going to listen for this event?

mickel1138: Not every event always has a listener. But in our case we might remove it.

praneethpr (approved):

src/Event/FileProcessListener.php
```diff
@@ -30,4 +49,38 @@ public function preUpdate(File $file, LifecycleEventArgs $args): void
...
+        }
+
+        $fileset = $file->getFileset();
+        $datasetSubmission = $fileset->getDatasetSubmission();
+
+        if (!$datasetSubmission instanceof DatasetSubmission) {
+            return;
+        }
+
+        if ($fileset->isDone()) {
```
fightingtexasaggie, line 75: Since this is getting called with every single file, this method is going to get called a lot, I worry a little about the efficiency of the criteria call, but I guess we can optimize it later if it becomes an issue.

fightingtexasaggie (approved):
I'm assuming y'all decided not to wait for dataset review to be accepted before generation of the zipfile, because that's how it's currently working. Approving contingent on that being the intent.

src/Event/FileProcessListener.php
mickel1138: This call is being resolved in memory, 

> I'm assuming y'all decided not to wait for dataset review to be accepted before generation of the zipfile, because that's how it's currently working. Approving contingent on that being the intent.

No, approval is not a factor in zipping.

fightingtexasaggie:
I just re-tested with 1GB of data spread across 100 files and quite a few of them were skipped. I'm thinking maybe there was an http timeout, but will now retest in async mode.

PR #826, opened 2021-01-28
All API errors will display as pop ups to the user

praneethpr:


fightingtexasaggie (commented):
errors are being displayed now, however, I am also noticing any time I try and rename a folder, (which it seems to let me), I get the same " File with same name and folder already exists" regardless of what I try and name it to.

Errors are propagating and being displayed, so if this issue belongs on a different ticket, comment, and I'll approve.

fightingtexasaggie (approved):
errors are being displayed as expected, including the current error (which turns into a nice test case for this PR) that is displayed when renaming folders.

PR #828, opened 2021-01-29
Added upload button and corresponding events to tie it to dropzone

praneethpr:


fightingtexasaggie (approved):
Reviewed and discussed in group, and tested. Works fine in Firefox too.

PR #830, opened 2021-01-29
Fixed "bug" for duplicate file names.

mickel1138:


praneethpr (approved):
Works fine

src/Util/RenameDuplicate.php
```diff
@@ -27,7 +27,7 @@ public function renameFile(string $filePathName) :string
         $fileName = $pathParts['filename'];
         $extension = $pathParts['extension'] ?? '';
 
-        $patterns = array('/^(.*)\((\d+)\)(\.?.*)$/','/(^((?!\.|\(\d+\)).)*$)()/');
+        $patterns = array('/^(.*)\((\d+)\)(\.?.*)$/','/(^(?:(?!\.|\(\d+\)).)*$)()/');
```
praneethpr, line 30: Tested with Filename: "filedata-1.dat", on uploading a duplicate file. It renames the 2nd file to "filedata-1(2).dat", it should be "filedata-1(1).dat". The numbering of duplicate files should start from 1

praneethpr (commented):
The regex needs to be modified a little bit more, it almost works

src/Util/RenameDuplicate.php
mickel1138, line 30:
```suggestion
        $patterns = array('/^(.*)\((\d+)\)(\.?.*)$/','/(^(?:(?!\(\d+\)).)*$)()/');
```

praneethpr (approved):

fightingtexasaggie (approved):

fightingtexasaggie:
We'll have to explain to the PO how it fills missing sequence numbers, i.e. if there's a c.data and a c(2).data, the next use of c.data will result in a c(1).data, not a c(2).data. 

If you have a c.data, c(1).data, c(2).data, c(3).data, and a c(5).data the addition of a new c.data results in the expected c(4).data, but if instead a c(5).data is added, it does not fill the gap. This is a weird edge case I don't think we should worry about, and maybe it's the right behavior anyway.

PR #832, opened 2021-02-04
Added dropdown menu for upload, and added functionality for upload fo…

praneethpr:
…lders


mickel1138 (approved):
Works great, I like this solution.

PR #833, opened 2021-02-08
Feature/pelagos 4368 refactor hash worker to file worker

mickel1138:


src/MessageHandler/ProcessFileHandler.php
```diff
@@ -0,0 +1,158 @@
...
+        $filePath = $file->getPhysicalFilePath();
+        $fileStream = fopen($filePath, 'r');
+        try {
+            $context = hash_init('sha256');
+            hash_update_stream($context, $fileStream);
+            $fileHash =  hash_final($context);
+            $file->setFileSha256Hash($fileHash);
+        } catch (\Exception $e) {
+            $this->logger->error(sprintf('Unable to hash file. Message: %s', $e->getMessage()));
+            return;
```
mickel1138, line 122: Don't return!

src/MessageHandler/ProcessFileHandler.php
```diff
@@ -0,0 +1,158 @@
...
+            'dataset_submission_id' => $datasetSubmission->getId(),
+        );
+
+        $file->setStatus(File::FILE_IN_PROGRESS);
+        $filePath = $file->getPhysicalFilePath();
+        $fileStream = fopen($filePath, 'r');
+        try {
+            $context = hash_init('sha256');
+            hash_update_stream($context, $fileStream);
+            $fileHash =  hash_final($context);
```
mickel1138, line 118: Make this into an utility.

src/Event/FileProcessListener.php
```diff
@@ -37,17 +37,17 @@ public function __construct(MessageBusInterface $messageBus)
      */
     public function preUpdate(File $file, LifecycleEventArgs $args): void
     {
```
fightingtexasaggie, line 39: Commented out in the commit?

mickel1138: This file has to be removed.

tests/unit/Util/StreamInfoTest.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Unit test setup.
+     *
+     * @return void
+     */
+    public function setUp()
+    {
+        $this->root = vfsStream::setup();
+
+        $file = vfsStream::newFile('test.txt')
```
fightingtexasaggie, line 33: Does this test.txt file need to be added to git?

fightingtexasaggie (commented):
Looked over code...quite clean imho, but I'm getting a 500 on DIF creation during testing, noting being logged.

src/Util/StreamInfo.php
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param array $fileStream The file stream array.
+     *
+     * @return string The hash calculated from the stream.
+     */
+    public static function calculateHash(array $inputFileStream, string $algo = 'sha256') :string
+    {
+        $fileStream = $inputFileStream['fileStream'] ?? null;
+        $context = hash_init($algo);
+        hash_update_stream($context, $fileStream);
```
praneethpr, line 21: This would fail if $fileStream is null on type error

src/Util/StreamInfo.php
```diff
@@ -0,0 +1,38 @@
...
+     * Get the file size for the file stream.
+     *
+     * @param array $fileStream The file stream array.
+     *
+     * @return integer The file size of the stream.
+     */
+    public static function getFileSize(array $inputFileStream) :int
+    {
+        $fileStream = $inputFileStream['fileStream'] ?? null;
+        $fstat = fstat($fileStream);
```
praneethpr, line 35: Same here, this would fail if $fileStream is null on type error

src/Util/StreamInfo.php
```diff
@@ -0,0 +1,38 @@
...
+class StreamInfo
+{
+    /**
+     * Get the hash for file stream.
+     *
+     * @param array $fileStream The file stream array.
+     *
+     * @return string The hash calculated from the stream.
+     */
+    public static function calculateHash(array $inputFileStream, string $algo = 'sha256') :string
```
praneethpr, line 17: Can use constant DatasetSubmission::SHA256

src/MessageHandler/ProcessFileHandler.php
```diff
@@ -0,0 +1,166 @@
...
+        $dataset = $datasetSubmission->getDataset();
+        $udi = $dataset->getUdi();
+        $loggingContext = array(
+            'fileId' => $fileId,
+            'dataset_id' => $dataset->getId(),
+            'udi' => $udi,
+            'dataset_submission_id' => $datasetSubmission->getId(),
+        );
+
+        $file->setStatus(File::FILE_IN_PROGRESS);
```
praneethpr, line 124: If we are not going to use this, we should probably get rid of this status

src/MessageHandler/ProcessFileHandler.php
```diff
@@ -0,0 +1,166 @@
...
+     * @param EntityManagerInterface $entityManager           The entity handler.
+     * @param FileRepository         $fileRepository          The file Repository.
+     * @param LoggerInterface        $datasetFileHasherLogger Name hinted dataset_file_hasher logger.
+     * @param Datastore              $datastore               Datastore utility instance.
+     * @param MessageBusInterface    $messageBus              Symfony messenger bus interface instance.
+     */
+    public function __construct(
+        EntityManagerInterface $entityManager,
+        FileRepository $fileRepository,
+        LoggerInterface $datasetFileHasherLogger,
```
praneethpr, line 76: We need to change the logger name, it does more than hashing and it shouldn't be logged in this log file

src/Util/StreamInfo.php
```diff
@@ -0,0 +1,38 @@
...
+        $fileStream = $inputFileStream['fileStream'] ?? null;
+        $context = hash_init($algo);
+        hash_update_stream($context, $fileStream);
+        return hash_final($context);
+    }
+
+    /**
+     * Get the file size for the file stream.
+     *
+     * @param array $fileStream The file stream array.
```
praneethpr, line 28: Incorrect variable name in annotation
```suggestion
     * @param array $inputFileStream The file stream array.
```
```suggestion
     * @param array $fileStream The file stream array.
```

src/Util/StreamInfo.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A utility class to get various information from a file stream.
+ */
+class StreamInfo
+{
+    /**
+     * Get the hash for file stream.
+     *
+     * @param array $fileStream The file stream array.
```
praneethpr, line 13: Need to update annotations for the method

src/MessageHandler/ProcessFileHandler.php
```diff
@@ -0,0 +1,166 @@
...
+        $file->setStatus(File::FILE_DONE);
+
+        if ($fileset->isDone()) {
+            $datasetSubmissionId = $datasetSubmission->getId();
+            $fileIds = array();
+            foreach ($fileset->getProcessedFiles() as $file) {
+                $fileIds[] = $file->getId();
+            }
+            // Dispatch message to zip files
+            $zipFiles = new ZipDatasetFiles($fileIds, $datasetSubmissionId);
```
praneethpr, line 159: We are building the zip after processing every file?

praneethpr: Ok, tested this again. I didn't see the $fileset->isDone() condition above, this is fine 👍

src/Util/StreamInfo.php
mickel1138: Yes, and that's intentionally. I don't want to catch any errors here, because the caller does a try catch anyway, so it's the same scenario if we would call the native functions.

src/Util/StreamInfo.php
mickel1138: Does it exist? Is it better to have the constant in this class?

src/Util/StreamInfo.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * A utility class to get various information from a file stream.
+ */
+class StreamInfo
+{
+    /**
+     * Get the hash for file stream.
+     *
+     * @param array $inputFileStream The file stream array.
```
praneethpr, line 15: Annotation missing for 2nd argument

praneethpr (approved):
Tested and works

fightingtexasaggie (approved):
Be on the lookout for double entry in file/fileset entities, however, this probably was due to my original testing setup. I could not get it to make double entries after trying the same test twice more. Reviewed in group after standup Thurs.

PR #834, opened 2021-02-10
Added zip file size and sha256 hash attributes

praneethpr:


src/Entity/Fileset.php
```diff
@@ -255,4 +273,63 @@ public function setZipFilePath(string $zipFilePath): void
...
+    public function setZipFileSha256Hash(string $zipFileSha256Hash): void
+    {
+        $this->zipFileSha256Hash = $zipFileSha256Hash;
+    }
+
+    /**
+     * Check if zip file exists for the fileset.
+     *
+     * @return bool
+     */
```
fightingtexasaggie, line 325: Would this be more correct with an OR across all 3 attribute instead of the AND? I know they're all rather atomic, so maybe this is more of an academic argument, but I think presence of any of these fields means it exists.

fightingtexasaggie (commented):
What happens if the zipfile creation hasn't yet happened when the submission is persisted?

src/Entity/Fileset.php
praneethpr: True, but we want all these fields to be populated, and it gets populated by the same message handler class ZipDatasetFilesHandler class.

praneethpr:
> What happens if the zipfile creation hasn't yet happened when the submission is persisted?

ZipFile is created after the submission is persisted

fightingtexasaggie (approved):
Tested, confirmed attributes in DB are indeed populated. Concerns resolved earlier about timing issues per discussion with Praneeth. This looks ready to merge, and since contains a migration the sooner the better.

mickel1138 (approved):
Reviewed and tested, looks good.

PR #835, opened 2021-02-15
Added streamedresponse for single file download

praneethpr:


src/Controller/DownloadController.php
```diff
@@ -93,60 +97,71 @@ public function defaultAction(int $id)
...
+        $response = new Response(
+            null,
+            Response::HTTP_NO_CONTENT
         );
+        if ($datasetSubmission instanceof DatasetSubmission) {
+            $fileset = $datasetSubmission->getFileset();
+
+            if ($fileset instanceof Fileset) {
+                if (!$fileset->doesZipFileExist()) {
+                    $filePhysicalPath = $fileset->getProcessedFiles()->first()->getPhysicalFilePath();
```
mickel1138, line 120: Make sure this is something, I tried it with an old dataset, and it didn't work because it didn't have files.

src/Controller/DownloadController.php
```diff
@@ -93,60 +97,71 @@ public function defaultAction(int $id)
...
-        $downloadDirectory = $this->downloadBaseDir . '/' . $uniqueDirectory;
-        mkdir($downloadDirectory, 0755);
-        $datasetFileName  = $dataset->getDatasetSubmission()->getDatasetFileName();
-        symlink(
-            $downloadFileInfo->getRealPath(),
-            $downloadDirectory . '/' . $datasetFileName
+        $datasetSubmission = $dataset->getDatasetSubmission();
+        $response = new Response(
+            null,
+            Response::HTTP_NO_CONTENT
```
mickel1138, line 113: This response does have content, it's streamed content.

src/Controller/DownloadController.php
```diff
@@ -93,60 +97,71 @@ public function defaultAction(int $id)
...
         );
+        if ($datasetSubmission instanceof DatasetSubmission) {
+            $fileset = $datasetSubmission->getFileset();
+
+            if ($fileset instanceof Fileset) {
+                if (!$fileset->doesZipFileExist()) {
+                    $filePhysicalPath = $fileset->getProcessedFiles()->first()->getPhysicalFilePath();
+                    $response = new StreamedResponse(function () use ($dataStore, $filePhysicalPath) {
```
mickel1138, line 121: Make callback smaller, only let it handle the output stream.

src/Controller/DownloadController.php
```diff
@@ -108,26 +108,22 @@ public function defaultAction(int $id)
...
-                        if (is_array($fileStream)) {
-                            stream_copy_to_stream($fileStream['fileStream'], $outputStream);
-                        } else {
-                            throw new BadRequestHttpException('Unable to open file');
-                        }
+                    } catch (\Exception $exception) {
+                        throw new BadRequestHttpException('Unable to open file');
+                    }
+                    $response = new StreamedResponse();
+                    $response->setCallback(function () use ($fileStream) {
```
mickel1138, line 123: You could probably put the callback back in the constructor, but this works too.

praneethpr: Either way I'm ok

mickel1138 (approved):
Tested, works. Now more robust. Approved.

PR #837, opened 2021-02-23
Fixed path.

mickel1138:


praneethpr (approved):

src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -177,7 +175,7 @@ protected function setFile(
 
             $file->setDescription('Original Dataset File');
 
-            $file->setPhysicalFilePath($fileName);
+            $file->setPhysicalFilePath("$udi/$udi.dat");
```
mickel1138, line 178: This is where the change needed to happen, because the path is now relative.
The `$fileName` above is also used for checking if the name exists.

fightingtexasaggie (approved):
debugged with Mike and tested the final changes...working properly now.

praneethpr (approved):

PR #838, opened 2021-02-23
Added imporovements to filer.

mickel1138:
The __destructor didn't file for some reason, so I moved this to the end.

Also when files were deleted, no files has to be processed, so nothing would get zipped, so now it zips when files are done in filer.


praneethpr (approved):

fightingtexasaggie (approved):
tested...works

PR #840, opened 2021-02-24
Resolved issue with extension less files not returning

praneethpr:


fightingtexasaggie (approved):
looks good, ui displayed extensionless file fine, and it was handled properly

PR #841, opened 2021-02-25
Bugfix/rename async config

praneethpr:


mickel1138 (approved):
This needed to added.

PR #842, opened 2021-02-25
Fix/hash zip calc

mickel1138:


praneethpr (approved):
Tested, checked the zip hash on API with calculating hash value on disk. Works

fightingtexasaggie (approved):
zip is working correctly now, verified hash is being calculated correctly, and is being updated with expected hash if files are added in review.

PR #847, opened 2021-02-26
Feature/pelagos 4371 create separate rabbit and messenger startup scripts

fightingtexasaggie:
Can directly test bash script with start/stop/reload command, from your .admin account via sudo. To test .service files, will have to install.


praneethpr (approved):

mickel1138 (dismissed):
I was thinking we could add stop-all-worker command in here. But this script is only for supervisor. So when we do want a restart, maybe we should stop supervisor manually, then stop the workers, then allow the reboot. Or we can see if  there is anything in the queue. Either way, this is good as it is.

fightingtexasaggie:
It's installed on griidc-test.

To deploy:
deploy to /opt/pelagos as usual.
Install system hooks.
sudo cp pelagos/share/bash/control-pelagos-messagemq-consumer.sh /usr/local/bin
sudo cp pelagos/share/service-files/pelagos-consumer.service /usr/lib/systemd/system/

to enable at bootup time:
sudo systemctl enable pelagos-messengemq

to start manually:
sudo systemctl start pelagos-messengemq

to stop manually:
sudo systemctl stop pelagos-messengemq

Summary:
the pelagos-messengemq.service file, when called with a start/stop argument calls the bash script in /usr/local/bin, which (re)starts the supervisord, then reloads any configuration from /opt/pelagos/var/supervisor and starts the consumer up.

mickel1138 (approved):
Seen demonstrated, looks ok.

praneethpr (approved):

PR #848, opened 2021-02-26
Added pruning script to be run in root/cron.

fightingtexasaggie:
This can be triggered by root's crontab. It's safe as it has no user input/output.

### DIFF
--- a/share/bash/prune-orphan-uploads.sh
+++ b/share/bash/prune-orphan-uploads.sh
@@ -4,5 +4,4 @@
 homeDir='/san/home'
 maxAgeDays=30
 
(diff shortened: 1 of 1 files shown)


praneethpr (approved):

fightingtexasaggie:
We need to discuss this with the PO and possible ramification of removing files that are still indexed in the database. Ultimately, I think we should instead create a tool that will remove files that are _not_ referenced in the database (excluding historical submission's fileset''s files). This should then be used after the delete dataset admin tool to clean up submissions that never made it out of draft.

fightingtexasaggie:
Also for disussion, I tihnk we need a missing-file finder...some tool that will traverse all current submissions and check that each and every file is on disk and report otherwise.

PR #850, opened 2021-03-02
Added additional check if zip file exists

praneethpr:


mickel1138 (approved):
Looks good, tested as much as possible.

PR #855, opened 2021-03-10
Removed uploaded file and directory

praneethpr:


mickel1138 (approved):
Tested and works, temp file was removed after filer ran.

PR #860, opened 2021-03-16
Changed order of window.stop, it caused issues.

mickel1138:


fightingtexasaggie (approved):
cancel works, tested after 1/2 way into a 1.4GB file without issue. It doesn't clean it's own chunks, but that's going to be handled anyway by the chunk-cleaner there's a PR for.

PR #861, opened 2021-03-17
Upgraded to 6.5.4 elliptic

mickel1138:
Also upgraded in this branch.


praneethpr (commented):
I'm not sure if we should explicitly add elliptic, probably do a yarn update?

mickel1138:
> I'm not sure if we should explicitly add elliptic, probably do a yarn update?

That's probably better, or we can drop packages.json, but this was the way I could get it to update that package.

PR #867, opened 2021-03-23
Changed permissions for new folders so they can be deleted.

mickel1138:


praneethpr (approved):
Watched it work on griidc-test

PR #871, opened 2021-03-24
Feature/pelagos 4436 add file name shortning function

mickel1138:


src/Util/Datastore.php
```diff
@@ -113,6 +113,7 @@ public function renameFile(string $oldFilePath, string $newFilePath, bool $delet
         if ($deleteFlag === false) {
             $newFilePath = $this->makeFileName($newFilePath);
         }
+        $newFilePath = FileUtilities::fixFileNameLength($newFilePath);
```
praneethpr, line 116: I think we should move this to makeFileName(), so that if it is a new file or rename file operation, it will always fix the file name length

src/Util/ZipFiles.php
```diff
@@ -38,6 +38,7 @@ public function start(array $outputFileStream, string $zipFileName): void
     public function addFile(string $fileName, array $fileStream): void
     {
         if (!empty($fileStream) and is_resource($fileStream['fileStream'])) {
+            $fileName = FileUtilities::fixFileNameLength($fileName);
```
praneethpr, line 41: I don't think we need this here

tests/unit/Util/FileUtilitiesTest.php
```diff
@@ -0,0 +1,53 @@
...
+     * Test calculating the file size.
+     */
+    public function testfixLongFileName()
+    {
+        $fakeFileName = "file/path/" . $this->generateRandomString() . ".txt";
+        $fileName = FileUtilities::fixFileNameLength($fakeFileName);
+        $this->assertLessThanOrEqual(256, strlen(basename($fileName)));
+    }
+
+    private function generateRandomString($length = 256) {
```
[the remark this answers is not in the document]
mickel1138: Ha, it's only a test! :-)

src/Util/Datastore.php
mickel1138: I guess that could work, would the function ever be used externally?

src/Util/ZipFiles.php
mickel1138: I have been looking into this some more, and it was said there is no limit, but a limit was made of 250, because of Windows.
PKWare said the limit is 64k Bytes, but that is split by name field, extra field, and comment field.

I guess this can be removed.

src/Util/Datastore.php
mickel1138: Also, delete bypasses makeFileName, so it becomes TOO long, it fails.

src/Util/FileUtilities.php
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Utilities for files.
+ */
+class FileUtilities
+{
+    /**
+     * A friendly name for this type of entity.
+     */
+    const MAX_FILE_NAME_LENGTH = 256;
```
mickel1138, line 13: Should be 255

src/Util/FileUtilities.php
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @param string $fileName Filename that needs to be fixed.
+     *
+     * @return string
+     */
+    public static function fixFileNameLength(string $fileNamePath, $maxFileNameLength = self::MAX_FILE_NAME_LENGTH) : string
+    {
+        $pathinfo = pathinfo($fileNamePath);
+        $filename = $pathinfo['filename'];
+        $extension = $pathinfo['extension'] ?? '';
```
mickel1138, line 26: Can remove null coalse.

src/Util/FileUtilities.php
```diff
@@ -0,0 +1,45 @@
...
+        if ($dirname === '.') {
+            $dirname = '';
+        } else {
+            $dirname .= '/';
+        }
+        
+        if (!isset($extension)) {
+            $extension = '';
+        } else {
+            $extension = '.' . $extension;
```
mickel1138, line 38: This is wrong!

tests/unit/Util/FileUtilitiesTest.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Test calculating the file size.
+     */
+    public function testfixLongFileName()
+    {
+        $fakeFileName = "file/path/" . $this->generateRandomString() . ".txt";
+        $fileName = FileUtilities::fixFileNameLength($fakeFileName);
+        $this->assertLessThanOrEqual(256, strlen(basename($fileName)));
```
mickel1138, line 41: 255

mickel1138: or use constant

praneethpr (approved):

fightingtexasaggie (approved):

PR #872, opened 2021-03-26
Added deleteDir to async list also.

mickel1138:


mickel1138:
The ZIP file will not delete, because it's owned by Pelagos, we should probably modify the deleteDir worker to also delete the zip file. And rename the worker to something like DeleteDatasetFiles or something.

fightingtexasaggie (approved):
confirmed after this change the deleteDir call is now running true async, under user pelagos, not apache.

src/Event/DatasetSubmissionListener.php
```diff
@@ -214,8 +214,24 @@ public function onAcceptReview(EntityEvent $event)
...
-        $this->messageBus->dispatch($deleteFileMessage);
+        $fileset = $datasetSubmission->getFileset();
+
+        if (!$fileset instanceof Fileset) {
+            $this->logger->warning('No files exist in this dataset.');
+            return;
+        }
+
+        foreach ($fileset->getDeletedFiles() as $deletedFile) {
+            try {
```
mickel1138, line 225: Why try? This is async!

src/MessageHandler/DeleteFileHandler.php
```diff
@@ -56,33 +54,10 @@ public function __construct(LoggerInterface $deleteFileLogger, Datastore $datast
...
-
-        $this->logger->info(sprintf('Processing Dataset Submission with ID: "%s"', $datasetSubmissionId));
-
-        $datasetSubmission = $this->entityManager->getRepository(DatasetSubmission::class)->find($datasetSubmissionId);
-
-        if (!$datasetSubmission instanceof DatasetSubmission) {
-            $this->logger->warning('Dataset Submission was not found.');
-            return;
+        try {
+            $this->datastore->deleteFile($deletedFile->getFilePath());
```
mickel1138, line 58: Check for succes.

praneethpr: Added it in the datastore

praneethpr (approved):

PR #875, opened 2021-03-30
This breaks, because it's not public.

mickel1138:
This static can not be reached, because it's not public.

### DIFF
--- a/src/Util/FileUtilities.php
+++ b/src/Util/FileUtilities.php
@@ -52,7 +52,7 @@ class FileUtilities
      *
      * @return string
      */
-    private function makeFileName(string $fileName) : string
+    public function makeFileName(string $fileName) : string
     {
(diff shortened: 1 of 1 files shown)


src/Util/FileUtilities.php
```diff
@@ -52,7 +52,7 @@ public static function fixFileNameLength(string $fileNamePath, $maxFileNameLengt
      *
      * @return string
      */
-    private function makeFileName(string $fileName) : string
+    public function makeFileName(string $fileName) : string
```
mickel1138, line 55:
```suggestion
    public static function makeFileName(string $fileName) : string
```

praneethpr: This looks good, you can commit them

praneethpr (approved):

fightingtexasaggie (approved):

PR #876, opened 2021-03-30
Added dir filter.

mickel1138:
Do recursive delete
add folder switch


src/Util/Datastore.php
```diff
@@ -96,13 +96,25 @@ public function addFile(array $fileStream, string $filePathName): string
      *
      * @return bool
      */
-    public function deleteFile(string $filePath): bool
+    public function deleteFile(string $filePath, $deleteDir = false): bool
```
praneethpr, line 99: Need annotation for 2nd param

PR #877, opened 2021-03-31
Added switch to delete files outside of the store, owned by Pelagos.

mickel1138:


praneethpr (approved):
Tested and works fine

src/Message/DeleteFile.php
```diff
@@ -11,14 +11,22 @@ class DeleteFile
...
+     */
+    protected $inStore;
+
     /**
      * Constructor.
      *
      * @param string $filePath The file path of the file to be deleted.
      */
-    public function __construct(string $filePath)
+    public function __construct(string $filePath, bool $inStore = true)
```
praneethpr, line 26: Need annotation for the 2nd param

fightingtexasaggie (approved):
tested good, was able to delete zip

praneethpr (approved):

PR #881, opened 2021-04-12
Added additional state of in_queue for files

praneethpr:


src/Entity/File.php
```diff
@@ -31,6 +31,11 @@ class File extends Entity
      */
     const FILE_ERROR = 'error';
 
+    /**
+     * Status value for a File that is added to process queue.
+     */
+    const FILE_IN_QUEUE = 'queue';
```
mickel1138, line 37: Maybe change the value to 'queued' because this may we looked at in the API.

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -126,7 +126,7 @@ public function __invoke(DatasetSubmissionFiler $datasetSubmissionFiler)
                 $this->messageBus->dispatch($zipFiles);
             }
 
-            foreach ($fileset->getNewFiles() as $file) {
+            foreach ($fileset->getQueuedFiles() as $file) {
                 if ($file instanceof File) {
                     $fileId = $file->getId();
                     $processFile = new ProcessFile($fileId);
```
mickel1138, line 132: Move dispatched event to ProcessFileHandler. And remove filesize setter.

mickel1138 (approved):
Reviewed and tested.

PR #892, opened 2021-04-21
Modified text and styling

praneethpr:


assets/js/vue/FileManager.vue
```diff
@@ -61,24 +61,25 @@
...
                 <DxButton
-                    text="Cancel Upload!"
+                    text="Cancel Upload"
                     type="danger"
                     styling-mode="contained"
                     @click="stopProcess"
+                    :element-attr="cancelUploadBtn"
                  />
                  <p>
                     If the filesize is large? Click Cancel!
```
mickel1138, line 85: Should we change this text?

mickel1138 (approved):
Looks good,  but maybe the one lines (about large file) should be removed?

assets/js/vue/FileManager.vue
praneethpr: Just saw that comment from William, will do it

PR #899, opened 2021-05-03
Marx solution.

mickel1138:


fightingtexasaggie (approved):
tested, working much faster.

praneethpr (approved):
LGTM!

config/supervisor/messenger-worker.ini
```diff
@@ -1,7 +1,7 @@
 [program:messenger-consume]
 command=php bin/console messenger:consume async --time-limit=3600
```
mickel1138, line 2: Should I also add `--limit=10` (that means, the worker dies after having processed 10 messages from the queue.)

PR #903, opened 2021-05-05
This should fix it.

mickel1138:


src/Command/PelagosMigrateDatasetFilesCommand.php
```diff
@@ -174,7 +174,7 @@ protected function setFile(
 
             $fileName = "$dataStore/$udi/$udi.dat";
 
-            if (!file_exists($fileName) or !$ignoreFileExistCheck) {
+            if (file_exists($fileName) and $ignoreFileExistCheck) {
```
praneethpr, line 177: We could probably remove this command, because it is a one time thing

fightingtexasaggie: Yeah, I think that's the cleaner approach too. We can always hack it around for testing.

mickel1138: Can we merge it, it was correct at least once, then we will remove it later in another clean up PR.

mickel1138:
Gonna delete this anyway.

PR #904, opened 2021-05-05
Remove rabbitmq commands from scripts

praneethpr:


fightingtexasaggie (approved):
tested...one more annoyance gone! Thank you Praneeth, good catch.

PR #906, opened 2021-05-05
Feature/pelagos 4474 add profile page app

praneethpr:


templates/Person/index.html.twig
```diff
@@ -0,0 +1,17 @@
+{% extends 'base.html.twig'%}
+{% block title %}User Profile{% endblock %}
+
+{%block stylesheets %}
+    {{ parent() }}
+    {{ encore_entry_link_tags('person-profile') }}
+
```
mickel1138, line 7: line

assets/js/vue/pages/PersonProfile.vue
```diff
@@ -0,0 +1,111 @@
...
+
+        axiosInstance
+            .get(Routing.generate('pelagos_api_get_person') + "/" + this.personId)
+            .then(response => {
+                this.personProfileData = response.data;
+                this.showProfile = true;})
+            .catch(error => {
+                console.log(error);
+                this.showProfile = false;
+            });
```
mickel1138, line 72: I don't know if this is a style choice, but we should pick either one tab over like this, or what we have done before, it have this one tab back.

praneethpr: I looked up the Eslint rules for promise chaining, they say that new line for each chained call

src/Controller/Api/PersonController.php
```diff
@@ -462,4 +462,38 @@ public function deleteAction(int $id)
...
+     * Returns a single person.
+     *
+     * @param Person $person The id of the Person.
+     *
+     * @Route(
+     *     "/api/person/{id}",
+     *     name="pelagos_api_get_person",
+     *     methods={"GET"},
+     *     defaults={"_format"="json"},
+     *     requirements={"id"="\d+"}
```
fightingtexasaggie, line 477: maybe we could make id optional, and if logged in and left null, it returns the json of the logged-in user..idk. Works well as-is too.

fightingtexasaggie (approved):
looks clean

mickel1138 (approved):
Double check to rights to the Anonymous image. Otherwise works!

PR #910, opened 2021-05-12
Updated Webpack args

mickel1138:
Because new Webpack can no longer sneak in extra argvs, so instead just using export env variables directly.


praneethpr (approved):
Tested on griidc-test with GRIIDC and GRP

fightingtexasaggie (approved):
tested, works...no errors.

PR #913, opened 2021-05-13
Feature/pelagos 4500 add submit info page for grp

praneethpr:


mickel1138:
It's being weird again, some of these files should already be in develop.

praneethpr:
> It's being weird again, some of these files should already be in develop.

Merged develop into this branch, i'm not sure if that is the correct way to do it

templates/nas-grp-base.html.twig
```diff
@@ -122,7 +130,13 @@
             </div>
         </div>
     </div>
-</div>
+
+    <!-- Footer -->
+    <div class="bootstrap">
+        <footer class="py-5 bg-dark fixed-bottom">
```
mickel1138, line 136: Remove py-5 and I think it's fine.

mickel1138, line 136: Just put the footer back to where it came from.

src/Controller/UI/DatasetSubmissionController.php
```diff
@@ -526,4 +527,21 @@ private function defaultDistributionPoint(DatasetSubmission $datasetSubmission,
...
+    /**
+     * Info page for Dataset Submission.
+     *
+     * @param string $customTemplate Custom base template from the environment variable.
+     *
+     * @Route("/dataset-submission/how-to-submit-data", name="pelagos_app_ui_datasetsubmission_info", methods={"GET"})
+     *
+     * @return Response A Response instance.
+     */
+    public function submissionInfoPage(string $customTemplate): Response
```
mickel1138, line 540: I don't think we need a flag for custom template, it can show anywhere.

config/services.yaml
```diff
@@ -189,3 +189,7 @@ services:
     App\MessageHandler\ZipDatasetFilesHandler:
         bind:
             $downloadDirectory: "%env(DOWNLOAD_PATH)%"
+
+    App\Controller\UI\DatasetSubmissionController:
+        bind:
+            string $customTemplate: "%env(CUSTOM_BASE_TEMPLATE)%"
```
mickel1138, line 195: Remove, see comment where this is used.

mickel1138 (approved):
Looks good now.

PR #917, opened 2021-05-19
Feature/pelagos 4506 fix line breaks in dataset title field

praneethpr:


fightingtexasaggie (approved):
This works great.
AS to existing data:
select id from dataset_submission where title like '%'||chr(10)||'%';
(25 dataset submissions have a line break in title, however, none are current submissions)

I think this is right for limiting search to only current submissions:
select udi from dataset left join dataset_submission on dataset.dataset_submission_id = dataset_submission.id WHERE dataset_submission.title like '%'||chr(10)||'%';

I also found no chr(13) (carriage returns) anywhere.

PR #918, opened 2021-05-20
Feature/pelagos 4507 add restriction to download api

praneethpr:


mickel1138:
We also have to do this for the ZIP download, and Filemanager API calls.

fightingtexasaggie (approved):
I really think we should gray out the download icon on dataland for restricted datasets, or otherwise visually indicate it's been disabled, but that's an improvement, and kinda unrelated to this ticket.

Verified by making dataset restricted after letting dataland load. Controller protects properly.

PR #927, opened 2021-06-02
Feature/pelagos 4509 create basic layout for new dataland

praneethpr:


assets/js/vue/pages/Dataland.vue
```diff
@@ -0,0 +1,162 @@
...
+                        </div>
+                    </div>
+                    <div>
+                        <div v-if="dataset.availabilityStatus !== 0 && dataset.doi.doi">
+                            <strong>DOI:</strong> {{ dataset.doi.doi }}
+                        </div>
+                        <div>
+                            <strong>UDI:</strong> {{ dataset.udi }}
+                        </div>
+                        <div v-if="dataset.availabilityStatus !== 7 && dataset.datasetSubmission.fileset">
```
mickel1138, line 27: I don't like having so much logic and these pages. Also can we translate the numbers to constants?
Even with Twig we would have logic. Maybe we should retool the API, and not give values for DIVs we don't want to show, then just check if values exist, then show or not show div.

Thoughts?

assets/js/vue/pages/Dataland.vue
```diff
@@ -0,0 +1,162 @@
...
+                    </p>
+                </div>
+                <hr>
+                    <!--PUBLICATIONS -->
+<!--                <div>-->
+<!--                    -->
+<!--                </div>-->
+                <div class="py-2" v-show="isFileMangerReady">
+                    <h2>{{dataset.datasetSubmission.id }}</h2>
+                    <FileManager :datasetSubId="16799" :writeMode="false"></FileManager>
```
mickel1138, line 57: Hard coded?

PR #933, opened 2021-06-14
Added download popup with cancel download option

praneethpr:


assets/js/vue/FileManager.vue
```diff
@@ -453,19 +501,28 @@ const renameItem = (item, name) => {
...
+                    `${Routing.generate('pelagos_api_file_download')}/${response.data.id}`, config
                 ).then((response) => {
                     const url = window.URL.createObjectURL(new Blob([response.data]));
                     const link = document.createElement('a');
                     link.href = url;
                     link.setAttribute('download', getFileNameFromHeader(response.headers));
                     document.body.appendChild(link);
                     link.click();
+                    myFileManager.$parent.resetDownloadAttrs();
+
```
mickel1138, line 525: extra line

fightingtexasaggie (commented):
Under both Firefox and Google Chrome (both on linux), I'm unable to see any progress bar, besides the browser's normal downloads location.

fightingtexasaggie:
I got a low memory warning from xymon griidc-dev when downloading. We'll have to make sure we're not copying large files into memory otherwise this won't scale, but it could have been a coincidence, so further investigation needed.

fightingtexasaggie:
If I cancel the download and attempt another download, it fails with a type error on the console, pasted below. Also, after I cancel the refresh button yields a "The directory cannot be opened" error.

Uncaught (in promise) TypeError: error.response is undefined
    downloadItems FileManager.vue:529

fightingtexasaggie (approved):
tested and works, the only thing I see is that when a user cancels downloads then immediately re-downloads, the progress bar sorta behaves weird, but the download is successful.

PR #934, opened 2021-06-16
Feature/pelagos 4513 create map view vue component

mickel1138:


templates/Dataland/v2/index.html.twig
```diff
@@ -0,0 +1,111 @@
...
+                    <hr>
+                    <div class="py-2">
+                        <h6><strong>Abstract:</strong></h6>
+                        <p>
+                            {{ dataset.abstract }}
+                        </p>
+                    </div>
+                    <hr>
+                    <div class="py-2">
+                        <div id="dlolmap" style="width: 100%;height:480px;"
```
fightingtexasaggie, line 57: I see this hardcoded here, yet when testing see different values. I'm assuming something overwrites it, but where/what?

fightingtexasaggie (dismissed):
Diffed against 4517 for review. Tested for both with and without geometry cases. I had one question I commented on about hardcoding the map size, but it's minor.

praneethpr (approved):
Tested and works

PR #935, opened 2021-06-17
Added Symfony command to check user status.

fightingtexasaggie:
I needed a clean way to check user status anyway for a DMS ticket, and this may be handy to have available later, so adding command. Alternatively, we can also just keep it in the one-off's misc repo if preferred.

### DIFF
--- a/src/Command/CheckUserCommand.php
+++ b/src/Command/CheckUserCommand.php
@@ -11,9 +11,6 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 use App\Entity\Account;
-use App\Entity\Person;
-#use App\Entity\PersonResearchGroup;
-#use App\Entity\ResearchGroup;
 
 /**
  * Command to convert a user into a Pelagos DRPM.
--- a/src/Command/CheckUserCommand.php
+++ b/src/Command/CheckUserCommand.php
@@ -13,7 +13,7 @@ use Symfony\Component\Console\Style\SymfonyStyle;
 use App\Entity\Account;
 
 /**
- * Command to convert a user into a Pelagos DRPM.
+ * Command to display a user's active/inactive status.
  */
 class CheckUserCommand extends Command
 {
@@ -64,7 +64,7 @@ class CheckUserCommand extends Command
      *
      * @throws \Exception If username not found.
      *
-     * @return void
+     * @return int Return code.
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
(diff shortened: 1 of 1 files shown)


src/Command/CheckUserCommand.php
```diff
@@ -0,0 +1,98 @@
...
+     * Symfony command execute section.
+     *
+     * @param InputInterface  $input  Required by Command.
+     * @param OutputInterface $output Required by Command.
+     *
+     * @throws \Exception If username not found.
+     *
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
```
praneethpr, line 69: Needs to return 0 (for no issues) or 1 (if there are any). Most of the old commands return void, I don't know why we did that, but it looks like doesn't break anything though

https://symfony.com/doc/4.4/console.html#creating-a-command

src/Command/CheckUserCommand.php
```diff
@@ -0,0 +1,98 @@
...
+            $activeUser = false;
+            $person = $account->getPerson();
+            $researchGroups = $person->getResearchGroups();
+            foreach ($researchGroups as $researchGroup) {
+                if (false === $researchGroup->isLocked()) {
+                    $rgList[] = $researchGroup->getShortName();
+                    $activeUser = true;
+                }
+            }
+            if (true === $activeUser) {
```
praneethpr, line 89:
```suggestion
            if ($activeUser) {
```

src/Command/CheckUserCommand.php
```diff
@@ -0,0 +1,98 @@
...
+        $account = $accountRepository->findOneBy(
+            ['userId' => $username]
+        );
+
+        if ($account instanceof Account) {
+            $activeUser = false;
+            $person = $account->getPerson();
+            $researchGroups = $person->getResearchGroups();
+            foreach ($researchGroups as $researchGroup) {
+                if (false === $researchGroup->isLocked()) {
```
praneethpr, line 84:
```suggestion
                if (!$researchGroup->isLocked()) {
```

src/Command/CheckUserCommand.php
```diff
@@ -0,0 +1,98 @@
...
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+use App\Entity\Account;
+
+/**
+ * Command to convert a user into a Pelagos DRPM.
```
praneethpr, line 16: Change comment description

src/Command/CheckUserCommand.php
```diff
@@ -0,0 +1,98 @@
...
+            ['userId' => $username]
+        );
+
+        if ($account instanceof Account) {
+            $activeUser = false;
+            $person = $account->getPerson();
+            $researchGroups = $person->getResearchGroups();
+            foreach ($researchGroups as $researchGroup) {
+                if (false === $researchGroup->isLocked()) {
+                    $rgList[] = $researchGroup->getShortName();
```
praneethpr, line 85: Can declare $rgList as empty array at the top

src/Command/CheckUserCommand.php
fightingtexasaggie: oops, missed that. thx

src/Command/CheckUserCommand.php
fightingtexasaggie: I'll do what Mike had started doing, with the @return int Return code.

praneethpr (approved):
LGTM!

PR #936, opened 2021-06-21
Additional changes based on PO comments

praneethpr:


fightingtexasaggie (commented):
Looks good, just 2 minor issues:
I asked and the PO wants number of files in the files section, as well as number of downloads we already have. Also, the font in the title is "Merriweather", Arial, serif, and in search it is "Lato", Arial, sans-serif, so they look a bit different.

Also, I added a commit moving the inline SVG to a file. I would have done it on here as a "suggestion" but had to add the .svg file to make it work.

fightingtexasaggie (approved):
Looks good to go!

PR #943, opened 2021-06-29
Feature/add eslint

praneethpr:
Please run 
`yarn run eslint assets/js`


.travis.yml
```diff
@@ -53,3 +53,4 @@ script:
 notifications:
   email: false
   slack: griidc:JosH6istHStJ7cm24B8CC58F
+
```
praneethpr, line 56: Can discard change

assets/js/vue/mixins/themeFunctions.js
```diff
@@ -1,7 +1,9 @@
...
-    }
-  }
-}
+    getThemeProperty(prop) {
+      return getComputedStyle(document.body).getPropertyValue(`--${prop}`);
+    },
+  },
+};
+
+export default themeFunctions;
```
mickel1138, line 9: This causes a build error. I don't think you can export default and the function. My suggestion makes it work.
```suggestion
export {themeFunctions};
```

mickel1138: Fixed now with additional configuration.

.github/workflows/eslint.yml
```diff
@@ -0,0 +1,32 @@
...
+          php-version: '7.4'
+
+      - name: Install dependencies
+        run: composer install --no-progress -q
+
+      - name: Install node modules
+        run: yarn install --silent
+
+      - name: Run ESLint
+        run: yarn run eslint
```
mickel1138, line 29: We can add `--quiet` so warnings don't show up when running the action/check.
```suggestion
        run: yarn run eslint --quiet
```

.github/workflows/eslint.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: composer install --no-progress -q
+
+      - name: Install node modules
+        run: yarn install --silent
+
+      - name: Run ESLint
+        run: yarn run eslint --quiet
+
+      - name: Webpack Build
+        run: yarn run encore prod
```
mickel1138, line 32: This could just be yarn (run) build, just like you used the script alias above.
```suggestion
        run: yarn run build
```

.github/workflows/eslint.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup PHP with PECL extension
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+
```
mickel1138, line 21: Shall we add composer version, so we can see which version it was tested on?

Same with phpcs, and eslint too I guess? Wadda ya think.

package.json
```diff
@@ -21,7 +26,8 @@
         "dev-server": "encore dev-server",
         "dev": "encore dev",
         "watch": "encore dev --watch",
-        "build": "encore production --progress"
+        "build": "encore production --progress",
+        "eslint": "eslint assets/js --ext .js,.vue"
```
mickel1138, line 30: Maybe the alias command should be just "lint", this was a bit confusing, because you can run eslint by using `yarn eslint` aswell, and maybe I just want to use eslint, and run a linting process.
```suggestion
        "lint": "eslint assets/js --ext .js,.vue"
```

.github/workflows/eslint.yml
mickel1138: We probably don't need it, we can get the installed software on the image read me.

.github/workflows/eslint.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Install dependencies
+        run: composer install --no-progress -q
+
+      - name: Install node modules
+        run: yarn install --silent
+
+      - name: Run ESLint
+        run: yarn run eslint --quiet
```
mickel1138, line 29:
```suggestion
        run: yarn run lint --quiet
```

mickel1138 (dismissed):
Test relevant parts, every works. And builds.

mickel1138:
I would also add a `yarn install --force` to refresh the yarn.lock file. (After development is merged.)

mickel1138 (approved):

PR #952, opened 2021-07-11
Feature/pelagos 4536 add custom template for hri

praneethpr:


assets/js/vue/components/search/Facet.vue
```diff
@@ -10,7 +10,7 @@
                 <div class="input-group pb-3" v-show="['researchGroup', 'projectDirector'].includes(facetName.queryParam)">
                     <input class="form-control" placeholder="Search" type="text" v-model="facetSearch">
                     <div class="input-group-append">
-                        <button class="btn btn-primary" type="button">
+                        <button class="btn btn-alternate" type="button">
```
mickel1138, line 13: This will also change the color for the other templates, are "we" ok with that?

mickel1138 (approved):
LGTM!

assets/js/vue/components/search/Facet.vue
mickel1138: It appears they are the same colors, so should be ok.

praneethpr: Yes, it should be because for other templates we have primary and alternate as the same color

PR #955, opened 2021-07-20
Feature/pelagos 4540 add globus tab on submission

praneethpr:


fightingtexasaggie (approved):
This looks good, but I really think we need to somehow communicate the mutually-exclusive nature of the tabs, maybe graying out the unselected tabs or with some helptext to make it clear that the "last one wins".

src/Form/DatasetSubmissionType.php
```diff
@@ -282,6 +282,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     ])
                 ]
             ))
+            ->add('datasetFilePath', Type\TextType::class, array(
```
mickel1138, line 285: I'm not sure how we are going to do this, I think the file manager might be here again.

praneethpr: I'm assuming the global ingest would be a folder path. On submission form POST, we get the path, populate fileset and file objects and then the filer would process the files. Instead of getting it from the upload folder, it would get from the globus ingest

praneethpr:
> This looks good, but I really think we need to somehow communicate the mutually-exclusive nature of the tabs, maybe graying out the unselected tabs or with some helptext to make it clear that the "last one wins".

Yes, we can communicate that to the POs and we can add some help text for that

PR #958, opened 2021-07-27
Feature/pelagos 4543 add home page for hri

praneethpr:


src/Controller/DefaultController.php
```diff
@@ -114,8 +71,39 @@ private function getProjectDirectorList(array $leadershipPersons): array
...
+
+            foreach ($fundingCycles as $fundingCycle) {
+                $fundingCycleList[] = array(
+                    'id' => $fundingCycle->getId(),
+                    'name' => $fundingCycle->getName(),
+                    'researchGroups' => $this->getResearchGroupsArray($fundingCycle)
+                );
+            }
+
+            return $this->render("Default/$customTemplate", array(
```
mickel1138, line 101: I figured out why there was an error, if this template doesn't exist, it breaks and doesn't even go to the default.
We should again rethink the switching mechanism, but switching by filter is not the same as switching by name/site/template.

praneethpr: We can talk about this

fightingtexasaggie (commented):
Are we going to use a separate index on this subsite the same way we do for GRP? If not, we'd show all datasets on the map-search page.

praneethpr:
> Are we going to use a separate index on this subsite the same way we do for GRP? If not, we'd show all datasets on the map-search page.

Need a separate ticket for creating HRI search index

fightingtexasaggie (approved):
I'm good with this initial site, and have tested the components (dif/submission/review/dataland/stats/monitoring). We still have to decide on the index, but as Praneeth said, that should be a separate ticket.

PR #959, opened 2021-07-27
Added authentication check for form submit post

praneethpr:


mickel1138:
I think this is a very good first step to solving this problem. It may even all we need.

mickel1138 (approved):

fightingtexasaggie (approved):
Tested by killing my logged-in session after loading submission/review. The re-direct works as desired.

PR #960, opened 2021-07-29
Redirect back to the submission/review after login

praneethpr:


fightingtexasaggie (commented):
If logged out in dataset review, the error re-direct is never reached, instead a "This page is only accessible by Data Repository Managers." message is displayed.

fightingtexasaggie (approved):
working

PR #962, opened 2021-08-05
Modified cold storage command

praneethpr:


fightingtexasaggie:
This now works. The workflow (which I'll document in a wiki) would involve first using dataset-review to delete any existing files, and to insert the 3 manifest files. After this is done, accept then use this command to flag as cold stored. ex:
bin/console pelagos:dataset-flag-coldstorage --udi R6.x824.000:0004 --originalfilesize 1000000000 --originalfilehash  3884d94f281f62ef40f1a41f2cdc1e267af47c6c4c70174688b6146e09fc8d2e --originalfilename bigfile.hdf

src/Command/ColdStorageFlagCommand.php
```diff
@@ -66,8 +57,12 @@ public function __construct(EntityManagerInterface $entityManager, MessageBusInt
     protected function configure()
     {
         $this
-            ->setDescription('Marks dataset in input file is cold-stored and updates datafile')
-            ->addArgument('infofile', InputArgument::REQUIRED, 'Filename of the coldinfo-formatted textfile required.');
+            ->setDescription('Marks specified dataset as cold-stored and updates fileset with provided files')
+            ->addOption('udi', null, InputOption::VALUE_REQUIRED, 'UDI of dataset to flag as cold stored')
+            ->addOption('originalfilesize', null, InputOption::VALUE_REQUIRED, 'Original file size')
+            ->addOption('originalfilehash', null, InputOption::VALUE_REQUIRED, 'Original sha256 hash')
+            ->addOption('originalfilename', null, InputOption::VALUE_REQUIRED, 'Original file name')
```
mickel1138, line 64: No aliases for these?

src/Command/ColdStorageFlagCommand.php
```diff
@@ -78,78 +73,45 @@ protected function configure()
...
+        $hash = $input->getOption('originalfilehash');
+        $originalFilename = $input->getOption('originalfilename');
+
+        $io->note("UDI: ($udi)");
+        $io->note("Original Size: ($size)");
+        $io->note("Original Hash: ($hash)");
+        $io->note("Original Filename: ($originalFilename)");
+        $io->note("Attempting to flag $udi as Cold Stored.");
+
+        $datasets = $this->entityManager->getRepository(Dataset::class)->findBy(array('udi' => $udi));
```
mickel1138, line 95: Here you can use `findOneBy` to just return only one.

src/Command/ColdStorageFlagCommand.php
```diff
@@ -78,78 +73,45 @@ protected function configure()
...
-                    $this->messageBus->dispatch($datasetSubmissionFilerMessage);
-                }
-            } else {
-                throw new \Exception("Could not open $stubFileName, expected to be at same location as $infoFileName.");
-            }
+        $io = new SymfonyStyle($input, $output);
+        $udi = $input->getOption('udi');
+        $size = $input->getOption('originalfilesize');
+        $hash = $input->getOption('originalfilehash');
+        $originalFilename = $input->getOption('originalfilename');
```
mickel1138, line 87: I would keep the variables somewhat consistent, just like `originalFilename` (besides caps) it's the same everywhere.

src/Command/ColdStorageFlagCommand.php
fightingtexasaggie: nah, it was a bit more confusing with them as I couldn't really make them meaningful and 1-character.

src/Command/ColdStorageFlagCommand.php
```diff
@@ -78,78 +73,44 @@ protected function configure()
...
+        } else {
+            $io->note('Submission Found.');
+
+            $datasetSubmission->setModifier($systemPerson);
+            $datasetSubmission->setDatasetFileColdStorageAttributes($originalFileSize, $originalFileHash, $originalFileName);
+
+            $this->entityManager->persist($datasetSubmission);
+            $this->entityManager->flush();
+        }
         return 0;
```
mickel1138, line 114: I think we could with with a success message here: (It was a bit confusing, this was it really says it's done.
```suggestion

        $io->success('Done!');
        
        return 0;
```

src/Command/ColdStorageFlagCommand.php
```diff
@@ -78,78 +73,44 @@ protected function configure()
...
+        $datasetSubmission = $dataset->getDatasetSubmission();
+        if (!($datasetSubmission instanceof DatasetSubmission)) {
+            throw new \Exception('Could not find Dataset Submission.');
+        } else {
+            $io->note('Submission Found.');
+
+            $datasetSubmission->setModifier($systemPerson);
+            $datasetSubmission->setDatasetFileColdStorageAttributes($originalFileSize, $originalFileHash, $originalFileName);
+
+            $this->entityManager->persist($datasetSubmission);
```
praneethpr, line 111: Can remove persist() here, not required if it is not a new object

src/Command/ColdStorageFlagCommand.php
```diff
@@ -78,78 +73,44 @@ protected function configure()
...
+        $io->note("Original Hash: ($originalFileHash)");
+        $io->note("Original File Name: ($originalFileName)");
+        $io->note("Attempting to flag $udi as Cold Stored.");
+
+        $dataset = $this->entityManager->getRepository(Dataset::class)->findOneBy(array('udi' => $udi));
+        if ($dataset instanceof Dataset) {
+            $io->note('Dataset Found.');
         } else {
-            throw new \Exception("Error: Could not open $infoFileName.");
+            throw new \Exception('Could not find a dataset with the udi provided.');
```
mickel1138, line 99: As an alternative you could use std error output, and return 1. Because it's not really an exception, just bad parameters.
```suggestion
            $io->error('Could not find a dataset with the udi provided.');
            return 1;
```

mickel1138 (approved):
Testing, and works as expected.

praneethpr (commented):
Looks good to me

PR #964, opened 2021-08-06
Feature/pelagos 4564 handle api authentication failures

mickel1138:


praneethpr (approved):
Tested, very good solution!

PR #965, opened 2021-08-09
Added popup error message when authentication fails

praneethpr:


assets/js/vue/FileManager.vue
```diff
@@ -354,6 +354,11 @@ const initDropzone = () => {
     maxFilesize: null,
     clickable: '#upload-file-button',
     timeout: 0,
+    error: function error(file, message) {
```
mickel1138, line 357: The third parameter is `xhr` you can get the Status from it, that may be better than checking the message.code, because if something else went wrong, you might not get a message, and then BOOM!

assets/js/vue/FileManager.vue
```diff
@@ -354,6 +354,11 @@ const initDropzone = () => {
...
+      if (message.code === 401) {
```
mickel1138, line 358:
```suggestion
    error: function error(file, message, xhr) {
      if (xhr.status === 401) {
```

assets/js/vue/FileManager.vue
```diff
@@ -354,6 +354,11 @@ const initDropzone = () => {
...
+        myFileManager.$parent.showPopupError(message.message);
```
mickel1138, line 359: We should probably do a bit more than just show the error message. It'll be confusing for the user.
So maybe instead it should be a message like: "You are no (longer) logged in!" Click OK (button) to login. Or something.
May have to discuss this with the PO.

praneethpr: I think the message should be set in API? Regarding the button, i'm not sure whether we need it because this is a edge case scenario

mickel1138: Edge case maybe, but when someone receives this message, they don't know what to do. Also the message is a Symfony error message, but we can change it.

If this happens, the only thing a user should be able/can do, is go back to to the login screen, and I think we should help them with that.

src/Security/LoginFormAuthenticator.php
```diff
@@ -333,7 +333,7 @@ public function start(Request $request, AuthenticationException $authException =
     {
         // Is this a JSON request?
         if (false !== strpos($request->getRequestFormat(), 'json')) {
-            throw new HttpException(Response::HTTP_UNAUTHORIZED, $authException->getMessage());
+            throw new HttpException(Response::HTTP_UNAUTHORIZED, 'Session expired! Please log in again!');
```
mickel1138, line 336: This is a fine message. I think we should forward them to the login page, but maybe to serve this to the PO this way and see what they say.

assets/js/vue/FileManager.vue
fightingtexasaggie: It's going to be a super edge case once we adjust the parameters back, and extend the garbage collection startup from 8 hrs to 24 hrs.

fightingtexasaggie (approved):
I'm ok with this, pending the PO is as it'll likely only very rarely (if ever) be encountered.

mickel1138 (approved):
Tested, and works fine.

PR #967, opened 2021-08-10
Feature/pelagos 4567 add token authentication guard

mickel1138:


src/Security/PersonTokenAuthenticator.php
```diff
@@ -144,4 +118,43 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
...
+     */
+    public function start(Request $request, AuthenticationException $authException = null)
+    {
+        throw new \Exception('This should not be reached, token?');
+    }
+
+    /**
+     * Remember me is not supported.
+     */
+    public function supportsRememberMe()
```
praneethpr, line 156: Missing return type

src/Security/PersonTokenAuthenticator.php
```diff
@@ -144,4 +118,43 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
...
+     * @param TokenInterface $token       A Symfony user token, req by interface.
+     * @param string         $providerKey The name of the used firewall key.
+     *
+     * @return Response The response or null to continue request.
+     */
+    // Next line to be ignored because implemented function does not have type-hint on $providerKey.
+    // phpcs:ignore
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
+    {
+        return;
```
praneethpr, line 135: " If you return null, the current request will continue, and the user
will be authenticated. This makes sense, for example, with an API" 

Above comment is from the AuthenticatorInterface, so i think we can return a null response?

src/Security/PersonTokenAuthenticator.php
```diff
@@ -144,4 +118,43 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
...
+     *
+     * @param Request        $request     A Symfony Request, req by interface.
+     * @param TokenInterface $token       A Symfony user token, req by interface.
+     * @param string         $providerKey The name of the used firewall key.
+     *
+     * @return Response The response or null to continue request.
+     */
+    // Next line to be ignored because implemented function does not have type-hint on $providerKey.
+    // phpcs:ignore
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
```
praneethpr, line 133: Does it break if it is type "string"?

src/Security/PersonTokenAuthenticator.php
```diff
@@ -39,82 +39,56 @@ public function __construct(\Twig_Environment $twig)
...
+     * @param string        $credentials Credentials Array.
+     * @param UserInterface $user        The user.
      *
-     * @return boolean Whether the token is supported.
+     * @return boolean True if the credentials are valid.
      */
-    // Next line to be ignored because implemented function does not have type-hint on $providerKey.
-    // phpcs:ignore
-    public function supportsToken(TokenInterface $token, $providerKey)
+    public function checkCredentials($credentials, UserInterface $user)
```
praneethpr, line 88: Missing return type

src/Security/PersonTokenAuthenticator.php
```diff
@@ -39,82 +39,56 @@ public function __construct(\Twig_Environment $twig)
...
-     *                                  userProvider that is not a PersonTokenUserProvider.
+     * @throws AuthenticationException When the token is invalid or expired.
      *
-     * @return PreAuthenticatedToken An authenticated Symfony authentication token.
+     * @return UserInterface Return the user.
      */
-    // Next line to be ignored because implemented function does not have type-hint on $providerKey.
-    // phpcs:ignore
-    public function authenticateToken(TokenInterface $token, UserProviderInterface $userProvider, $providerKey)
+    public function getUser($credentials, UserProviderInterface $userProvider)
```
praneethpr, line 75: Missing return type

src/Security/PersonTokenAuthenticator.php
```diff
@@ -39,82 +39,56 @@ public function __construct(\Twig_Environment $twig)
...
-     * @param string  $providerKey The provider key.
+     * @param Request $request A Request object.
      *
-     * @return PreAuthenticatedToken|null A new authentication token or null if token is not set.
+     * @return boolean True if person_token exists.
      */
-    // Next line to be ignored because implemented function does not have type-hint on $providerKey.
-    // phpcs:ignore
-    public function createToken(Request $request, $providerKey)
+    public function supports(Request $request)
```
praneethpr, line 48: Missing return type

src/Security/PersonTokenAuthenticator.php
mickel1138: Yes, it can be null. There is not additional checking needed. I the token is valid, it's done. But this function HAS to be implemented, so something has to be returned.

src/Security/PersonTokenAuthenticator.php
mickel1138: The prototype will fail, I think it's fixed in Symfony 5.

src/Security/PersonTokenAuthenticator.php
praneethpr: From the documentation,
`    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey): ?Response
    {
        // on success, let the request continue
        return null;
    }`

I think it is more elegant returning null and it would match the return type as well

mickel1138: In an example there was no return statement at all, and it had this same comment. Can you make a code suggestion?

praneethpr: I got that example code from this 
https://symfony.com/doc/4.4/security/guard_authentication.html

praneethpr: I'm unable to make multi-line code suggestion, it only lets me do it for single line

praneethpr (approved):

fightingtexasaggie (approved):
tested password resets and verified identity was established via token (via email) instead of local auth. Verified changed password worked as expected.

PR #968, opened 2021-08-11
Alwys show errors.

mickel1138:


assets/static/js/dataset-submission.js
```diff
@@ -258,16 +258,14 @@ $(function() {
...
-                    var n = noty(
-                    {
-                        layout: "top",
-                        theme: "relax",
-                        type: "error",
-                        text: message,
-                        modal: true,
-                    });
-                }
+                var n = noty(
```
praneethpr, line 261: Shouldn't we check for response code like how we do it in javascript? For 401?

mickel1138: I figured at this level, any error should be shown, but maybe it should be just 40x errors.

praneethpr: I don't think we should show all errors, it might a db error or it could be any other generic exception when the form is submitted and this ticket is to deal with unauthorized users in dataset-submission

praneethpr (approved):
Did a test merge with develop, Tested and works

PR #969, opened 2021-08-11
Upgrade to node-sass 6 to mitigate security issues.

mickel1138:
We don't need node 14 or higher, but we should upgrade.
Also node-sass is depricated, so one day we should switch to dart-sass.


praneethpr (approved):
Tested, works fine

fightingtexasaggie (approved):

PR #970, opened 2021-08-12
Throw exception if file size does not match.

mickel1138:


src/Util/FileUploader.php
```diff
@@ -94,10 +96,16 @@ public function combineChunks(Request $request): array
...
         }
-        // Success
-        fclose($targetFile);
         rmdir($chunksFolder);
+        $targetFileSize = StreamInfo::getFileSize(array('fileStream' => $targetFile));
+        fclose($targetFile);
+        if ($targetFileSize !== $fileSize) {
+            unlink($targetFileName);
+            rmdir($targetDirectory);
+            throw new HttpException(Response::HTTP_BAD_REQUEST, "File size does not match!\n Expected file size is:$fileSize, and got $targetFileSize!");
```
praneethpr, line 105: This message is not getting displayed in PROD mode

mickel1138: This is going to be a problem, I'm going to look into this deeper. I cannot return a response here.

praneethpr:
```suggestion
            throw new BadRequestHttpException("File size does not match!\n Expected file size is:$fileSize, and got $targetFileSize!");
```

praneethpr: Maybe "Filesize doesn't match" is a good enough error message to the user

praneethpr (dismissed):

src/Util/FileUploader.php
```diff
@@ -94,10 +97,16 @@ public function combineChunks(Request $request): array
             fclose($chunk);
             unlink($chunksFolder . DIRECTORY_SEPARATOR . $i);
         }
-        // Success
-        fclose($targetFile);
         rmdir($chunksFolder);
+        $targetFileSize = StreamInfo::getFileSize(array('fileStream' => $targetFile));
+        fclose($targetFile);
+        if ($targetFileSize === $fileSize) {
```
mickel1138, line 103:
```suggestion
        if ($targetFileSize !== $fileSize) {
```

mickel1138: Only if they are NOT the same. ;-)

praneethpr (approved):

PR #972, opened 2021-08-16
Changed Http exception to json response type

praneethpr:


src/Security/LoginFormAuthenticator.php
```diff
@@ -322,18 +322,16 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
...
      *
-     * @return RedirectResponse
+     * @return Response
      */
     public function start(Request $request, AuthenticationException $authException = null)
     {
         // Is this a JSON request?
         if (false !== strpos($request->getRequestFormat(), 'json')) {
-            throw new HttpException(Response::HTTP_UNAUTHORIZED, 'Session expired! Please log in again!');
+            return new JsonResponse(['code' => 401, 'message' => 'Session expired! Please log in again'], 401);
```
mickel1138, line 334:
```suggestion
            return new JsonResponse(['code' => 401, 'message' => 'Session expired! Please log in again'], Response::HTTP_UNAUTHORIZED);
```

mickel1138: Use the constant instead of the number.

PR #973, opened 2021-08-19
Added authentication check for Datasetsubmission PUT API

praneethpr:


src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -182,6 +183,9 @@ public function postAction(Request $request)
      *     )
      * )
      *
+     * @IsGranted("IS_AUTHENTICATED_FULLY")
+     * @IsGranted("CAN_EDIT", subject="datasetSubmission")
```
mickel1138, line 187: I'm not sure if this `subject` is available, you may have to do it with that magic conversion method?

mickel1138: Actually, don't need this line at all, entityHandler checks that.

mickel1138 (approved):
This works always now.

fightingtexasaggie (approved):
works, athough there isn't a re-direct when clicking between tabs, just the noty, but the message is reasonable.

PR #975, opened 2021-08-20
Added DxPopup to submission

mickel1138:


assets/static/js/dataset-submission.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
mickel1138, line 1: I did this because VS studio showed everything is "red"

praneethpr: I have made my editor to ignore that folder, we could solve it by adding a .eslintignore file with "assets/static/js"

praneethpr (approved):
Tested and works

assets/static/js/dataset-submission.js
```diff
@@ -257,7 +280,9 @@ $(function() {
                 }
             },
             error: function(jqXHR, textStatus, errorThrown) {
-                if ([400, 401].includes(jqXHR.status)) {
```
fightingtexasaggie, line 260: extra logic?

mickel1138: It's not really needed now, but I left it, incase we want to add to this list or codes. But it could be changed to `401 === jqXHR.status` add a code suggestion if you feel it should be changed.

fightingtexasaggie (approved):
This works per the ticket, for cases of switching tabs / saving. There are 2 other similar cases though as the popup after hitting submit and the popup while uploading a file don't contain the login link either. I think maybe we should add them there too to be consistent. We should at least discuss.

praneethpr (approved):

PR #976, opened 2021-08-23
Added sample data to API to populate dropdown for folder list

praneethpr:


src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -404,4 +404,34 @@ public function validateUrlAction(int $id, Request $request, UrlValidation $urlV
...
+     *     defaults={"_format"="json"},
+     *     )
+     *
+     * @View()
+     *
+     * @return Response
+     */
+    public function getGlobalIngestFolders(Request $request): Response
+    {
+        $path = $request->get('user');
```
mickel1138, line 426: User name is not need for this request. We can get it from Symfony using: `$this->security->getUser();`
I don't think there would be a need to get this for another user by someone else.
So, then you can also get rid of the cookie reading function.

Technically we could do this in the Form, and prepopulate it there, but I kinda like the Javascrip approach, in case they are uploading on another screen or so. In that case, I would make sure the list is fetch on every time. Like on an `onChange` event.

assets/static/js/dataset-submission.js
```diff
@@ -228,6 +230,34 @@ $(function() {
...
+
+    function populateFolderDropDownList() {
+        let dropdown = $('#datasetFilePath');
+
+        dropdown.empty();
+
+        dropdown.append('<option selected="true" disabled>Choose Folder</option>');
+        dropdown.prop('selectedIndex', 0);
+
+        const url = Routing.generate("pelagos_api_get_folder_list_dataset_submission", { "user" : getCookie("GRIIDC_USERNAME")} );
```
fightingtexasaggie, line 251: Should we try and minimize (or phase out) the use of this cookie? I thought it was only for drupal, which at some point may be moved to another system altogether.

mickel1138: This cookie won't be needed, the code is being changed elsewhere for that.

praneethpr: I'll remove this, not required. I can get the user in the Controller like how @mickel1138  mentioned

src/Controller/Api/DatasetSubmissionController.php
praneethpr: Made the required changes

mickel1138:
While the drop down works fine, I am unable to submit it, it says I need a file or remotly hosted url.
Just add this new field to the validate rules.

mickel1138 (approved):
The dropdown part works, but the submission doesn't work yet.

fightingtexasaggie (approved):
looks good, but needs another approval now since I added a commit

templates/DatasetReview/file-submit.html.twig
```diff
@@ -48,7 +48,7 @@
                                     <img class="info" src="{{ asset('build/images/info32.png') }}" alt="info image">
                                     <div class="tooltiptext">
                                         <p><strong>Dataset File Path:</strong></p>
-                                        <p>This is the path on the server to the data file being submitted. Click "Browse..." to find and select the data file you uploaded via SFTP/Ingest.</p>
+                                        <p>Select the folder associated with the dataset you uploaded via SFTP/Globus. This list contains the folders directly under your 'incoming' folder. Deeper nested folders cannot be seleted.</p>
```
mickel1138, line 51:
```suggestion
                                        <p>Select the folder associated with the dataset you uploaded via SFTP/Globus. This list contains the folders directly under your 'incoming' folder.</p>
```

PR #979, opened 2021-08-24
Added button (DxToolBarItem) to popup, and show only for session expired

mickel1138:


praneethpr (dismissed):
Tested and works.

assets/js/vue/FileManager.vue
```diff
@@ -355,7 +364,9 @@ const initDropzone = () => {
     clickable: '#upload-file-button',
     timeout: 0,
     error: function error(file, errorMessage, xhr) {
-      if ([400, 401].includes(xhr.status)) {
+      if ([401].includes(xhr.status)) {
```
praneethpr, line 367: Move it to a common function so all the API errors are handled

PR #985, opened 2021-08-27
Feature/pelagos 4583 add api to get global ingest directories

fightingtexasaggie:
can test against griidc-ingest on dev/test, let me know if the node instance needs to be turned on, but I'll leave it on for now.


fightingtexasaggie:
not needed...on other branch

PR #986, opened 2021-08-27
Add Ingest util class.

fightingtexasaggie:
2nd time is the charm...  You should be able to test against the node service on griidc-ingest. If it's offline, let me know and I'll bring it up for testing. This PR removes the call to the stub and replaces it with a call to a nodeexpress server that generates the user's ingest directory listings.


src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -406,4 +409,26 @@ public function validateUrlAction(int $id, Request $request, UrlValidation $urlV
...
+     *     )
+     *
+     * @View()
+     *
+     * @return Response
+     */
+    public function getGlobalIngestFolders(IngestUtil $ingestUtil): Response
+    {
+        if (!($this->getUser() instanceof Account)) {
+            throw new AccessDeniedException('User is either not logged in or does not have an account');
```
mickel1138, line 430: You can't tell from this this if the user has an account or not. So here it just means that the user is not logged in. 
The text could also say something like "Must be logged in".
```suggestion
            throw new AccessDeniedException('User is either not logged in');
```

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param string $ingestApiUrl The API URL.
+     */
+    public function __construct(
+        string $ingestApiUrl
+    ) {
+        $this->url = $ingestApiUrl;
```
mickel1138, line 33: url is too generic, please name is insgestApiUrl as well. This is consistent with have we have done this in other places.

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,61 @@
...
+        $url = $this->url . '/' . $username;
+        try {
+            $response = $client->request('GET', $url);
+        } catch (ClientException $exception) {
+            throw new HttpClientErrorException($exception->getMessage(), $exception->getCode());
+        } catch (ServerException $exception) {
+            throw new HttpServerErrorException($exception->getMessage(), $exception->getCode());
+        }
+        $data = json_decode($response->getBody()->getContents(), true);
+        return $data;
```
mickel1138, line 59: No need to make the variable if you are going to return it right away.

config/services.yaml
```diff
@@ -99,6 +99,9 @@ services:
         $doiApiPassword: "%env(DOI_API_PASSWORD)%"
         $doiApiPrefix: "%env(DOI_API_PREFIX)%"
         $doiApiUrl: "%env(DOI_API_URL)%"
+    
+    App\Util\IngestUtil:
+        $ingestApiUrl: "%env(INGEST_API_URL)%"
```
mickel1138, line 104: Where is the in the `.env` file?

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,60 @@
...
+        $metadata = array();
+        $url = $this->ingestApiUrl . '/' . $username;
+        try {
+            $response = $client->request('GET', $url);
+        } catch (ClientException $exception) {
+            throw new HttpClientErrorException($exception->getMessage(), $exception->getCode());
+        } catch (ServerException $exception) {
+            throw new HttpServerErrorException($exception->getMessage(), $exception->getCode());
+        }
+        return(json_decode($response->getBody()->getContents(), true));
```
mickel1138, line 58: Normally we just `return someting` and not call it like a function.

mickel1138: From the PHP manual: 

> Note: Note that since return is a language construct and not a function, the parentheses surrounding its argument are not required and their use is discouraged.

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace App\Util;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\GuzzleException;
```
praneethpr, line 7: Not used

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,60 @@
...
+    public function __construct(
+        string $ingestApiUrl
+    ) {
+        $this->ingestApiUrl = $ingestApiUrl;
+    }
+
+    /**
+     * This function will get a listing of ingest folders.
+     *
+     * @param string $user Username to get ingest/incoming/* folder listing for.
```
praneethpr, line 39:
```suggestion
     * @param string $username Username to get ingest/incoming/* folder listing for.
```

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @throws HttpClientErrorException Exception thrown when response code is 4xx negotiating with Ingest server REST API.
+     * @throws HttpServerErrorException Exception thrown when response code is 5xx negotiating with Ingest server REST API.
+     *
+     * @return array
+     */
+    public function getUsersIngestFoldersInIncomingDir(string $username)
```
praneethpr, line 46:
```suggestion
    public function getUsersIngestFoldersInIncomingDir(string $username): array
```

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $client = new Client();
+        $metadata = array();
```
praneethpr, line 49: Not used

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -421,17 +424,12 @@ public function validateUrlAction(int $id, Request $request, UrlValidation $urlV
...
-            "H1.x123.456.7890",
-            "H1.x123.456:7891",
-            "R2.x137.108.0003",
-            "Spaces in foldername",
-            "c1.x123.456:7890"
-        ]);
+        if (!($this->getUser() instanceof Account)) {
+            throw new AccessDeniedException('Must be logged in');
+        }
+        $jsonResponse = $this->makeJsonResponse($ingestUtil->getUsersIngestFoldersInIncomingDir($this->getUser()->getUserID()));
```
praneethpr, line 432:
```suggestion
        return $this->makeJsonResponse($ingestUtil->getUsersIngestFoldersInIncomingDir($this->getUser()->getUserID()));
```

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,60 @@
...
+        $url = $this->ingestApiUrl . '/' . $username;
+        try {
+            $response = $client->request('GET', $url);
+        } catch (ClientException $exception) {
```
praneethpr, line 53: I don't think we need to do this, we can just throw the exception back. This was done in DOIUtil to differentiate between client exceptions (400's Bad requests., etc) vs server exceptions(when their service was done) in order to re-try after few minutes

src/Controller/Api/DatasetSubmissionController.php
mickel1138: For some reason i'm not seeing the earlier fix in the file changes, this is the cleanest looking I think .

src/Controller/Api/DatasetSubmissionController.php
```diff
@@ -421,17 +424,12 @@ public function validateUrlAction(int $id, Request $request, UrlValidation $urlV
...
-            "H1.x123.456:7891",
-            "R2.x137.108.0003",
-            "Spaces in foldername",
-            "c1.x123.456:7890"
-        ]);
+        if (!($this->getUser() instanceof Account)) {
+            throw new AccessDeniedException('Must be logged in');
+        }
+        return $this->makeJsonResponse($ingestUtil->getUsersIngestFoldersInIncomingDir($this->getUser()->getUserID()));
+        return $jsonResponse;
```
mickel1138, line 433: Two returns? How fancy!

src/Util/IngestUtil.php
```diff
@@ -0,0 +1,52 @@
...
+    public function getUsersIngestFoldersInIncomingDir(string $username): array
+    {
+        $client = new Client();
+        $url = $this->ingestApiUrl . '/' . $username;
+        try {
+            $response = $client->request('GET', $url);
+        } catch (\Exception $exception) {
+            throw new HttpException($exception->getCode(), $exception->getMessage());
+        } 
+        return(json_decode($response->getBody()->getContents(), true));
```
mickel1138, line 50: This is still here?

praneethpr (approved):

PR #993, opened 2021-09-07
Also detect server down, and show all server error messages.

mickel1138:


fightingtexasaggie (approved):
Tested and works, however, do we need something similar in review?

assets/static/js/dataset-submission.js
```diff
@@ -281,8 +281,13 @@ $(function() {
             error: function(jqXHR, textStatus, errorThrown) {
                 if ([401].includes(jqXHR.status)) {
                     loggedOutPopup.show();
-                } else if ([400].includes(jqXHR.status)) {
-                  var message = jqXHR.responseJSON == null ? errorThrown: jqXHR.responseJSON.message;
+                } else {
+                    if (jqXHR.status != 0) {
+                        var message = jqXHR.responseJSON == null ? errorThrown: jqXHR.responseJSON.message;
```
praneethpr, line 286: variable initialization can be moved out of the if-else block, and we can use let instead

mickel1138:
> Tested and works, however, do we need something similar in review?

Yes maybe, but reviewers know about the maintenance window. And should not be on at those times.

assets/static/js/dataset-submission.js
```diff
@@ -279,23 +279,22 @@ $(function() {
...
             error: function(jqXHR, textStatus, errorThrown) {
+                let message = "Server is Unreachable, please try again later!";
                 if ([401].includes(jqXHR.status)) {
                     loggedOutPopup.show();
                 } else {
                     if (jqXHR.status != 0) {
-                        var message = jqXHR.responseJSON == null ? errorThrown: jqXHR.responseJSON.message;
-                    } else {
-                        var message = "Server is Unreachable, please try again later!";
+                        message = jqXHR.responseJSON == null ? errorThrown: jqXHR.responseJSON.message;
```
fightingtexasaggie, line 287: shouldn't this be jqXHR.responseJSON != null ?

fightingtexasaggie (commented):
While this did eventually work, it took quite a long time before I saw the noty, sometime between 30 and 60 seconds (varied) Is there a way to reduce the timeout, or should we attempt? 401 errors are instant, as the server is there to respond, but if we lose connectivity to the server (i.e. shut off apache for maintenance), it takes quite a while.

long-time

mickel1138:
> While this did eventually work, it took quite a long time before I saw the noty, sometime between 30 and 60 seconds (varied) Is there a way to reduce the timeout, or should we attempt? 401 errors are instant, as the server is there to respond, but if we lose connectivity to the server (i.e. shut off apache for maintenance), it takes quite a while.
> 
> long-time

There is a timeout, I agree this in not very user friendly, I'm going to try something else.

PR #994, opened 2021-09-07
Corrected logic for granted for admin only

mickel1138:


praneethpr (approved):

fightingtexasaggie (approved):
tested as DRPM, and not a DRPM...worked as expected.

PR #998, opened 2021-09-08
Added validation check if remotely hosted value exists

praneethpr:


fightingtexasaggie (approved):
tested this out and works as expected. There is a quirk if you fill the form, blank the value, let it error, then fill the form again with the same value it blocked submit but still had the valid class applied. I only saw this once though, and could not get it to repeat so it was probably just my test environment, but you way want to double-check.

mickel1138 (approved):
Tested, works.

PR #999, opened 2021-09-08
Added basis for maintenance window detection.

fightingtexasaggie:
We can do something like this to trigger a noty (or maybe full screen modal) as our maintence window rolls around on Tuesday at 3:00 pm central. There's still some timezone fun to be had, but you get the idea.


mickel1138:
I feel this is a bad idea.

fightingtexasaggie:
closing...there are better approaches to consider.

PR #1002, opened 2021-09-13
Feature/pelagos 4604 show server errors for submission

mickel1138:
Different approach from PR #993

### DIFF
--- a/assets/static/js/dataset-submission.js
+++ b/assets/static/js/dataset-submission.js
@@ -317,7 +317,6 @@ $(function() {
         }).always(function(){
             dsNoty.close();
         });
-
     }
 
     $("#temporalExtentBeginPosition").datepicker({
--- a/assets/static/js/dataset-submission.js
+++ b/assets/static/js/dataset-submission.js
@@ -233,20 +233,43 @@ $(function() {
         closeOnOutsideClick: false,
         showCloseButton: false,
         toolbarItems: [{
-          widget: "dxButton",
-          toolbar: "bottom",
-          options: {
+            widget: "dxButton",
+            toolbar: "bottom",
+            options: {
             type: "danger",
             text: "Continue to Login Form",
             onClick: function(e) {
                 window.location.href = Routing.generate("security_login", {"destination":window.location.href})
               }
             },
-          }]
-      }).dxPopup("instance");
+        }]
+    }).dxPopup("instance");
+
+    const errorPopup = $("#errorPopup").dxPopup({
+        width: 300,
+        height: 250,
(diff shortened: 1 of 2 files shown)


assets/static/js/dataset-submission.js
```diff
@@ -256,6 +279,7 @@ $(function() {
             url: url + "/" + datasetSubmissionId + "?validate=false",
             method: "PUT",
             data: formData,
+            timeout: 30000,
```
fightingtexasaggie, line 282: For these PUT operations, I think 30 seconds is really generous. I'd think what doesn't finish in 5 seconds will likely not finish in 30.  Maybe we should shorten?

fightingtexasaggie (commented):
currently the auto-save when switching between tabs is no longer working, and also the hidden divs with the error messages are being exposed...see screenshot.

Screenshot from 2021-09-13 15-26-22

assets/static/js/dataset-submission.js
mickel1138: It's less than the 60 seconds before, but less than 30 would be better, not sure what the magic number would be.

mickel1138:
> currently the auto-save when switching between tabs is no longer working, and also the hidden divs with the error messages are being exposed...see screenshot.
> 
Good catch, I will get rid of it. The auto save should work, but i'll check the logic.

mickel1138:
> > currently the auto-save when switching between tabs is no longer working, and also the hidden divs with the error messages are being exposed...see screenshot.
> 
> Good catch, I will get rid of it. The auto save should work, but i'll check the logic.

Ok, I double checked, and the DIVs don't show, this may be a integrity check/cache issue.

fightingtexasaggie (dismissed):
This works. Maybe a later improvement might be to display a "saving your information..." along with a "done" noty where they close maybe a second after success. If  the server is down, the user will still see the attempt is being made, with this error dialog after 30 sec.

praneethpr (approved):
Tested and works

assets/static/js/dataset-submission.js
```diff
@@ -279,22 +303,22 @@ $(function() {
...
-                  var n = noty(
-                  {
-                      layout: "top",
-                      theme: "relax",
-                      type: "error",
-                      text: message,
-                      modal: true,
-                  });
+                } else {
+                    if (jqXHR.status != 0) {
```
praneethpr, line 310: Add type comparison

praneethpr (approved):

fightingtexasaggie (approved):

PR #1003, opened 2021-09-15
Bugfix/pelagos 4602 add valid check to remotely hosted field

praneethpr:


mickel1138:
Request revisions fails, this needs the same fix as applied in the Large File Upload epic, where the fileset copy is moved.

mickel1138:
I couldn't get this to work on submission, after much looking, I got it to work like this, see commit: 62107ab

assets/static/js/dataset-submission.js
```diff
@@ -144,18 +156,25 @@ $(function() {
...
-                $("#regForm").prop("unsavedChanges", false);
-                form.submit();
-            }
+            formHash = $("#regForm").serialize();
+            $("#regForm").prop("unsavedChanges", false);
+            form.submit();
         },
     });
 
+    $.validator.addClassRules("files", {
```
praneethpr, line 174: I think we should remove this

assets/static/js/dataset-submission.js
```diff
@@ -358,6 +381,19 @@ $(function() {
...
+
+            $("#filesUploaded").rules("add", {
+                require_from_group: [1,".files"]
+            });
+
+            $("#remotelyHostedUrl").rules("add", {
+                require_from_group: [1,".files"]
+            });
+
+            // $.validator.addClassRules("files", {
```
praneethpr, line 394: Can remove this uncommented code

assets/js/vue/FileManager.vue
```diff
@@ -221,16 +221,11 @@ const getItems = (pathInfo) => new Promise((resolve, reject) => {
     `${Routing.generate('pelagos_api_get_files_dataset_submission')}/${datasetSubmissionId}?path=${pathInfo.path}`,
   ).then((response) => {
     resolve(response.data);
-    const filesTabValidator = document.getElementById('filesTabValidator');
+    const filesUploaded = document.getElementById('filesUploaded');
     const datasetFileTransferType = document.getElementById('datasetFileTransferType');
     if (response.data.length > 0) {
-      filesTabValidator.value = 'valid';
+      filesUploaded.value = 'valid';
       datasetFileTransferType.value = 'upload';
```
praneethpr, line 228: We need to add the else case here, because when we upload a file it sets it to "valid" and then we delete it, it is still valid and you can submit the form without any files. I know this is an edge case, but we still need to set filesUploaded.value to null or empty string if response.data.length is 0

assets/static/js/datasetReview.js
```diff
@@ -44,15 +44,38 @@ $(document).ready(function(){
 
     var remoteURL = Routing.generate("pelagos_api_dataset_submission_validate_url", { id: $("form[datasetsubmission]").attr("datasetsubmission") });
 
+    if ($("#remotelyHostedUrl").val()) {
+        $("#datasetFileTransferType").val("HTTP");
+    }
+
+    $("#remotelyHostedUrl, #files-uploaded").on("keyup change", function() {
```
praneethpr, line 51:
```suggestion
    $("#remotelyHostedUrl, #filesUploaded").on("keyup change", function() {
```

templates/DatasetSubmission/file-submit.html.twig
```diff
@@ -15,12 +15,12 @@
                 <li datasetFileTransferType="upload"><a href="#tabs-1">Upload</a></li>
                 <li datasetFileTransferType="HTTP"><a href="#tabs-2">Enter URL</a></li>
             </ul>
-            {{ form_widget(form.filesTabValidator, { attr: { class: 'ignore'}}) }}
-
+            <label for="files" class="error"></label>
```
praneethpr, line 18: Do you also want to add this to templates/DatasetReview/file-submit.html.twig

praneethpr (commented):
Tested and works fine. All edge cases covered

assets/static/js/datasetReview.js
```diff
@@ -44,15 +44,38 @@ $(document).ready(function(){
...
+        let datasetFileTransferType = $("#filetabs .ui-tabs-active").attr("datasetFileTransferType");
+        // set the datasetFileTransferType
+        $("#datasetFileTransferType").val(datasetFileTransferType);
+    });
+
     regForm.validate({
         rules: {
             temporalExtentBeginPosition: "trueISODate",
             temporalExtentEndPosition: "trueISODate",
+            filesUploaded:{
```
mickel1138, line 63: Do we want to add the `groups` here too?

mickel1138 (approved):
Everything looks good now.

fightingtexasaggie (approved):
seems to work as expected. I did notice that in dataset review, the 'remotely hosted function' selection is not validated until the review is accepted, not simply when flipping tabs. This is possibly too minor to matter.

PR #1004, opened 2021-09-16
Upgraded webpack encore because of security bug in ansi-html used by dev-server

mickel1138:
Because ansi-html is not maintained, or the maintainer(s) are not willing to fix the security issue, a community version has been created (ansi-html-community). This is only being used by encore dev-server.


praneethpr (approved):
Build works and tested

fightingtexasaggie (approved):
ran fine

PR #1006, opened 2021-09-20
Set content type explicitly.

fightingtexasaggie:


src/Util/Datastore.php
```diff
@@ -4,7 +4,7 @@
 
 use League\Flysystem\FileExistsException;
 use League\Flysystem\FilesystemInterface;
-
+use phpDocumentor\Reflection\Types\Boolean;
```
mickel1138, line 7: What does this do?

src/Util/Datastore.php
```diff
@@ -62,6 +62,18 @@ public function getFile(string $filePath): array
...
     }
+    
+    /**
+     * Queries Flysys for a file's mimetype
+     *
+     * @param string $filePath The file's path.
+     *
+     * @return string
+     */
+    public function getMimeType(string $filePath): string
```
mickel1138, line 73: This can not guaranteed to return a string, so leave it blank, also GCS error.

src/Controller/DownloadController.php
```diff
@@ -136,13 +137,19 @@ public function httpAction(Dataset $dataset, Datastore $dataStore, LogActionItem
...
                     $filename = basename($zipFilePath);
+                    $mimeType = 'application/zip';
                 }
 
                 $disposition = HeaderUtils::makeDisposition(
                     HeaderUtils::DISPOSITION_ATTACHMENT,
                     $filename
                 );
+
+                if (null === $mimeType or false === $mimeType) {
```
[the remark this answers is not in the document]
mickel1138: Also, you could set this WAY on top, and if it get's overwritten, good, if not, it's set.

src/Util/Datastore.php
```diff
@@ -62,6 +62,18 @@ public function getFile(string $filePath): array
...
     }
+    
+    /**
+     * Queries Flysys for a file's mimetype
+     *
+     * @param string $filePath The file's path.
+     *
+     * @return string|false
+     */
+    public function getMimeType(string $filePath)
```
praneethpr, line 73: return type missing

fightingtexasaggie: yes, because the filesystemInterface->getMimeType() itself returns string or false. In PHP 8 we'll be able to specify union types. There are suggestions of making a return type interface to get around this now, but that's probably way overkill.

praneethpr (approved):

praneethpr (approved):

PR #1009, opened 2021-09-22
Added check for authentication for dif

praneethpr:


fightingtexasaggie (approved):
Screenshot from 2021-09-23 14-39-01
Maybe the error message is a little confusing, but it does tell the user what to do, so I think we can move forward and refine later per PO.

PR #1011, opened 2021-09-28
Changed style for maps to give spatial extent more darker outline

praneethpr:


fightingtexasaggie (approved):
looks a lot better over land imho

PR #1014, opened 2021-10-04
Only show logged out errors and server unreachable error

praneethpr:


assets/static/js/dataset-submission.js
```diff
@@ -325,11 +325,10 @@ $(function() {
                 if ([401].includes(jqXHR.status)) {
                     loggedOutPopup.show();
                 } else {
-                    if (jqXHR.status !== 0) {
-                        message = jqXHR.responseJSON == null ? errorThrown: jqXHR.responseJSON.message;
+                    if (jqXHR.status === 0) {
+                        errorPopup.show();
+                        $("#errorPopupText").html(message);
```
mickel1138, line 330: The message can be hard-coded here, it will never we anything else.
This is going to get refactors in another branch anyway tho.

mickel1138 (approved):

mickel1138:
It still tries to patch/put the submission between tab changes, but we'll fix that somewhere else.

PR #1015, opened 2021-10-04
Feature/pelagos 4616 disable submit button if form not valid

praneethpr:


praneethpr:
> a minor edge case is if you enter a remote URL, then enter a file, then delete the file. The submit button goes gray even though a remote value exists. This can be worked around by adding and deleting a character at the end of the URL. This is probably too minor an edge case to matter much imho, but I wanted to let you know.

Definitely a bug, will fix it

mickel1138:
When you cause the form to validate from a new submission, and fill everything out, then come back to file submit, the RED error text still shows.

PR #1016, opened 2021-10-06
Feature/pelagos 4619 populate dropdown for large file url will parse string

fightingtexasaggie:
This handles the parsing for dataset-submission.

### DIFF
--- a/assets/static/js/dataset-submission.js
+++ b/assets/static/js/dataset-submission.js
@@ -268,7 +268,8 @@ $(function() {
         // Populate dropdown with list of folders
         $.getJSON(url, function (data) {
             $.each(data, function (key, value) {
-                dropdown.append($('<option></option>').attr('value', value).text(key));
+                let display = value.substring(0,10);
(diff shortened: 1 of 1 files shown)


assets/static/js/dataset-submission.js
```diff
@@ -268,7 +268,17 @@ $(function() {
...
+                value = value.replace(/\/$/, "");
+                let splitPath = value.toString().split("/");
+                dropdown.append($('<option></option>').attr('value', value).text(
+                    '/'
+                    + splitPath[splitPath.length - 3]
+                    + '/'
+                    + splitPath[splitPath.length - 2]
+                    + '/'
+                    + splitPath[splitPath.length - 1]
+                ));
```
mickel1138, line 281: I think a regex looks cleaner than a split and concat.

```suggestion
                const regex = new RegExp('(?:.(?!\/.*\/.*\/))+$');
                let text = regex.exec(value)[0];
                dropdown.append($('<option></option>').attr('value', value).text(text));
```

PR #1033, opened 2021-10-28
Feature/pelagos 4633 add switch or parameter to turn on off large file tab

fightingtexasaggie:


assets/static/css/dataset-submission.css
```diff
@@ -158,3 +158,8 @@ table.dataTable.stripe tbody tr:hover {
 #clearLargeFilePath {
     margin-left: 30px;
 }
+
+.disabled-tab
```
praneethpr, line 162:
```suggestion
.ui-tabs .ui-tabs-panel.disabled-tab
```

templates/DatasetSubmission/file-submit.html.twig
```diff
@@ -31,7 +33,14 @@
             </div> <!-- tabs-1 -->
 
             {% set sftpuser=app.user.isPosix() %}
-            <div id="tabs-2"> <!-- Upload via SFTP/Ingest-->
+            {% if largefile_tab != 'disabled_hidden' %}
+                {% if largefile_tab == 'disabled_message'%}
+                    <div id="tabs-2" class="disabled-tab" style="background-color: lightgray"> <!-- Upload via SFTP/Ingest-->
```
praneethpr, line 38: remove the inline css

mickel1138: Or add to class

.env
```diff
@@ -103,3 +103,6 @@ MESSENGER_TRANSPORT_DSN=doctrine://default
 ###< symfony/messenger ###
 
 CLAMD_SOCKET='unix:///run/clamd.scan/clamd.sock'
+
+# Valid states are enabled (default), disabled_hidden, disabled_message
```
mickel1138, line 107: I'd like to suggest options: `enabled` `hidden` and `disabled`

templates/DatasetSubmission/file-submit.html.twig
fightingtexasaggie: It doesn't work added to the class. I too don't like using inline, but I couldn't get it to override otherwise.

mickel1138 (approved):
LGTM! Tested, works.

PR #1037, opened 2021-11-02
Misc/doctrine upgrade 2

mickel1138:


.env
```diff
@@ -22,9 +22,11 @@ APP_SECRET=7fd781d3c8b25811ddd82260d88cc496
...
 ###> doctrine/doctrine-bundle ###
 # Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
-# For an SQLite database, use: "sqlite:///%kernel.project_dir%/var/data.db"
-# Configure your db driver and server_version in config/packages/doctrine.yaml
-DATABASE_URL=postgres://user:password@127.0.0.1:3306/db_name
+# IMPORTANT: You MUST configure your server version, either here or in config/packages/doctrine.yaml
+#
+# DATABASE_URL="sqlite:///%kernel.project_dir%/var/data.db"
+# DATABASE_URL="mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7"
+DATABASE_URL="postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=13&charset=utf8"
```
fightingtexasaggie, line 29: Do we want to put in 9.6 for default, or perhaps 14?

fightingtexasaggie (approved):
some of the lists take a long time to load, but I think this isn't due to doctrine's update....we just seldom check them. We should probably agree on :void vs : void at some point, but both satisfy linting.

I gave it a pretty good shake and everything seems to be working, aside of a timeout on DIF that we addressed elsewhere.

.env
mickel1138: It's just an example

PR #1047, opened 2021-11-12
Remove PDF link.

fightingtexasaggie:
Per discussion earlier, we're leaving the actual pdf-formatted view in place for possible future use, so this just removes the link, which is now broken anyway.


mickel1138 (approved):
PDF button is no longer there.

PR #1048, opened 2021-11-15
Added spinner and button disable

praneethpr:


fightingtexasaggie (approved):
tested this out and it appears to work as intended.

mickel1138 (approved):
This should do it.

PR #1049, opened 2021-11-15
Feature/pelagos 4656 update doctine config removing db ver

fightingtexasaggie:
This was an oversight from before when we updated the .env for the new DB configuration. We should have removed it from the doctrine.yaml config at that time. It still worked, but wasn't ideal.

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -9,9 +9,6 @@ doctrine:
                 class: App\DoctrineExtensions\DBAL\Types\CITextType
                 commented: false
 
-        # IMPORTANT: You MUST configure your server version,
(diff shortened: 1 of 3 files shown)


config/packages/doctrine.yaml
```diff
@@ -9,9 +9,6 @@ doctrine:
                 class: App\DoctrineExtensions\DBAL\Types\CITextType
                 commented: false
 
-        # IMPORTANT: You MUST configure your server version,
-        # either here or in the DATABASE_URL env var (see .env file)
-        server_version: '9.6'
```
mickel1138, line 14: Don't delete this section, just comment out the server version.

mickel1138 (approved):
All good.

PR #1050, opened 2021-11-15
Refactored submission and review to use entity manager instead of han…

praneethpr:
…dler

### DIFF
--- a/src/Controller/UI/DatasetReviewController.php
+++ b/src/Controller/UI/DatasetReviewController.php
@@ -436,7 +436,8 @@ class DatasetReviewController extends AbstractController
      */
     public function postAction(Request $request, int $id = null, MessageBusInterface $messageBus)
     {
-        $datasetSubmission = $this->entityHandler->get(DatasetSubmission::class, $id);
+        $entityManager = $this->getDoctrine()->getManager();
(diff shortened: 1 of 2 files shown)


src/Controller/UI/DatasetReviewController.php
```diff
@@ -532,10 +507,18 @@ public function postAction(Request $request, int $id = null, MessageBusInterface
...
                     $datasetSubmission = $datasetSubmission->getDataset()->getDatasetSubmission();
                     $datasetSubmission->setFileset($newFileset);
-                    $this->entityHandler->update($datasetSubmission);
+                    $entityManager->persist($newFileset);
                 }
             }
 
+            $entityManager->flush();
+
+            // update MDAPP logs after action is executed.
```
mickel1138, line 516: I don't think mdapp still exists. But why is this now added? (The event dispatch)

praneethpr: I moved it, and i think we still have the

mickel1138 (approved):

src/Controller/UI/DatasetReviewController.php
fightingtexasaggie: Yes, it still exists as a place to hold certain logs. This is indeed old, but I think still used, primarly just for the log content.

PR #1053, opened 2021-11-17
Disabled button on click

praneethpr:


mickel1138 (approved):
Tested, and works!

fightingtexasaggie (approved):
I think this'll do it! Tested well, and seems like the right way to handle the issue.

PR #1056, opened 2021-11-29
Feature/pelagos 4646 create file checker command

mickel1138:


fightingtexasaggie (approved):
Like the change to the description, I think we should probably rename the command itself at some point, but this is fine for now.

mickel1138:
> Like the change to the description, I think we should probably rename the command itself at some point, but this is fine for now.

It can be changed, what shall we change it to?

praneethpr (approved):

PR #1057, opened 2021-12-01
Catch mkdir error and reject, and addition logging

mickel1138:


src/MessageHandler/ProcessFileHandler.php
```diff
@@ -110,7 +109,9 @@ public function __invoke(ProcessFile $processFile)
         @$fileStream = fopen($filePath, 'r');
 
         if ($fileStream === false) {
-            $file->setDescription("Unreadable Queued File:" . error_get_last()['message']);
+            $lastErrorMessage = error_get_last()['message'];
+            $this->logger->error(sprintf('Unreadable Queued File: "%s"', $lastErrorMessage, $loggingContext));
+            $file->setDescription('Unreadable Queued File:' . $lastErrorMessage);
```
fightingtexasaggie, line 114: Should we consider setting some sort of max-tries? We'd have to store it in the entity if we did though, but currently if a file is truly unreadable, wouldn't it try forever?

mickel1138: This section ends with a return, so no retry.
But otherwise retries are currently limited to (default) 3.

fightingtexasaggie (dismissed):
tested on the same files that previously failed, no files were left in error state. This seems good, but I also feel we should compare the resulting zips with the original unpack dirs, just as a failsafe. I've got some similar comparison tools I used on LTFS that I can re-purpose here.

praneethpr (approved):
LGTM!

fightingtexasaggie (approved):
tested, like the idea

PR #1058, opened 2021-12-02
Fix for 0 byte uploading error

mickel1138:


praneethpr (approved):
Tested! I think this approach works

praneethpr (approved):

fightingtexasaggie (approved):
tested and confirmed works. The only thing not preserved is empty folders, but I think we're all ok with that. I also tested that a folder with only an empty file is preserved, as expected.

PR #1062, opened 2021-12-10
Added padding bottom

praneethpr:


templates/Dataland/v2/index.html.twig
```diff
@@ -19,7 +19,7 @@
 {% block body %}
     <div class="bootstrap">
         <div class="container">
-            <div class="row">
+            <div class="row pb-2">
```
mickel1138, line 22: This fixed it, but in my mind is a bit tight. Should be consider `pb-4`?

mickel1138 (dismissed):
Tested, good.

templates/Dataland/v2/index.html.twig
praneethpr: You are right, i changed it to increase the padding

mickel1138 (approved):

PR #1075, opened 2022-01-19
Added file ID to file manager

mickel1138:
Updated delete API method to accept ID also

### DIFF
--- a/src/Entity/Fileset.php
+++ b/src/Entity/Fileset.php
@@ -140,7 +140,7 @@ class Fileset extends Entity
      *
      * @throws Exception When the file by ID was not found.
      *
-     * @return File
+     * @return File|null
      */
     public function getFileById(int $id) : ?File
     {
--- a/src/Entity/Fileset.php
+++ b/src/Entity/Fileset.php
@@ -133,6 +133,28 @@ class Fileset extends Entity
         });
     }
 
+     /**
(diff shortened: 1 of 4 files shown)


src/Entity/Fileset.php
```diff
@@ -133,6 +133,28 @@ public function getDeletedFiles() : Collection
...
+     * @param integer $id The file ID.
+     *
+     * @throws Exception When the file by ID was not found.
+     *
+     * @return File
+     */
+    public function getFileById(int $id) : ?File
+    {
+        $files = $this->files->filter(function (File $file) use ($id) {
+            return $file->getId() === $id;
```
fightingtexasaggie, line 148: wouldn't this end up returning a boolean?

src/Entity/Fileset.php
```diff
@@ -133,6 +133,28 @@ public function getDeletedFiles() : Collection
...
     }
 
+     /**
+     * Getter for file by ID.
+     *
+     * @param integer $id The file ID.
+     *
+     * @throws Exception When the file by ID was not found.
+     *
+     * @return File
```
mickel1138, line 143:
```suggestion
     * @return File|null
```

fightingtexasaggie (approved):
Thanks for the walk-through! Looks good.

PR #1078, opened 2022-01-21
Added HRI to template switcher

mickel1138:
If a custom array of settings is not defined, the default (GRIIDC) will be used.


fightingtexasaggie (approved):
displays as expected. 👍🏼

PR #1082, opened 2022-01-24
Changed V2 routes, and removed old download function.

mickel1138:
A download method data/udi/download wasn't working, I think because we moved to the "new" datastore. The link was not being used anywhere, so this function was removed.


fightingtexasaggie (approved):
dataland route now displaying v2 as per ticket.

PR #1087, opened 2022-01-27
Added reset, and required for RGs also Restrict to DRPMs

mickel1138:


assets/js/vue/components/information-product/InformationProductForm.vue
```diff
@@ -190,7 +193,17 @@ export default {
       addedRgShortName: '',
     };
   },
-
+  computed: {
+    researchGroupsSelected () {
+      return this.form.selectedResearchGroups.length > 0;
```
fightingtexasaggie, line 198: I think this is too late here. We want to return whether or not there is a value for the add-research-groups field.

fightingtexasaggie (commented):
Can we check that there's a value in the 'add research group' field before allowing the 'add research group' button? Right now it is possible to add a blank research group, then submit the form with it.

fightingtexasaggie (approved):
Clear, and works nicely.

PR #1089, opened 2022-02-03
Made short name required

mickel1138:
Added required to form
Added assert to entity

Merge is going to DEV, so this will become required for the IP epic releases. 
Let me know if we think that must change.


praneethpr (approved):
Tested and works

PR #1091, opened 2022-02-07
Removed webpack-devserver older version dependency

mickel1138:


fightingtexasaggie (approved):
I gave it a fairly complete shakedown and everything seems in order.

PR #1093, opened 2022-02-08
Don't rely on DOI being set to display. Null ok if DOI service slow, …

fightingtexasaggie:
Found this testing...minor, unlikely, but will break dataland if the DOI service is slow.


mickel1138 (approved):
This is fine.

PR #1094, opened 2022-02-09
Feature/pelagos 4699 add additional properties

mickel1138:


src/Entity/InformationProduct.php
```diff
@@ -311,4 +321,28 @@ public function removeResearchGroup(ResearchGroup $researchGroup): self
 
         return $this;
     }
+
+    public function getRemoteUri(): ?string
```
praneethpr, line 325: Comments missing for all getters and setters

PR #1099, opened 2022-02-16
Fixed research group name length and dataset identified status

praneethpr:


src/Controller/UI/ReportFundingOrganizationController.php
```diff
@@ -81,7 +82,7 @@ private function createCsvReportFileName(string $fundingOrgName, string $funding
     {
         $nowDateTimeString = date(self::REPORTFILENAMEDATETIMEFORMAT);
 
-        $researchGroupNameSubstring = substr($fundingOrgName, 0, self::MAXRESEARCHGROUPLENGTH);
+        $researchGroupNameSubstring = substr($fundingOrgName, 0, self::MAXFUNDINGORGLENGTH);
```
mickel1138, line 85: Where is this used? Also the variable name is confusing?

praneethpr: Not used, copy paste code and i don't know why i made changes to that. Fixed it now

mickel1138 (approved):
This should work.

PR #1100, opened 2022-02-16
Feature/pelagos 4700 add file dropzone for chunking

mickel1138:


assets/js/vue/components/information-product/InformationProductForm.vue
```diff
@@ -321,6 +345,70 @@ export default {
...
+  chunkData.fileName = fileName;
+  chunkData.dztotalfilesize = file.size;
+  postApi(
+    // eslint-disable-next-line no-undef
+    `${Routing.generate('pelagos_api_add_file_information_product')}`,
+    chunkData,
+  ).then((response) => {
+    const fileId = response.data.id;
+    file.fileId = fileId;
+    thisComponent.form.file = fileId;
```
praneethpr, line 373: If the file is removed after uploading, we need to set it to null again

praneethpr (commented):
Works fine, but we could do some more improvements and validation checks

assets/js/vue/components/information-product/InformationProductForm.vue
mickel1138: Worse than that! We should also delete it from disk. Glad you noticed it too.

praneethpr (approved):

praneethpr (approved):

PR #1101, opened 2022-02-16
Epic/information products

mickel1138:


fightingtexasaggie (commented):
Can we get an environment variable for storage location other than using HOMEDIR_PREFIX? I feel weird about putting intended-permanent stuff in there.

PR #1102, opened 2022-02-16
Cmd to get aggregate FS from cold manifest.

fightingtexasaggie:
needed this for now...might be useful later


fightingtexasaggie:
not needed

PR #1103, opened 2022-02-16
Moved zipping of files inside datasetsubmission handler

praneethpr:


mickel1138 (approved):
Tested, and works for me.

fightingtexasaggie (approved):
We just have to realize zip messages related to submission will now be in filer.log and not zipper.log.

PR #1104, opened 2022-02-16
Fixed yarn lock, for dev-server

mickel1138:
This should be better now, the last time this didn't resolve the Yarn lock conflict.


fightingtexasaggie (approved):
I hammered on this pretty good and everything seems to build and work fine on all the changed versions.

PR #1106, opened 2022-02-18
Feature/pelagos 4703 add remote uri to UI

mickel1138:


fightingtexasaggie (approved):
seems reasonable. I think the file xor uri is pretty intuitive to the end user.

PR #1108, opened 2022-02-22
Look for transfer status instead of files being procd.

mickel1138:
Also added transfer state after submit and review.


fightingtexasaggie (approved):
I'm good with this. We can test on test by adding a sleep(60) line in DatasetSubmissionFileHandler before it sets the status.

PR #1112, opened 2022-02-24
Added IP filer

mickel1138:
Also added creator for each thing (we should do this more)


src/MessageHandler/InformationProductFilerHandler.php
```diff
@@ -0,0 +1,144 @@
...
+        $file->setModifier($systemPerson);
+
+        $loggingContext = array(
+            'information_product_id' => $informationProductId,
+            'file' => $filePath,
+        );
+
+        $this->logger->info('Dataset submission process started', $loggingContext);
+
+        $destinationPath = 'information_products'
```
fightingtexasaggie, line 89: 👍🏼

fightingtexasaggie (approved):
yes, this is working, logging properly and migrating files into <location>/information_product/<first-20-of-title>/originalfilename_xxxxx. I think the normal handler is fine for it too, as these should be smallish anyway, so good there.

PR #1114, opened 2022-03-01
Feature/pelagos 4743 add processed file filter

mickel1138:
Only show file link, if file status = done.


mickel1138:
Also get file from store instead now!

fightingtexasaggie (commented):
I'm getting a 500 error from the file upload api saying "The identifier id is missing for a query of App\\Entity\\InformationProduct"

Failing Api call is to: /api/add_file_to_information_product

Confirmed in sync mode, filesystem dir owned by symfony server username, clean env.

fightingtexasaggie (approved):
Ok, this is working now with both immediate deletes and deleting from the 'edit' mode. I think this gets us going for now.

PR #1117, opened 2022-03-01
Check for accepted status on http d/l controller.

fightingtexasaggie:
way simplier than expected...I missing something?

### DIFF
--- a/src/Controller/DownloadController.php
+++ b/src/Controller/DownloadController.php
@@ -111,6 +111,9 @@ class DownloadController extends AbstractController
         if ($dataset->isRestricted()) {
             throw new BadRequestHttpException('Unable to download restricted dataset');
         }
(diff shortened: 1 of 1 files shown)


src/Controller/DownloadController.php
```diff
@@ -111,6 +111,9 @@ public function httpAction(Dataset $dataset, Datastore $dataStore, LogActionItem
         if ($dataset->isRestricted()) {
             throw new BadRequestHttpException('Unable to download restricted dataset');
         }
+        if ($dataset->getDatasetStatus() !== Dataset::DATASET_STATUS_ACCEPTED) {
```
praneethpr, line 114: You could combine the above condition with restricted and make a single `if` block with availability status

mickel1138 (approved):
LGTM

PR #1120, opened 2022-03-08
Added message to Update

mickel1138:


src/Controller/Api/InformationProductController.php
```diff
@@ -57,7 +57,8 @@ public function getInformationProduct(InformationProduct $informationProduct, Se
     /**
      * Creates a new Information Product
      *
-     * @param Request $request
+     * @param Request             $request
```
praneethpr, line 60: Missing descriptions for the variables

src/Controller/Api/InformationProductController.php
```diff
@@ -104,8 +105,9 @@ public function createInformationProduct(Request $request, MessageBusInterface $
     /**
      * Updates the Information Product
      *
-     * @param Request $request
-     * @param InformationProduct $informationProduct
+     * @param Request             $request
```
praneethpr, line 108: Same here

fightingtexasaggie (commented):
When attempting to delete a file just uploaded, this is getting thrown as the string "null" is getting passed into a parameter, so the if($param) is true on line 397 of  src/Controller/Api/InformationProductController.php 

"An exception occurred while executing 'SELECT t0.title AS title_1, t0.creators AS creators_2, t0.publisher AS publisher_3, t0.external_doi AS external_doi_4, t0.published AS published_5, t0.remote_resource AS remote_resource_6, t0.remote_uri AS remote_uri_7, t0.id AS id_8, t0.creation_time_stamp AS creation_time_stamp_9, t0.modification_time_stamp AS modification_time_stamp_10, t0.file_id AS file_id_11, t0.creator_id AS creator_id_12, t0.modifier_id AS modifier_id_13 FROM information_product t0 WHERE t0.id = ?' with params [\"null\"]:\n\nSQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for type integer: \"null\"\nCONTEXT: unnamed portal parameter $1 = '...'"

fightingtexasaggie (approved):
looks good, tested good.

PR #1125, opened 2022-03-09
Add tool to detect missing files.

fightingtexasaggie:


src/Command/PelagosCheckForMissingFilesCommand.php
```diff
@@ -0,0 +1,113 @@
...
+        foreach ($datasets as $dataset) {
+            $checkedUdi = $dataset->getUdi();
+            $io->note("checking dataset: $checkedUdi\n");
+            /** @var Dataset $dataset */
+            $datasetSubmission = $dataset->getDatasetSubmission();
+            if ($datasetSubmission instanceof DatasetSubmission) {
+                $fileset = $datasetSubmission->getFileset();
+                if ($fileset instanceof Fileset) {
+                    $files = $fileset->getAllFiles()->filter(function (File $file) {
+                        return $file->getStatus() !== File::FILE_ERROR;
```
mickel1138, line 94: How about new files, or file in queue?

src/Command/PelagosCheckForMissingFilesCommand.php
```diff
@@ -0,0 +1,113 @@
...
+                    });
+                    $fileCount = count($files);
+                    $io->writeln('Dataset ' . $dataset->getUdi() . ' has ' . $fileCount . ' files.' . "\n");
+                    /** @var File $file */
+                    foreach ($files as $file) {
+                        $originalFile = $file->getPhysicalFilePath();
+                        if (substr($originalFile, 0, 1) != '/') {
+                            $originalFile = $this->dataStoreDirectory . "/$originalFile";
+                        }
```
mickel1138, line 103: Let the data store check the existence, don't do it like this. `filesystem->has`

src/Command/PelagosCheckForMissingFilesCommand.php
fightingtexasaggie: This should be getting them...any file not in error.

src/Command/PelagosCheckForMissingFilesCommand.php
fightingtexasaggie: switching to Filesystem component for this.

src/Command/PelagosCheckForMissingFilesCommand.php
fightingtexasaggie: In the end, decided to make it just check the done files.

config/services.yaml
```diff
@@ -125,6 +125,11 @@ services:
             string $passwordExpiryWarn: "%env(ACCOUNT_PASSWORD_EXPIRATION_WARN)%"
             string $hostName: "%env(HOSTNAME)%"
 
+    App\Command\PelagosCheckForMissingFilesCommand:
+        bind:
+            string $dataStoreDirectory: "%env(DATA_STORE_DIRECTORY)%"
+            $datastoreFlysystem: "@oneup_flysystem.datastore_flysystem_filesystem"
```
praneethpr, line 131: We should inject src/Util/Datastore, we shouldn't be using it directly either from a controller or command

src/Command/PelagosCheckForMissingFilesCommand.php
```diff
@@ -0,0 +1,122 @@
...
+                if ($fileset instanceof Fileset) {
+                    $files = $fileset->getAllFiles()->filter(function (File $file) {
+                        return $file->getStatus() === File::FILE_DONE;
+                    });
+                    $fileCount = count($files);
+                    $io->writeln('Dataset ' . $dataset->getUdi() . ' has ' . $fileCount . ' files.' . "\n");
+                    /** @var File $file */
+                    foreach ($files as $file) {
+                        $filePath = $file->getPhysicalFilePath();
+                        if (!$this->datastoreFlysystem->has($filePath)) {
```
fightingtexasaggie, line 112: Extend the Datastore util instead!

praneethpr (dismissed):

mickel1138 (approved):

PR #1127, opened 2022-03-11
Added new transfer state

mickel1138:
Set the Being Processed in review.

### DIFF
--- a/src/Controller/UI/DatasetReviewController.php
+++ b/src/Controller/UI/DatasetReviewController.php
@@ -479,6 +479,7 @@ class DatasetReviewController extends AbstractController
                 foreach ($fileset->getNewFiles() as $file) {
                     $file->setStatus(File::FILE_IN_QUEUE);
                 }
+                $datasetSubmission->setDatasetFileTransferStatus(DatasetSubmission::TRANSFER_STATUS_BEING_PROCESSED);
             }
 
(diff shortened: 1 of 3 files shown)


fightingtexasaggie (dismissed):
This solution seems appropriate.

src/Controller/UI/DatasetReviewController.php
```diff
@@ -479,6 +479,7 @@ public function postAction(Request $request, int $id = null, MessageBusInterface
                 foreach ($fileset->getNewFiles() as $file) {
                     $file->setStatus(File::FILE_IN_QUEUE);
                 }
+                $datasetSubmission->setDatasetFileTransferStatus(DatasetSubmission::TRANSFER_STATUS_BEING_PROCESSED);
```
praneethpr, line 482: If we are going to set it, it needs to be outside the fileset condition because it is not dependent on that

praneethpr, line 482: We need to make the same change in Dataset Submission when a dataset is submitted

src/Entity/DatasetSubmission.php
```diff
@@ -115,6 +115,11 @@ class DatasetSubmission extends Entity
      */
     const TRANSFER_STATUS_REMOTELY_HOSTED = 'RemotelyHosted';
 
+    /**
+     * Indicates that the dataset is remotely hosted.
```
praneethpr, line 119: Change description to match the constant

praneethpr (approved):

PR #1130, opened 2022-03-21
New API route for dataset json

mickel1138:


fightingtexasaggie (commented):
We should ask William how to handle the cold-stored datasets. Currently the fileset methods for count and size will return that of the stubs.

fightingtexasaggie (approved):

PR #1134, opened 2022-03-21
Add default in template for null numberOfFiles, show 0.

fightingtexasaggie:


mickel1138 (approved):
The improvement is fine, but this should not happen.

PR #1137, opened 2022-03-24
Feature/pelagos 4776 create product type entity

praneethpr:


src/Entity/InformationProductType.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
```
mickel1138, line 2: extra line

src/Entity/InformationProduct.php
```diff
@@ -377,4 +388,28 @@ public function setFile(?File $file): self
...
+    }
+
+    /**
+     * Remover for Information Product type.
+     *
+     * @param InformationProductType $informationProductType Single information product type to be removed.
+     *
+     * @return void
+     */
+    public function removeFile(InformationProductType $informationProductType): void
```
fightingtexasaggie, line 411: Why method name of removeFile?

src/Entity/InformationProductType.php
```diff
@@ -0,0 +1,15 @@
...
+use App\Repository\InformationProductTypeRepository;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity(repositoryClass=InformationProductTypeRepository::class)
+ */
+class InformationProductType extends Entity
+{
+    use DescriptorTypeTrait;
```
fightingtexasaggie, line 14: Is there additional refactor that will use this trait? Otherwise, why use a trait once?

migrations/Version20220324132210.php
```diff
@@ -0,0 +1,41 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220324132210 extends AbstractMigration
+{
+    public function getDescription() : string
+    {
+        return '';
```
mickel1138, line 17: Add description

src/Entity/InformationProductType.php
praneethpr: We thought we will use it for Digital Resource type which also would have only 'id' and 'description' attributes

src/Entity/InformationProduct.php
praneethpr: Good catch! Incorrect method type

src/Entity/InformationProductType.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace App\Entity;
+
+use App\Repository\InformationProductTypeRepository;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity(repositoryClass=InformationProductTypeRepository::class)
+ */
+class InformationProductType extends Entity
```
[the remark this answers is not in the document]
mickel1138: Just tested, and this can not be call `type` 

"Could not load type \"App\\Entity\\InformationProductType\": class does not implement \"Symfony\\Component\\Form\\FormTypeInterface\"."

src/Entity/InformationProduct.php
```diff
@@ -119,6 +119,17 @@ class InformationProduct extends Entity
...
     private $file;
 
+    /**
+     * The collection of types for this Information Product.
+     *
+     * @var Collection
+     *
+     * @Serializer\MaxDepth(1)
+     *
+     * @ORM\OneToMany(targetEntity="InformationProductTypeDescriptor", mappedBy="InformationProduct")
```
mickel1138, line 129: Since there is not reverse, the mappedBy may not be needed. (As it doesn't exist)

src/Entity/InformationProductType.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace App\Entity;
+
+use App\Repository\InformationProductTypeDescriptorRepository;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity(repositoryClass=InformationProductTypeDescriptorRepository::class)
+ */
+class InformationProductTypeDescriptor extends Entity
```
mickel1138, line 11: Rename the file too

src/Entity/InformationProduct.php
praneethpr: It requires that mappedBy or else it breaks

fightingtexasaggie (approved):
seems ready, lg.

mickel1138 (approved):
Tested, issues resolved.

PR #1143, opened 2022-04-05
Feature/pelagos 4785 add digitial resource type descriptor

mickel1138:


src/Entity/InformationProduct.php
```diff
@@ -414,6 +428,46 @@ public function removeInformationProductType(InformationProductTypeDescriptor $i
...
+     * Add Digital Resource Type Descriptor to Information Product.
+     *
+     * @param DigitalResourceTypeDescriptor $digitalResourceTypeDescriptor
+     *
+     * @return self
+     */
+    public function addDigitalResourceTypeDescriptor(DigitalResourceTypeDescriptor $digitalResourceTypeDescriptor): self
+    {
+        if (!$this->digitalResourceTypeDescriptors->contains($digitalResourceTypeDescriptor)) {
+            $this->digitalResourceTypeDescriptors[] = $digitalResourceTypeDescriptor;
```
praneethpr, line 451:
```suggestion
            $this->digitalResourceTypeDescriptors->add($digitalResourceTypeDescriptor);
```

src/Entity/InformationProduct.php
```diff
@@ -414,6 +428,46 @@ public function removeInformationProductType(InformationProductTypeDescriptor $i
...
+     *
+     * @param DigitalResourceTypeDescriptor $digitalResourceTypeDescriptor
+     *
+     * @return self
+     */
+    public function removeDigitalResourceTypeDescriptor(DigitalResourceTypeDescriptor $digitalResourceTypeDescriptor): self
+    {
+        $this->digitalResourceTypeDescriptors->removeElement($digitalResourceTypeDescriptor);
+
+        return $this;
```
praneethpr, line 468: Do we need the return on this?

src/Entity/InformationProduct.php
```diff
@@ -414,6 +428,46 @@ public function removeInformationProductType(InformationProductTypeDescriptor $i
...
+     *
+     * @return self
+     */
+    public function addDigitalResourceTypeDescriptor(DigitalResourceTypeDescriptor $digitalResourceTypeDescriptor): self
+    {
+        if (!$this->digitalResourceTypeDescriptors->contains($digitalResourceTypeDescriptor)) {
+            $this->digitalResourceTypeDescriptors[] = $digitalResourceTypeDescriptor;
+        }
+
+        return $this;
```
praneethpr, line 454: Do we need the return on this?

src/Event/InformationProductListener.php
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Listener class for Information Product and Types related events.
+ */
+class InformationProductListener
+{
+    /**
+     * Listener on Pre-Remove event.
+     *
+     * @param LifecycleEventArgs $args The Lifecycle event arguments.
```
praneethpr, line 19: Add throws tag

src/Entity/InformationProduct.php
mickel1138: It is added by make, I kinda like chaining, we should keep it.

src/Entity/InformationProduct.php
mickel1138: I like the chaining, so I think we should keep it.

fightingtexasaggie (approved):
looks good. If we end up making a bunch of these, should some of the more common crud functions be placed into a trait?

praneethpr (approved):
Looks good!

PR #1144, opened 2022-04-06
Feature/pelagos 4789 add event listener and ezadmin delete preventor

mickel1138:
Would like to rename the InformationProductTypeDescriptor Entity, but can be done in another PR.

### DIFF
--- a/src/Repository/InformationProductRepository.php
+++ b/src/Repository/InformationProductRepository.php
@@ -56,9 +56,9 @@ class InformationProductRepository extends ServiceEntityRepository
     }
 
     /**
-     * Get Information Products by Digital Resource Type.
+     * Get Information Products by Product Type Descriptor.
      *
-     * @param DigitalResourceTypeDescriptor $digitalResourceTypeDescriptor
+     * @param InformationProductTypeDescriptor $productTypeDescriptor
      *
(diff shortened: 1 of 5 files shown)


fightingtexasaggie (approved):
Looks mostly like the other one, and all tested well. LGTG!

src/Repository/InformationProductRepository.php
```diff
@@ -54,6 +55,22 @@ public function findByDigitalResourceTypeDescriptor(DigitalResourceTypeDescripto
...
+     *
+     * @return array An array of Information Products.
+     */
+    public function findByProductTypeDescriptor(InformationProductTypeDescriptor $productTypeDescriptor): array
+    {
+        $qb = $this->createQueryBuilder('ip');
+        $qb->setParameter('productTypeDescriptor', $productTypeDescriptor);
+        $qb->where($qb->expr()->isMemberOf(':productTypeDescriptor', 'ip.informationProductTypeDescriptors'));
+
+        return $qb->getQuery()->getResult();
```
praneethpr, line 71: Does this always return array even if it doesn't find anything?

mickel1138: It should yes. An empty one.

PR #1147, opened 2022-04-12
Misc/use psr7 streams

mickel1138:


composer.json
```diff
@@ -21,7 +21,8 @@
         "guzzlehttp/guzzle": "~6.0",
         "hylianshield/base32-crockford": "^1.0",
         "jms/serializer-bundle": "^3.0",
-        "maennchen/zipstream-php": "^2.1",
+        "laminas/laminas-diactoros": "^2.9",
```
mickel1138, line 24: Because we already have Guzzle installed, maybe we should use it instead? Thoughts?

- [ ] Keep it, it's probably better.
- [ ] Get rid of it, Guzzle has enough PSR7 coverage for what we need.

composer.json
```diff
@@ -21,7 +21,8 @@
...
+        "maennchen/zipstream-php": "dev-master",
```
fightingtexasaggie, line 25: I'd feel better with a hard-coded static commit we can test against than a moving target at the bleeding edge ;-)

mickel1138: At the time we run the composer add, that will be the locked version.

src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -192,22 +193,23 @@ private function processFile(File $file, array $loggingContext) :void
         $fileId = $file->getId();
         $fileset = $file->getFileset();
         $filePath = $file->getPhysicalFilePath();
-        @$fileStream = fopen($filePath, 'r');
+        @$fileHandle = fopen($filePath, 'r');
 
-        if ($fileStream === false) {
+        if ($fileHandle === false) {
```
mickel1138, line 198: Maybe change this to `!is_resource`

mickel1138:
This was a good effort, but am going to close this for now. Maybe we will do this again another time.

PR #1149, opened 2022-04-12
Added filters to json stats

mickel1138:


src/Repository/DatasetRepository.php
```diff
@@ -66,15 +83,32 @@ public function countRegistered()
...
 
-        if ($this->fundingOrgFilter->isActive()) {
+        if ($accepted === true) {
+            $qb
+            ->andWhere('dataset.availabilityStatus IN (:available)')
+            ->setParameter('available', array(
+                DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE,
+                DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE_REMOTELY_HOSTED,
+                DatasetSubmission::AVAILABILITY_STATUS_RESTRICTED,
+                DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE_REMOTELY_HOSTED,
```
praneethpr, line 101: Same parameter in here twice

src/Repository/DatasetRepository.php
```diff
@@ -40,14 +40,31 @@ public function __construct(ManagerRegistry $registry, FundingOrgFilter $funding
...
```
praneethpr, line 57: Same parameter in here twice

fightingtexasaggie (approved):
I think we probably don't need the total-only option as the total is in the json regardless, but seems good.

praneethpr (approved):

PR #1150, opened 2022-04-13
Added check for if published or not

praneethpr:


fightingtexasaggie (approved):
Works fine. Possible improvement: Should we even display the InformationProducts tab at all if there's not at least one, or maybe grayed out, or maybe a count to the left of the tabs?

PR #1152, opened 2022-04-14
Feature/pelagos 4778 add additional logging the filer

fightingtexasaggie:


src/MessageHandler/DatasetSubmissionFilerHandler.php
```diff
@@ -142,16 +143,17 @@ public function __invoke(DatasetSubmissionFiler $datasetSubmissionFiler)
                 $filesInfo[$file->getId()]['filePathName'] = $file->getFilePathName();
                 $filesInfo[$file->getId()]['physicalFilePath'] = $file->getPhysicalFilePath();
             }
-            $this->logger->info('Zipfile opened: ' . $destinationPath);
+            $this->logger->info('Zipfile opened: ' . $destinationPath, array_merge($loggingContext, array('PHP_memory_usage' => memory_get_usage())));
```
mickel1138, line 146: Maybe you want to do this outside this function? Now it will only be used in this particular log. Which may be intentional?

src/MessageHandler/ZipDatasetFilesHandler.php
```diff
@@ -93,7 +93,7 @@ public function __construct(
      */
     public function __invoke(ZipDatasetFiles $zipDatasetFiles)
     {
-        $this->logger->info('Zip worker starting');
+        $this->logger->info('Zip worker starting (ZipDatasetFilesHandler)');
```
mickel1138, line 96: Might as well change the entire text here, this was a bad cut and paste.

src/MessageHandler/DatasetSubmissionFilerHandler.php
fightingtexasaggie: The intent was to capture any change in memory consumption, so the memory_get_usage has to be called right before and after each file.  I also called after zipfile->finish to ensure any allocated memory is released, or to otherwise try and detect memory leaks. I figured in the capture, the better, until we get to the bottom of what's happening.

src/MessageHandler/ZipDatasetFilesHandler.php
fightingtexasaggie: Ok, made the text a bit better, and added the PID for the worker in the messages.

mickel1138 (approved):
Good improvements

PR #1155, opened 2022-04-19
Feature/pelagos 4802 add cold storage attributes to dataset dataset submission entity

fightingtexasaggie:


src/Entity/DatasetSubmission.php
```diff
@@ -954,6 +954,20 @@ class DatasetSubmission extends Entity
      */
     private $datasetLinks;
 
+    /**
+     * For cold-stored, the total unpacked bytecount.
+     *
+     * @ORM\Column(type="integer", nullable=true)
```
fightingtexasaggie, line 960: int not big enough, need bigint!

src/Entity/DatasetSubmission.php
```diff
@@ -2783,4 +2805,26 @@ public function isRemotelyHosted(): bool
         }
         return $isMarked;
     }
+
+    public function getColdStorageTotalUnpackedSize(): ?int
```
mickel1138, line 2809: Need docblocks... (use phpdocblocker plugin for VS)

tests/unit/Entity/DatasetSubmissionTest.php
```diff
@@ -661,10 +673,10 @@ public function testGetTemporalExtentDescChoices()
     public function testGetRestrictionsChoices()
     {
         $restrictionsChoices = DatasetSubmission::getRestrictionsChoices();
-        $this->assertInternalType('array', $restrictionsChoices);
+        $this->assertIsArray($restrictionsChoices);
         foreach ($restrictionsChoices as $index => $value) {
-            $this->assertInternalType('string', $index);
-            $this->assertInternalType('string', $value);
+            $this->assertIsString($index);
```
mickel1138, line 678: Glad these warnings are fixed. Good timing for this.

src/Entity/DatasetSubmission.php
mickel1138: I suggest `ludicrousint`.

src/Entity/DatasetSubmission.php
```diff
@@ -2811,22 +2806,38 @@ public function isRemotelyHosted(): bool
...
+    /**
+     * @return integer|null
+     */
     public function getColdStorageTotalUnpackedSize(): ?int
     {
         return $this->coldStorageTotalUnpackedSize;
     }
 
+    /**
+     * @param integer|null $coldStorageTotalUnpackedSize Aggregate bytecount of all cold-stored files in a dataset, as unpacked.
```
mickel1138, line 2818: At a minimum need a description of the function too.

migrations/Version20220419142744.php
```diff
@@ -0,0 +1,38 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220419142744 extends AbstractMigration
+{
+    public function getDescription() : string
+    {
+        return '';
```
mickel1138, line 17: Add a description

src/Entity/DatasetSubmission.php
```diff
@@ -2783,4 +2805,50 @@ public function isRemotelyHosted(): bool
...
+    {
+        return $this->coldStorageTotalUnpackedCount;
+    }
+
+    /**
+     * Setter for total cold-storded unpacked file count.
+     *
+     * @param integer|null
+     *
+     * @return integer|null $coldStorageTotalUnpackedSize Aggregate filecount of all cold-stored files in a dataset, as unpacked.
```
mickel1138, line 2847: The return here is `self`

src/Entity/DatasetSubmission.php
```diff
@@ -2783,4 +2805,50 @@ public function isRemotelyHosted(): bool
...
+    {
+        return $this->coldStorageTotalUnpackedSize;
+    }
+
+    /**
+     * Setter for total cold-stored unpacked size, in bytes.
+     *
+     * @param integer|null $coldStorageTotalUnpackedSize Aggregate bytecount of all cold-stored files in a dataset, as unpacked.
+     *
+     * @return integer|null
```
mickel1138, line 2824: Wrong return type here.

src/Entity/DatasetSubmission.php
```diff
@@ -954,6 +954,20 @@ class DatasetSubmission extends Entity
      */
     private $datasetLinks;
 
+    /**
+     * For cold-stored, the total unpacked filecount (not dirs).
+     *
+     * @ORM\Column(type="integer", nullable=true)
```
mickel1138, line 960: The type could also just be text, there may not be a real need to set this to int or bigint unless it a sequence.

src/Entity/DatasetSubmission.php
fightingtexasaggie: thanks...will fix

migrations/Version20220419142744.php
fightingtexasaggie: ...my bad

src/Entity/DatasetSubmission.php
mickel1138: https://github.com/griidc/pelagos/blob/develop/src/Repository/DatasetRepository.php#L95

mickel1138: Then should this type here be bigint?

mickel1138: This it not file size but count. But in the case of filesize, I agree we should use bigInt to avoid casting.

fightingtexasaggie: I can see a future need to get average, max, min filecounts in a report, so I'd rather leave it numeric.

src/Form/DatasetSubmissionType.php
```diff
@@ -273,6 +273,16 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'mapped' => false,
                 'required' => false,
             ))
+            ->add('coldStorageTotalUnpackedCount', Type\IntegerType::class, array(
+                'label' => 'Cold Storage Total Unpacked File Count',
+                'mapped' => false,
```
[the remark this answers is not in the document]
mickel1138: We have set all the other ones to False too. It is set to false when the form is being processed, because these items on not usually on the submission form.

mickel1138 (approved):

PR #1157, opened 2022-04-20
Feature/pelagos 4803 update dataland coldstorage

fightingtexasaggie:
To test, merge in 4802 then use dataset-review to set size/count and view in dataland. (If for some reason you want to use the cold-storage command instead to set the count/size, make sure and re-accept in dataset-review.)


templates/Dataland/v2/download-card.html.twig
```diff
@@ -28,14 +54,16 @@
...
                         {{ dataset.datasetSubmission.distributionFormatName }}
                     </p>
-                    <hr>
-                    <p class="card-text text-dark">
-                        <a href="#individual-files">Individual Files
-                            <img src="{{ asset('build/images/down-arrow.svg') }}"><br />
-                        </a>
-                    </p>
-                {% endif %}
+                    {% if dataset.availabilityStatus == constant('App\\Entity\\DatasetSubmission::AVAILABILITY_STATUS_PUBLICLY_AVAILABLE') %}
```
praneethpr, line 57: I don't think this check is necessary as it is the same as line 1

fightingtexasaggie: It's there in the event the dataset is remotely hosted. We don't want to show the file manager link in that case.

mickel1138: I agree, this line is not needed.

mickel1138 (approved):
Looks ok.

PR #1158, opened 2022-04-20
Add default to twig for possible null submission.

fightingtexasaggie:
This fixes broken dataland for dif-only datasets, but this might only really show up in dev mode. I noticed this in testing 4803 on a new dif.


mickel1138:
I'm not a big fan of these changes that only affect DEV and don't have a production problem associated with them.
That said, the changes are mild, so we can add these if we need to.

PR #1159, opened 2022-04-20
Feature/pelagos 4805 improve ip list

mickel1138:


assets/css/template.css
```diff
@@ -503,3 +503,11 @@ div#header #griidc-logo-wrapper {
 table.datasets td {
     padding: 10px !important;
 }
+
+.dx-header-filter {
+    color: rgba(255, 0, 0, 1)!important;
```
mickel1138, line 508: This will be changed with PO's input.

PR #1161, opened 2022-04-21
Required for new Composer version.

fightingtexasaggie:
We were on 2.0.14. Going to the current 2.3.5 requires this change to composer.json.


mickel1138:
Why was this done?

mickel1138 (approved):
Verified.

PR #1164, opened 2022-04-22
Feature/pelagos 4807 issuetrackerlink on dataland

fightingtexasaggie:


config/services.yaml
```diff
@@ -164,6 +164,10 @@ services:
             string $downloadBaseDir: '%env(DOWNLOAD_BASE_DIRECTORY)%'
             string $downloadBaseUrl: '%env(DOWNLOAD_BASE_URL)%'
 
+    App\Controller\UI\DatalandController:
+        bind:
+            string $issueTrackingBaseUrl: '%env(ISSUE_TRACKING_BASE_URL)%'
```
praneethpr, line 169: I don't this is the right link, wouldn't work. We need to use `https://griidc.atlassian.net/browse/` as the prefix for the URL

src/Controller/UI/DatalandController.php
```diff
@@ -49,18 +49,26 @@ class DatalandController extends AbstractController
...
+     * @var string
+     */
+    protected $issueTrackingBaseUrl;
+
     /**
      * Dataland Controller constructor.
      *
      * @param EntityHandler $entityHandler The Entity Handler.
      * @param Geometry      $geoUtil       The Geomtery Util.
      * @param Metadata      $metadataUtil  The Metadata Util.
```
praneethpr, line 64: Missing annotation for the new param

config/services.yaml
fightingtexasaggie: yeah, I saw that in our .env we have it set to ISSUE_TRACKING_BASE_URL='https://griidc.atlassian.net/issues' and figured it just needed updating. Are they 2 different things?

templates/Dataland/v2/ticketing-system-card.html.twig
```diff
@@ -0,0 +1,20 @@
+{% if is_granted('ROLE_DATA_REPOSITORY_MANAGER') %}
+    <div class="py-2">
+        <div class="card text-center text-light">
+            <div class="card-header bg-chart-primary">
+                JIRA Issue
+            </div>
+            <div class="card-body">
+                <p class="card-text text-dark">
+                    {% if dataset.issueTrackingTicket %}
+                        <img src="{{ asset('build/images/up-right-arrow-icon.svg') }}">&nbsp;
```
mickel1138, line 10: Instead of using this image/icon, use Font Awesome.

templates/Dataland/v2/ticketing-system-card.html.twig
```diff
@@ -0,0 +1,20 @@
+{% if is_granted('ROLE_DATA_REPOSITORY_MANAGER') %}
+    <div class="py-2">
+        <div class="card text-center text-light">
+            <div class="card-header bg-chart-primary">
+                JIRA Issue
+            </div>
+            <div class="card-body">
+                <p class="card-text text-dark">
+                    {% if dataset.issueTrackingTicket %}
```
mickel1138, line 9: I don't like the empty block, I think it should have a N/A (or so) placeholder.

config/services.yaml
fightingtexasaggie: I updated the .env accordingly so the correct default is used. Evidently, at some point in the past Atlassian changed /issue to /browser, I guess.

templates/Dataland/v2/ticketing-system-card.html.twig
fightingtexasaggie: Yeah, I agree N/A would look better than leaving it empty.  I'll ask William as he's the one that mentioned to leave it empty.

mickel1138 (approved):
LGTM

praneethpr (approved):

PR #1165, opened 2022-04-25
Moved IF block up higher, restored top IF

mickel1138:


fightingtexasaggie (approved):
Looks good. R4.x260.000:0110 Is an available remotely hosted file dataset, for a quick verification.

PR #1170, opened 2022-05-02
Added check for descriptors

praneethpr:


assets/js/vue/components/information-product/InformationProductForm.vue
```diff
@@ -370,8 +370,8 @@ export default {
         ).then(() => {
           this.ipSuccessModalText = 'Information Product is Updated!';
           this.ipCreatedSuccessModal = true;
-        }).catch(() => {
-          this.errorMessage = 'Unable to update Information Product';
+        }).catch((response) => {
```
mickel1138, line 373: Maybe this should be name error, because that's what is being return, a struct based on class Error. Which by itself also has a response attribute in it.

src/Controller/Api/InformationProductController.php
```diff
@@ -482,4 +486,23 @@ private function deleteFile(File $file, MessageBusInterface $messageBus) : void
...
+     * @param array $idsSelected
+     * @param array $descriptorsToBeAdded
+     *
+     * @return void
+     */
+    private function checkIfDescriptorExists(array $idsSelected, array $descriptorsToBeAdded)
+    {
+        if (count($descriptorsToBeAdded) !== count($idsSelected)) {
+            foreach ($descriptorsToBeAdded as $descriptor) {
+                if (in_array($descriptor->getId(), $idsSelected)) {
```
mickel1138, line 502: You probably wanna check here if it's NOT in the array.

fightingtexasaggie (approved):
I tested by filling the form, then deleting a value in the DB. I think this is of very low likelihood to ever be encountered, but the catch works.

mickel1138 (approved):
GTG

PR #1174, opened 2022-05-05
Bump flex to ^1.17

em411:
https://symfony.com/blog/the-old-flex-infrastructure-is-shutting-down


mickel1138:
Hi @em411, thanks for you contribution. If you would have looked at our lock file, you can see we are already on Flex 1.18, but I understand the reasoning behind this PR, so we'll make sure to update it.

fightingtexasaggie (approved):
LGTM, being that we're already on v1.18.5...

PR #1175, opened 2022-05-09
Feature/pelagos 4827 allow deletion from datastore

fightingtexasaggie:


src/Command/PelagosClearErrorAndFileCommand.php
```diff
@@ -73,9 +85,15 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
                 $delId = $fileToDelete->getId();
                 $this->entityManager->remove($fileToDelete);
-                @unlink($fileToDelete->getPhysicalFilePath());
+                $uploadDir = preg_quote($this->homedirPrefix . '/upload/', '/');
+                if (preg_match("/$uploadDir/", $fileToDelete->getPhysicalFilePath())) {
+                    $physicalFileToDelete = $fileToDelete->getPhysicalFilePath();
+                } else {
+                    $physicalFileToDelete = $this->dataStoreDirectory . '/' . $fileToDelete->getPhysicalFilePath();
+                }
+                @unlink($physicalFileToDelete);
```
mickel1138, line 94: This will directly delete a file from the Datastore. Instead inject the Datastore util, and do exactly what DeleteFileHandler does. Datastore also will have those paths already injected.

config/services.yaml
```diff
@@ -135,6 +135,11 @@ services:
             string $passwordExpiryWarn: "%env(ACCOUNT_PASSWORD_EXPIRATION_WARN)%"
             string $hostName: "%env(HOSTNAME)%"
 
+    App\Command\PelagosClearErrorAndFileCommand:
+        bind:
+            string $homedirPrefix: "%env(HOMEDIR_PREFIX)%"
+            string $dataStoreDirectory: "%env(DATA_STORE_DIRECTORY)%"
+
```
mickel1138, line 142: These won't we needed if you inject Datastore, it can determine is a file is in the store.

src/Command/PelagosClearErrorAndFileCommand.php
```diff
@@ -22,18 +22,30 @@ class PelagosClearErrorAndFileCommand extends Command
...
+
     /**
      * Class constructor for dependency injection.
      *
      * @param EntityManagerInterface $entityManager A Doctrine EntityManager.
      */
-    public function __construct(
-        EntityManagerInterface $entityManager
-    ) {
+    public function __construct(EntityManagerInterface $entityManager, string $homedirPrefix, string $dataStoreDirectory)
```
mickel1138, line 44: Inject datastore here instead of the paths.

fightingtexasaggie: still need homedirprefix, but removed datastore path. I still might move that section out to a util.

mickel1138: A bit more expensive, but you can also use 'has' from data store.

src/Command/PelagosClearErrorAndFileCommand.php
```diff
@@ -102,4 +102,22 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+     * @param Datafile $file
+     * @param string   $uploadDir injected
+     *
+     * @return string
+     */
+    protected function getStorageLocation($file)
+    {
+        $uploadDir = preg_quote($this->homedirPrefix . '/upload/', '/');
+        if (preg_match("/$uploadDir/", $file->getPhysicalFilePath())) {
+            return 'uploaddir';
```
mickel1138: I don't like using words, maybe if they were constants, it's probably easier to return a boolean, and that can also be done with the `datastore->has` method. Which may be a bit more expensive, but this doesn't run that often.

src/Command/PelagosClearErrorAndFileCommand.php
```diff
@@ -71,11 +84,16 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         if ($fileToDelete instanceof File) {
             try {
-                $delId = $fileToDelete->getId();
+                $fileId = $fileToDelete->getId();
                 $this->entityManager->remove($fileToDelete);
-                @unlink($fileToDelete->getPhysicalFilePath());
+                $physicalFileToDelete = $fileToDelete->getPhysicalFilePath();
+                if ($this->datastore->getStorageLocation($fileToDelete) === $this->datastore::STORED_IN_DATASTORE) {
```
mickel1138, line 90:
```suggestion
                if ($this->datastore->has($fileToDelete)) {
```

src/Util/Datastore.php
```diff
@@ -178,4 +197,21 @@ public function deleteDir(string $dirPath): bool
...
+     */
+    public function getStorageLocation($file)
+    {
+        $uploadDir = preg_quote($this->homedirPrefix . '/upload/', '/');
+        if (preg_match("/$uploadDir/", $file->getPhysicalFilePath())) {
+            return $this::STORED_IN_UPLOADS;
+        } else {
+            return $this::STORED_IN_DATASTORE;
+        }
+    }
```
mickel1138, line 216: Not needed if using `has`

mickel1138 (approved):
Reviewed, seen working.

PR #1176, opened 2022-05-11
Added issue ticket to DIF

mickel1138:
Changed DIF API to serializer to allow to get virtual property.
Changed javascript accordingly.
Added tracking ticket to twig.

### DIFF
--- a/src/Controller/Api/DIFController.php
+++ b/src/Controller/Api/DIFController.php
@@ -2,29 +2,23 @@
 
 namespace App\Controller\Api;
 
-use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\Form\FormInterface;
-use Symfony\Component\Routing\Annotation\Route;
-
-use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
-use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
-
-use Nelmio\ApiDocBundle\Annotation\Operation;
-use Nelmio\ApiDocBundle\Annotation\Model;
-use Swagger\Annotations as SWG;
-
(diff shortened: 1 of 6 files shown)


src/Controller/Api/DIFController.php
```diff
@@ -467,7 +465,11 @@ public function postAction(Request $request, EntityEventDispatcher $entityEventD
      */
     public function putAction(int $id, Request $request)
     {
-        $this->handleUpdate(DIFType::class, DIF::class, $id, $request, 'PUT');
+        /** @var DIF $dif */
+        $dif = $this->handleUpdate(DIFType::class, DIF::class, $id, $request, 'PUT');
+        // Update the Dataset too because Issue Tracker Ticket.
+        $this->entityHandler->update($dif->getDataset());
```
fightingtexasaggie, line 471: Before we update, should we check if the user is a DRPM, and if not, ignore/remove the Jira ticket from the dif, just in case they managed to maliciously inject it in their PUT/PATCH action?

fightingtexasaggie (approved):
I don't think users being able to set the ticket if they craft a custom post outside of the UI is that big a deal. They're our users anyway. If it was a public non-auth form, it would be more an issue, so I'm good with this as-is, although wanted to bring it up in case you felt different.

mickel1138:
> I don't think users being able to set the ticket if they craft a custom post outside of the UI is that big a deal. They're our users anyway. If it was a public non-auth form, it would be more an issue, so I'm good with this as-is, although wanted to bring it up in case you felt different.

Even if they had the number, there is nothing they can do with it.

src/Controller/Api/DIFController.php
mickel1138: We would have to very deliberately check if someone was trying to change the Jira ticket. There are other options, but I feel it's not neccesary.

praneethpr (approved):

PR #1178, opened 2022-05-19
Added easy admin crud for funding org

praneethpr:


src/Entity/Entity.php
```diff
@@ -157,6 +157,22 @@ public function getCreator()
...
+     *
+     * @return string.
+     */
+    public function getCreatorName(): string
+    {
+        $creatorName = '';
+        if ($this->getCreator()) {
+            $creatorName = $this->getCreator()->getFirstName() . ' ' . $this->getCreator()->getLastName();
+        }
+        return $creatorName;
```
mickel1138, line 173:
```suggestion
        return $creatorName ?? '';
```

praneethpr: I think initializing the variable to a string is better so that we dont get the variable not defined warning

fightingtexasaggie (dismissed):
Ok, since the FC will be taken care of elsewhere, everything else looks solid. I also tested the delete without linking FCs and was able to delete as expected.

src/Controller/Admin/DataRepositoryCrudController.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace App\Controller\Admin;
+
+use App\Entity\DataRepository;
+use EasyCorp\Bundle\EasyAdminBundle\Controller\AbstractCrudController;
+use EasyCorp\Bundle\EasyAdminBundle\Field\IdField;
+use EasyCorp\Bundle\EasyAdminBundle\Field\TextField;
+
+class DataRepositoryCrudController extends AbstractCrudController
```
mickel1138, line 10: Needed?

mickel1138 (approved):
LGTM

PR #1182, opened 2022-05-31
Added e-mail check

mickel1138:
Try to get person from e-mail if possible,


fightingtexasaggie (dismissed):
Seems to work as expected. I tested that username (not email) was passed to the call to the griidc-ingest service, and that seems to workfine too.

praneethpr (approved):
LGTM! Works as expected

PR #1188, opened 2022-06-07
Changed date picker to DevExtreme

mickel1138:


praneethpr (dismissed):

fightingtexasaggie (commented):
It looks pretty good but I'd like to suggest alignment changes in the small and mid-size views. As you can see in the attached screenshots, the fields don't quite line up. Also, I think more people expect month/day/year than year-month-day (although the latter makes so much more sense imho)
Screenshot from 2022-06-08 09-46-53
Screenshot from 2022-06-08 09-47-15

praneethpr (approved):

praneethpr (dismissed):

fightingtexasaggie (approved):
Alignment is cleared up. I'm a little concerned how it's counter-intuitive to edit a date...can't edit a single digit. If this can be turned off, that could increase usability.

fightingtexasaggie (approved):
still works post-merge ;-)

PR #1192, opened 2022-06-09
Misc/symfony 5 deprecations

mickel1138:


src/Entity/DataCenter.php
```diff
@@ -188,7 +188,7 @@ class DataCenter extends Entity
      *
      * @Assert\Email(
      *     message="Email address is invalid",
-     *     strict=true
+     *     mode="strict"
```
fightingtexasaggie, line 191: 👍🏼 🌮

fightingtexasaggie (approved):
discussed and reviewed, seems fine

PR #1196, opened 2022-06-15
Using Simple Query for phrase search

mickel1138:


praneethpr (approved):
Tested multiple scenarios, works as expected

PR #1200, opened 2022-06-22
PSR-12

mickel1138:


fightingtexasaggie (approved):
All changes are either () related, extra lines (most), or missing lines after <?php, or method() calls. XML export checked out so I guess using new tidy as new tidy() works now.

praneethpr (approved):
It looks good! But let us merge this at last, because it would be become really difficult in figuring out if it is a real change vs psr12 change

PR #1203, opened 2022-06-24
Feature/pelagos 4855 mk symfony cmd jira bkfill tool

fightingtexasaggie:


src/Command/SetIssueTrackerTicketCommand.php
```diff
@@ -0,0 +1,109 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+use App\Entity\DatasetSubmission;
+use App\Entity\Dataset;
+
+/**
+ * This command sets the issue ticket on a datset (submisison).
+ *
```
mickel1138, line 19: this line not needed

src/Command/SetIssueTrackerTicketCommand.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        $this
+            ->setDescription('Sets the issue tracker ticket on a dataset.')
+            ->addOption('csvfile', null, InputOption::VALUE_REQUIRED, 'Filename of csv containing UDI, ticket')
+            ;
+    }
+
+    /**
+     * Symfony command execution section.
+
```
mickel1138, line 63: extra line

src/Command/SetIssueTrackerTicketCommand.php
```diff
@@ -0,0 +1,109 @@
...
+                if (!($datasetSubmission instanceof DatasetSubmission)) {
+                    $io->error('Could not find Dataset Submission in dataset ' . $udi);
+                    continue;
+                } else {
+                    $io->note("Setting Issue Tracking Ticket $issueTrackingTicket for Dataset $udi.");
+                    if ($issueTrackingTicket === '') {
+                        $datasetSubmission->setIssueTrackingTicket(null);
+                    } else {
+                        $datasetSubmission->setIssueTrackingTicket($issueTrackingTicket);
+                    }
```
mickel1138, line 100: The issueTrackingTicket is set on the Dataset, not the submission, that function is just a reach-back to setter.

So just set it on the dataset, checking for a submission is not needed.

src/Command/SetIssueTrackerTicketCommand.php
```diff
@@ -0,0 +1,109 @@
...
+                    $this->entityManager->flush();
```
mickel1138, line 101: Maybe flush outside the loop?

src/Command/SetIssueTrackerTicketCommand.php
fightingtexasaggie: good catch...also this will allow setting a ticket for a dif-only.

src/Command/SetIssueTrackerTicketCommand.php
```diff
@@ -0,0 +1,102 @@
...
+                $udi = trim($data[0]);
+                $issueTrackingTicket = trim($data[1]);
+                $dataset = $this->entityManager->getRepository(Dataset::class)->findOneBy(array('udi' => $udi));
+                if (!($dataset instanceof Dataset)) {
+                    $io->warning("Could not find a dataset with UDI ($udi)");
+                    continue;
+                } else {
+                    if ($issueTrackingTicket === '') {
+                        $io->note("Removing Issue Tracking Ticket for Dataset $udi.");
+                        $dataset->setIssueTrackingTicket(null);
```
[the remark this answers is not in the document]
fightingtexasaggie: That was my intent, to allow a previous value to be cleared by leaving it blank in the CSV. I'll ask the PO to be sure.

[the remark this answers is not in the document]
fightingtexasaggie: I asked the PO and he said that if nothing is present in the ticket field, populate the value with nothing/null, overwriting the potential existing value with the blank.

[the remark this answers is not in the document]
fightingtexasaggie: PO decided it was best to just throw an error for a CSV entry missing the ticket number instead of assuming the intention to null out an existing value. DRPMs will have to use the interface to null out an existing value instead. The tool will still replace an existing value with something else.

src/Command/SetIssueTrackerTicketCommand.php
```diff
@@ -0,0 +1,101 @@
...
+        if (($fileHandle = fopen($filename, "r")) !== false) {
+            while (($data = fgetcsv($fileHandle, 100, ",")) !== false) {
+                $udi = trim($data[0]);
+                $issueTrackingTicket = trim($data[1]);
+                $dataset = $this->entityManager->getRepository(Dataset::class)->findOneBy(array('udi' => $udi));
+                if (!($dataset instanceof Dataset)) {
+                    $io->warning("Could not find a dataset with UDI ($udi)");
+                    continue;
+                } else {
+                    if ($issueTrackingTicket === '') {
```
mickel1138, line 87: It's safer to use `empty`, any maybe make this a positive comparison, with the else being the failure.

mickel1138 (dismissed):
LGTM

PR #1204, opened 2022-06-24
Fix for json-schema security issue.

mickel1138:
Updated to node-sass 7 and sass-loader 12

json-schema is now at 0.4.0


fightingtexasaggie (approved):
Built fine with test merge into the symfony 5 branch. Everything seems to work.

PR #1206, opened 2022-06-30
Added search terms report for the search app

praneethpr:


fightingtexasaggie (approved):
As long as we're not concerned with the session id in this ticket, and I think we agreed this is the case, I think this is ready.

PR #1211, opened 2022-07-06
Changed functionality of clear button to show all datasets

praneethpr:


fightingtexasaggie (approved):
LGTM, one slight comment I'll make is it now shows the previous results for a second after clearing, before showing the new set of results, but I don't really think it's an issue.

PR #1221, opened 2022-07-13
Feature/pelagos 4884 create ip search controller

praneethpr:


fightingtexasaggie (approved):
Tested and looks like everything is working as intended. I sorta follow it.

PR #1224, opened 2022-07-15
Allowed for null in TokenInterface

mickel1138:


fightingtexasaggie (approved):
Tested, logout already logged out now works without error.

PR #1227, opened 2022-07-19
Feature/pelagos 4893 fix ip indexer

mickel1138:


fightingtexasaggie (approved):
Now is working as expected.

assets/scss/data-land.scss
```diff
@@ -11,6 +11,6 @@
   }
 }
 
-.supplemental-info-description {
+.pre-line-white-space {
```
praneethpr, line 14: Don't we usually merge develop into the epic branch rather than a feature branch that is going into the epic

praneethpr (approved):

PR #1236, opened 2022-07-27
Feaature/pelagos 4899 add multi index

mickel1138:


src/Search/ElasticaToModelTransformerCollection.php
```diff
@@ -0,0 +1,101 @@
+<?php
+
+namespace App\Search;
```
mickel1138, line 3: Because we didn't really write this file, I do want move this to up the `Transformer` namespace.

praneethpr (approved):

fightingtexasaggie (approved):

PR #1237, opened 2022-08-01
Feature/pelagos 4904 separate collection date range allowing singular usage

fightingtexasaggie:
Since we were already using a lte and gte query internally, I really just had to pass each condition separately.


mickel1138 (approved):
Works as expected.

praneethpr (approved):

PR #1242, opened 2022-08-16
Release/6.31.0

fightingtexasaggie:
When testing this on PHP8.0, make sure all the requisite php.ini vars are set. It does not require php 8.0 though. R1.x132.134:0066 has an ERDDAP link for testing.


fightingtexasaggie (commented):
reviewed and tested on griidc-demo. It's good to go.

mickel1138 (approved):

PR #1245, opened 2022-08-17
Create new dif-status aware filter.

fightingtexasaggie:
I wish there was a way to pass in more than one callbacks in the fos_elastica.yaml, but I'm not seeing it so created kind of a wrapper callback that adds the constraint of approved-only DIFs IFF there's not already a submission.

### DIFF
--- a/src/Util/DatasetIndexFilter.php
+++ b/src/Util/DatasetIndexFilter.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace App\Util;
+
+use App\Entity\Dataset;
+use App\Entity\DIF;
+use App\Util\FundingOrgFilter;
+
+/**
+ * A utility to determine if a Dataset should be indexed.
+ */
+class DatasetIndexFilter
+{
+    /**
+     * Returns true if passed dataset should be indexed.
+     *
+     * @param \App\Entity\Dataset $dataset The dataset to considered for indexing.
+     *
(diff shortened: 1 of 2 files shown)


src/Util/DatasetIndexFilter.php
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Returns true if passed dataset should be indexed.
+     *
+     * @param \App\Entity\Dataset $dataset The dataset to considered for indexing.
+     *
+     * @return boolean If the dataset should be indexed, which is all but datasets of only an unapproved DIF.
+     */
+    public static function indexable(Dataset $dataset): bool
+    {
+        return (FundingOrgFilter::canIndex($dataset) and (true === $dataset->hasDatasetSubmission() or (true === $dataset->hasDif() and $dataset->getDif()->getStatus() === DIF::STATUS_APPROVED)));
```
mickel1138, line 23: Sorry, Yoda notation in no longer cool, and confusing it is.

mickel1138:
Does this also remove the dataset from the index when a DIF is unlocked?

fightingtexasaggie:
I confirmed that this does remove an unlocked DIF from the index. Because of the logic used, this only happens if there is not a submission present though. Once a submission is present, the DIF state doesn't matter.

In summary pertaining to fos elastica, the indexable callback, if false, will remove an existing document.

praneethpr (approved):

PR #1246, opened 2022-08-19
Fixed prod log writing

mickel1138:
Also removed logs and session folders keep file.


fightingtexasaggie (approved):
LGTM, logs now being written.

PR #1250, opened 2022-08-24
IP search now works ok with multi search

mickel1138:


src/Search/Transformer/ElasticaToModelTransformerCollection.php
```diff
@@ -59,7 +59,7 @@ public function transform(array $elasticaObjects)
         $transformed = [];
         foreach ($sorted as $type => $objects) {
             $transformedObjects = $this->transformers[$type]->transform($objects);
-            $identifierGetter = 'get'.ucfirst($this->transformers[$type]->getIdentifierField());
+            $identifierGetter = 'get' . ucfirst($this->transformers[$type]->getIdentifierField());
```
praneethpr, line 62: Didn't know about ucfirst 👍

src/Controller/UI/InformationProductSearchController.php
```diff
@@ -40,6 +40,8 @@ public function searchForInformationProduct(Request $request, InformationProduct
         $searchOptions->setResearchGroupFilter($researchGroupFilter);
         $searchOptions->setProductTypeDescFilter($productTypeDescFilter);
         $searchOptions->setDigitalTypeDescFilter($digitalTypeDescFilter);
+        $searchOptions->onlyPublishedInformationProducts(true);
+        $searchOptions->setMaxPerPage(1000);
```
praneethpr, line 44: Should we set this back to 10?

assets/js/vue/components/info-search/ResultSet.vue
```diff
@@ -49,14 +49,7 @@ export default {
...
     resultCount() {
-      let resultCount = 0;
-      const informationProducts = this.results.informationProducts ?? [];
-      informationProducts.forEach((informationProduct) => {
-        if (informationProduct.published) {
-          resultCount += 1;
-        }
-      });
-      return resultCount;
+      return this.results.count;
```
praneethpr, line 52: We can remove this, as it is not computing anything and we can display `this.results.count`

src/Controller/UI/InformationProductSearchController.php
fightingtexasaggie: until we have pagination, and since we don't have that many info products currently, I think it might make sense to just show them all.

fightingtexasaggie (approved):
nice and clean too

PR #1259, opened 2022-09-06
improved ResearchGroup Landing API

mickel1138:


assets/js/vue/components/research-group/PublicationsTab.vue
```diff
@@ -18,7 +18,7 @@ export default {
       const publications = [];
       const publicationId = [];
       this.datasets.forEach((dataset) => {
-        dataset.publications.forEach((publication) => {
+        dataset.datasetPublications.forEach((publication) => {
```
mickel1138, line 21: needs work, still pubs still broken

src/Controller/Api/ResearchGroupController.php
```diff
@@ -485,4 +486,58 @@ public function putLogoAction(int $id, Request $request)
     {
         return $this->putProperty(ResearchGroup::class, $id, 'logo', $request);
     }
+
+    /**
+     * Undocumented function
```
mickel1138, line 491: Need better text.

assets/js/vue/components/research-group/PublicationsTab.vue
praneethpr: Done 👍

src/Util/JsonSerializer.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Creates a response from JSON.
+     *
+     * @return Response
+     */
+    public function createJsonResponse(): Response
+    {
+        $header = array('Content-Type', 'application/json');
+        return new Response($this->json, Response::HTTP_OK, $header);
```
praneethpr, line 78: We can remove $header variable as it is used only once

src/Util/JsonSerializer.php
```diff
@@ -0,0 +1,80 @@
...
+     * @var SerializerInterface $serializer
+     */
+    private $serializer;
+
+    /**
+     * The json string, containing the serialized string.
+     *
+     * @var string
+     */
+    private $json = '{}';
```
praneethpr, line 26: I'm not sure if this would work in the version of PHP we are in. 
`
```suggestion
    private string $json = '{}';
```

assets/js/vue/components/search/DatasetRow.vue
```diff
@@ -56,6 +58,24 @@ export default {
...
+        return this.datasetRowData.fileSize ?? xbytes(this.datasetRowData.datasetSubmission.datasetFileSize);
+      }
+      return this.datasetRowData.fileSize ?? null;
+    },
+    coldStorage() {
+      if (this.datasetRowData.datasetSubmission) {
+        return this.datasetRowData.coldStorage ?? this.datasetRowData.datasetSubmission.coldStorage;
+      }
+      return this.datasetRowData.coldStorage ?? false;
+    },
```
praneethpr, line 78: It is a bit redundant/confusing, we should rewrite this into a simple if-else?

mickel1138: This should go away in a future ticket.

src/Util/JsonSerializer.php
mickel1138: No complaints from php 8+ when using 3v4l, but I like the type hint.

fightingtexasaggie (approved):
Ok, it all seems to be working, and way more memory friendly now.

PR #1260, opened 2022-09-08
Only show citation if Accepted

mickel1138:


templates/Dataland/v2/index.html.twig
```diff
@@ -69,13 +69,15 @@
                         </div>
                         <hr>
                     {% endif %}
+                    {% if dataset.datasetStatus == constant('App\\Entity\\Dataset::DATASET_STATUS_SUBMITTED') %}
```
fightingtexasaggie, line 72: indent?

fightingtexasaggie: I also don't think we need to include DATASET_STATUS_IN_REVIEW, DATASET_STATUS_ACCEPTED as I think the PO is considering this more of a milestone (to include subsequent states) in the dataset than a specific state.

fightingtexasaggie (approved):
I think this makes good sense.

PR #1262, opened 2022-09-09
Feature/pelagos 4962 log additional field

fightingtexasaggie:
I'm using the custom template names to detect subsite, and shortening per the ticket. I don't know if we should we consider a friendly name .env var instead, or if this is alright as is.

### DIFF
--- a/src/Controller/UI/SearchPageController.php
+++ b/src/Controller/UI/SearchPageController.php
@@ -39,9 +39,14 @@ class SearchPageController extends AbstractController
     {
         $this->logActionItemEventDispatcher = $logActionItemEventDispatcher;
         if (empty($customTemplate)) {
+            // If custom template is not set, subSite is 'GRIIDC'.
             $this->subSite = 'GRIIDC';
+        } elseif ($customTemplate == 'nas-grp-base.html.twig') {
+            $this->subSite = 'GRP';
+        } elseif ($customTemplate == 'hri-base.html.twig') {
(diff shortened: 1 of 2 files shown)


src/Controller/UI/SearchPageController.php
```diff
@@ -23,14 +23,31 @@ class SearchPageController extends AbstractController
...
+     */
+    protected $subSite;
+
     /**
      * Constructor for this Controller, to set up default services.
      *
      * @param LogActionItemEventDispatcher $logActionItemEventDispatcher The log action item event dispatcher.
      */
-    public function __construct(LogActionItemEventDispatcher $logActionItemEventDispatcher)
+    public function __construct(LogActionItemEventDispatcher $logActionItemEventDispatcher, ?string $customTemplate)
```
mickel1138, line 38: Injectable should be never NULL, or else the framework throws an error, so the ? can go.

src/Controller/UI/SearchPageController.php
```diff
@@ -23,14 +23,31 @@ class SearchPageController extends AbstractController
...
     {
         $this->logActionItemEventDispatcher = $logActionItemEventDispatcher;
+        if (empty($customTemplate)) {
+            // If custom template is not set, subSite is 'GRIIDC'.
```
mickel1138, line 42: Maybe this should be always set to something, but that would be an improvement we might consider some day.

praneethpr (approved):

PR #1263, opened 2022-09-09
Feature/pelagos 4963 add additional field to the search terms report

fightingtexasaggie:
To test, make sure and disable the getGriidcStaff() call on line 207 of SearchTermsReportController.php

### DIFF
--- a/src/Controller/UI/SearchTermsReportController.php
+++ b/src/Controller/UI/SearchTermsReportController.php
@@ -186,7 +186,7 @@ class SearchTermsReportController extends ReportController
     {
         //prepare labels
         $labels = array('labels' => array(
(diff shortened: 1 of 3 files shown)


src/Controller/UI/SearchTermsReportController.php
```diff
@@ -226,10 +226,15 @@ protected function getDatav2(): array
                 $searchResults['1stScore'] = $result['payLoad']['elasticScoreFirstResult'] ?? '';
             }
 
+            if (empty($result['payLoad']['subSite'])) {
```
mickel1138, line 229: Maybe `key_exists` is safer?

fightingtexasaggie: Per https://www.php.net/manual/en/function.array-key-exists.php "array_key_exists() will search for the keys in the first dimension only. Nested keys in multidimensional arrays will not be found."

praneethpr (approved):

PR #1267, opened 2022-09-13
Added condition to check fields exist before displaying in dataland

praneethpr:


templates/Dataland/v2/detail-cards.html.twig
```diff
@@ -28,6 +28,7 @@
 </div>
 {% include 'Dataland/v2/point-of-contact.html.twig' %}
 {% include 'Dataland/v2/collection-period-card.html.twig' %}
+{% if dataset.datasetSubmission and  dataset.datasetSubmission.themeKeywords %}
```
fightingtexasaggie, line 31: Oops extra space

```suggestion
{% if dataset.datasetSubmission and dataset.datasetSubmission.themeKeywords %}
```

PR #1268, opened 2022-09-14
Moved api request to research group app component, so that api is not…

praneethpr:


fightingtexasaggie (approved):
Reviewed, confirms does not re-reload.

PR #1271, opened 2022-09-14
Add NCEI link to dataland.

fightingtexasaggie:


src/Entity/DatasetSubmission.php
```diff
@@ -2600,6 +2600,27 @@ public function getErdappDatasetLink(): ?DatasetLink
...
+        $datasetLinks = $this->getDatasetLinks()->filter(function (DatasetLink $datasetLink) {
+            return $datasetLink->getName() === DatasetLink::LINK_NAME_CODES["ncei"]["name"];
+        });
+
+        if ($datasetLinks->count() > 0) {
+            $nceiLink = $datasetLinks->first();
+        }
+
+        if ($nceiLink instanceof DatasetLink) {
+            return $nceiLink->getUrl();
```
mickel1138, line 2619: Not sure if we need a two-step function, like erdap, but this is ok. We can change it if we need specifics.

templates/Dataland/v2/status-badge.html.twig
```diff
@@ -17,7 +17,7 @@
     {% if dataset.datasetSubmission and dataset.datasetSubmission.isDatasetFileInColdStorage %}
         <span class="badge badge-coldstorage">Cold Storage</span>
     {% endif %}
-    {% if dataset.datasetSubmission.erddapUrl %}
+    {% if dataset.datasetSubmission.erddapUrl|default %}
```
mickel1138, line 20: Not really part of this ticket, was this breaking?

mickel1138 (dismissed):

templates/Dataland/v2/status-badge.html.twig
mickel1138: This should probably just be undone, as it's causing the merge conflict. It seems this was mitigated somewhere else.

fightingtexasaggie: undone

praneethpr (approved):

PR #1273, opened 2022-09-16
Added expression languge condition for exclude

mickel1138:


fightingtexasaggie (approved):
Tested against about/833 (which has datasets). It seems everthing needed from datasets is there.

PR #1276, opened 2022-09-21
Don't show unapproved difs in RG datasets tab.

fightingtexasaggie:


assets/js/vue/components/research-group/DatasetsTab.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <b-card-text class="vh-100">
-        <DatasetRow :datasetRowData="dataset" v-for="dataset in datasets" v-bind:key="dataset.udi"/>
+        <DatasetRow :datasetRowData="dataset" v-for="dataset in datasets" v-bind:key="dataset.udi" v-if="dataset.dif.status === 2"/>
```
mickel1138, line 3: Suggest to use v-show

praneethpr (dismissed):

src/Entity/Dataset.php
```diff
@@ -112,6 +112,8 @@ class Dataset extends Entity
      *
      * @var DIF
      *
+     * @Serializer\Groups({"card"})
```
mickel1138, line 115: Add card groups to DIF, and also add Group to array of RG controller API

PR #1277, opened 2022-09-21
Simplify display of difs in dataland.

fightingtexasaggie:
I'm not too keen on using (only) dataset.availabilityStatus == constant("App\\Entity\\DatasetSubmission::AVAILABILITY_STATUS_NOT_AVAILABLE" as basis for identifying DIF status datasets, but we're already doing this for the badges. We should probably discuss. Maybe it's alright.

### DIFF
--- a/templates/Dataland/v2/index.html.twig
+++ b/templates/Dataland/v2/index.html.twig
@@ -25,7 +25,7 @@
                         {% include 'Dataland/v2/status-badge.html.twig' %}
                         <h2> {{ dataset.title }} </h2>
                     </div>
-                    {% if dataset.dif.status == constant("App\\Entity\\DIF::STATUS_APPROVED") %}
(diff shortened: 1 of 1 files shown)


praneethpr (approved):

fightingtexasaggie:
Ok, now I'm only testing that a dataset has an approved dif, if so show full headings, otherwise show

templates/Dataland/v2/index.html.twig
```diff
@@ -25,26 +25,48 @@
                         {% include 'Dataland/v2/status-badge.html.twig' %}
                         <h2> {{ dataset.title }} </h2>
                     </div>
-                    <div class="d-flex justify-content-between py-2">
-                        <div>
+                    {% if dataset.datasetsubmissionstatus != constant("App\\Entity\\DatasetSubmission::STATUS_UNSUBMITTED") %}
```
mickel1138, line 28: Was thinking more along the lines of `if dataset.hasSubmittedSubmisson` or something like that.

PR #1280, opened 2022-09-27
Misc/add phpstan

mickel1138:


mickel1138:
You can add the phpstan plug for VScode and it will show you the errors.

phpstan.neon
```diff
@@ -0,0 +1,5 @@
+parameters:
+	level: 1
+	tmpDir: /var/tmp
```
fightingtexasaggie, line 3: The problem with this is that we all generate our own entries with the same filepaths, and by default can't overwrite each other's stuff. For purposes of development, we seem to need separated space since it doesn't guarantee unique file path naming.

fightingtexasaggie: I think a better place for it is our own per-user pelagos/var directories.

fightingtexasaggie, line 3:
```suggestion
	tmpDir: var/tmp
```

fightingtexasaggie, line 3:
```suggestion
	tmpDir: var/tmp
```

mickel1138: It can be a relative path, we could use var/tmp (in pelagos dir)

mickel1138: ok, I just has to scroll down...

fightingtexasaggie (approved):

PR #1287, opened 2022-10-05
Removed unnessesary constraints.

mickel1138:


src/Entity/FundingCycle.php
```diff
@@ -134,9 +132,6 @@ class FundingCycle extends Entity
      * @access protected
      *
      * @ORM\Column(type="date", nullable=true)
-     *
-     * @Assert\DateTime()
-     * @Assert\GreaterThan(propertyPath="startDate", message="End Date must be after Start Date")
```
fightingtexasaggie, line 139: This constraint was requested in Pelagos-3841 back in Aug 2019. Do we need to recreate this logic?

praneethpr (approved):

PR #1295, opened 2022-10-20
Feature/pelagos 5000 symfony command to set erddap dataset links

fightingtexasaggie:
Per discussion with the PO, we'll just run this once so manually re-indexing after the run is fine. (Since it only updates submission, it's not triggering index listener.)

### DIFF
--- a/src/Command/AddDatasetLinkCommand.php
+++ b/src/Command/AddDatasetLinkCommand.php
@@ -79,16 +79,19 @@ class AddDatasetLinkCommand extends Command
         // Since Symfony doesn't allow mandatory options, and arguments are ordered and undescribed, not using.  Instead force options mandatory.
         if (empty($udi) or empty($type) or empty($targetUrl)) {
             $io->error("UDI, type, and URL parameters are not optional.");
+            exit(1);
         }
 
         // Accept only known types, 'NCEI' or 'ERDDAP'.
         if (!in_array($type, array(DatasetLink::LINK_NAME_CODES["erddap"]["name"], DatasetLink::LINK_NAME_CODES["ncei"]["name"]))) {
             $io->error("Please specify either ERDDAP or NCEI.");
+            exit(1);
         }
 
         $dataset = $this->entityManager->getRepository(Dataset::class)->findOneBy(array('udi' => $udi));
         if (!($dataset instanceof Dataset)) {
             $io->error('Could not find a dataset with UDI (' . $udi . ')');
+            exit(1);
         } else {
             // get submission
             $submission = $dataset->getDatasetSubmission();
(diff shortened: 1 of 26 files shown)


src/Command/AddDatasetLinkCommand.php
```diff
@@ -0,0 +1,139 @@
...
+        // Since Symfony doesn't allow mandatory options, and arguments are ordered and undescribed, not using.  Instead force options mandatory.
+        if (empty($udi) or empty($type) or empty($targetUrl)) {
+            $io->error("UDI, type, and URL parameters are not optional.");
+            exit(1);
+        }
+
+        // Accept only known types, 'NCEI' or 'ERDDAP'.
+        if (!in_array($type, array(DatasetLink::LINK_NAME_CODES["erddap"]["name"], DatasetLink::LINK_NAME_CODES["ncei"]["name"]))) {
+            $io->error("Please specify either ERDDAP or NCEI.");
+            exit(1);
```
mickel1138, line 88: It likes return, not exit.

src/Command/AddDatasetLinkCommand.php
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        $dataset = $this->entityManager->getRepository(Dataset::class)->findOneBy(array('udi' => $udi));
+        if (!($dataset instanceof Dataset)) {
+            $io->error('Could not find a dataset with UDI (' . $udi . ')');
+            exit(1);
```
mickel1138, line 94: Same here, use return. Exit is old style.

src/Command/AddDatasetLinkCommand.php
```diff
@@ -0,0 +1,139 @@
...
+                    ($type === DatasetLink::LINK_NAME_CODES["erddap"]["name"] and !empty($submission->getErddapUrl())) or
+                    ($type === DatasetLink::LINK_NAME_CODES["ncei"]["name"] and !empty($submission->getNceiUrl()))
+                ) {
+                    $io->warning("$udi already has link of type $type. Not changing.");
+                } else {
+                    $link = new DatasetLink();
+
+                    // common elements
+                    $link->setProtocol('https');
+                    $link->setCreator($systemPerson);
```
mickel1138, line 110: At least creation date has to be set once too I think.

src/Command/AddDatasetLinkCommand.php
```diff
@@ -0,0 +1,144 @@
...
+                    $this->entityManager->flush();
+                    $io->success("Set $type URL on $udi");
+                }
+            } else {
+                $io->error("$udi has no submission.");
+                return(1);
+            }
+        }
+        $io->success('Command Complete.');
+        return(0);
```
mickel1138, line 142: Return is not a function.
```suggestion
        return 0;
```

mickel1138 (dismissed):
Tested, and works.

fightingtexasaggie:
sorry, someone needs to re-approve this because I accidently merged and reverted something in.

PR #1297, opened 2022-10-24
Feature/pelagos 5016 add system deps to composer json

fightingtexasaggie:
Please let me know if I left anything out, or put anything in you think we don't need.


mickel1138:
This looks great, but the composer lock is not needed. We could do an update and make worth our while. Not sure why you can have lock file?

composer.json
```diff
@@ -9,7 +9,14 @@
         "php": "^7.3.0",
         "ext-ctype": "*",
         "ext-iconv": "*",
+        "ext-intl": "*",
         "ext-json": "*",
+        "ext-ldap": "*",
+        "ext-mbstring": "*",
```
mickel1138, line 15: We don't use MBstring directly, but lost of our vendors do. Not sure if we should keep it here, because Pelagos doesn't require it, but it does need to be installed for the vendor stuff. They however do not specify, and I'm not sure if Composer would break if they had.

composer.json
```diff
@@ -9,7 +9,14 @@
         "php": "^7.3.0",
         "ext-ctype": "*",
         "ext-iconv": "*",
```
mickel1138, line 11: We also don't use iconv, but it is needed here because of some odd polyfill that gets installed.

mickel1138:
I think we should remove what we don't use directly. We can go over them.

mickel1138 (dismissed):

fightingtexasaggie (commented):
seems to need the composer.lock commited, along with requiring symfony/runtime still.

mickel1138 (dismissed):

praneethpr (approved):

PR #1298, opened 2022-10-26
Feature/pelagos 5022 use new security bundle

mickel1138:
E-mail verification, and login throttling can done by the new Security bundle itself. Right now we wrote our own. 
However it would be a lot of work to change these around.


praneethpr (dismissed):
LGTM!

fightingtexasaggie (approved):
Tested this out, including pw change. LGTM

PR #1301, opened 2022-10-28
Feature/pelagos 5023 create first available id generator

fightingtexasaggie:
Placeholder is by far simplier than trying to set .value of the field.

### DIFF
--- a/assets/static/js/entity/ResearchGroup.js
+++ b/assets/static/js/entity/ResearchGroup.js
@@ -40,8 +40,8 @@ $(document).ready(function()
     if ($("form[entityType=\"ResearchGroup\"] #id").val() !== "") {
         $("form[entityType=\"ResearchGroup\"] #id").attr("readonly",true);
     } else {
-        var nextId = $("#id").attr("next-id");
-        $("input[name='id']").val(nextId);
+        var nextId = $("form[entityType=\"ResearchGroup\"] #id").attr("next-id");
+        $("form[entityType=\"ResearchGroup\"] #id").val(nextId);
     }
 
     $("[fundingOrganization]").change(function () {
@@ -111,7 +111,7 @@ function addOptionsByEntity(selectElement, entity, filter)
                 );
             });
         });
-        
+
         this.resolve();
     });
 }
--- a/assets/static/js/entity/ResearchGroup.js
+++ b/assets/static/js/entity/ResearchGroup.js
@@ -39,6 +39,9 @@ $(document).ready(function()
     //Disable RIS ID if not in create mode
     if ($("form[entityType=\"ResearchGroup\"] #id").val() !== "") {
         $("form[entityType=\"ResearchGroup\"] #id").attr("readonly",true);
+    } else {
+        var nextId = $("#id").attr("next-id");
+        $("input[name='id']").val(nextId);
     }
 
     $("[fundingOrganization]").change(function () {
--- a/src/Entity/ResearchGroup.php
+++ b/src/Entity/ResearchGroup.php
@@ -42,6 +42,16 @@ class ResearchGroup extends Entity
     const FRIENDLY_NAME = 'Research Group';
 
     /**
+     * Highest acceptable ID number.
+     */
+    public const MAX_ID = 999;
+
+    /**
+     * Minimum acceptable ID number.
+     */
+    public const MIN_ID = 100;
+
+    /**
      * Name of a research group.
      *
      * @var string $name
--- a/src/Entity/ResearchGroup.php
+++ b/src/Entity/ResearchGroup.php
@@ -49,7 +49,7 @@ class ResearchGroup extends Entity
     /**
      * Minimum acceptable ID number.
      */
-    public const MIN_ID = 100;
+    public const MIN_ID = 1;
 
     /**
      * Name of a research group.
--- a/src/Form/ResearchGroupType.php
+++ b/src/Form/ResearchGroupType.php
@@ -62,7 +62,7 @@ class ResearchGroupType extends AbstractType
         $builder
             ->add('id', TextType::class, array(
                 'label' => 'Research Group ID:',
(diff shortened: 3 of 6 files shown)


src/Repository/ResearchGroupRepository.php
```diff
@@ -81,4 +81,20 @@ public function getResearchGroupsInfo(array $aggregations): array
...
+    /**
+     * Find the lowest-value ID available for use, in a specified range.
+     *
+     * @param int $min Range start.
+     * @param int $max Range end.
+     * @return int First available ID.
+     */
+    public function getNextAvailableId($min, $max)
+    {
+        $researchGroups = $this->findAll();
```
mickel1138, line 94: Here use the querybuilder, then only select ID. In results use Query::Hydrate_Array) and you'll have your array.

src/Repository/ResearchGroupRepository.php
```diff
@@ -81,4 +81,20 @@ public function getResearchGroupsInfo(array $aggregations): array
...
+        foreach ($researchGroups as $researchGroup) {
+            $ids[] = $researchGroup->getId();
+        }
+        return array_values(array_diff(range($min, $max), $ids))[0];
```
mickel1138, line 98: Index 0 may not exist. Reset or Current should work.

src/Form/ResearchGroupType.php
```diff
@@ -33,6 +61,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder
             ->add('id', TextType::class, array(
                 'label' => 'Research Group ID:',
+                'attr' => array('placeholder' => $this->researchGroupRepository->getNextAvailableId(self::MIN_ID, self::MAX_ID)),
```
mickel1138, line 64: This is ok with me, but check the PO. And you then may want to prefix it with: "next avail ID is...."

fightingtexasaggie: Believe me, there is no other way than to use placeholder without getting into a major mess with that current multi-modal form.

src/Repository/ResearchGroupRepository.php
fightingtexasaggie: QUERY::HYDRATE_SCALAR_COLUMN seemed to be the magic.

src/Form/ResearchGroupType.php
mickel1138: It is doable, but it's messy.

Look at the code below, you can find the attribute here, and set it. But it would also cause issues while editing, and this field should locked then.

https://github.com/griidc/pelagos/blob/develop/assets/static/js/entity/ResearchGroup.js#L39-L42

src/Repository/ResearchGroupRepository.php
```diff
@@ -85,16 +85,17 @@ public function getResearchGroupsInfo(array $aggregations): array
     /**
      * Find the lowest-value ID available for use, in a specified range.
      *
-     * @param int $min Range start.
-     * @param int $max Range end.
+     * @param  int $min Range start.
+     * @param  int $max Range end.
```
mickel1138, line 89: Spacing?

assets/static/js/entity/ResearchGroup.js
```diff
@@ -39,6 +39,9 @@ $(document).ready(function()
     //Disable RIS ID if not in create mode
     if ($("form[entityType=\"ResearchGroup\"] #id").val() !== "") {
         $("form[entityType=\"ResearchGroup\"] #id").attr("readonly",true);
+    } else {
+        var nextId = $("#id").attr("next-id");
+        $("input[name='id']").val(nextId);
```
mickel1138, line 44: Use the same selector as above, your can just copy the selector above, and set it to a variable.

src/Entity/ResearchGroup.php
```diff
@@ -41,6 +41,16 @@ class ResearchGroup extends Entity
...
 
+    /**
+     * Highest acceptable ID number.
+     */
+    public const MAX_ID = 999;
+
+    /**
+     * Minimum acceptable ID number.
+     */
+    public const MIN_ID = 100;
```
praneethpr, line 52: Is there a reason why this is not 1

src/Repository/ResearchGroupRepository.php
```diff
@@ -81,4 +81,21 @@ public function getResearchGroupsInfo(array $aggregations): array
...
+     * @return int First available ID.
+     */
+    public function getNextAvailableId($min, $max)
+    {
+        $qb = $this->createQueryBuilder('researchGroup')
+            ->select('researchGroup.id')
+            ->orderBy('researchGroup.id', 'ASC');
+        $existingIds = $qb->getQuery()->getResult(QUERY::HYDRATE_SCALAR_COLUMN);
+        $availableIds = array_values(array_diff(range($min, $max), $existingIds));
+        return reset($availableIds);
```
praneethpr, line 99: Great way to get the first element of the array! Didn't know about this

src/Repository/ResearchGroupRepository.php
```diff
@@ -81,4 +81,21 @@ public function getResearchGroupsInfo(array $aggregations): array
...
     }
+
+    /**
+     * Find the lowest-value ID available for use, in a specified range.
+     *
+     * @param  int $min Range start.
+     * @param  int $max Range end.
+     * @return int First available ID.
+     */
+    public function getNextAvailableId($min, $max)
```
praneethpr, line 92: Missing return type

mickel1138 (dismissed):

src/Repository/ResearchGroupRepository.php
mickel1138: Also docblock line spacing.

praneethpr (dismissed):

mickel1138 (dismissed):

praneethpr (approved):

PR #1302, opened 2022-11-02
Added symfony mailer, removed swift mailer

praneethpr:


config/packages/dev/mailer.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+    mailer:
+        # dsn: 'null://null'
+        envelope:
+            recipients: ['praneeth.pondicherry@tamucc.edu']
```
praneethpr, line 5: No new line, remove test code

src/Util/MailSender.php
```diff
@@ -76,23 +85,15 @@ public function sendEmailMessage(
...
-            ->addPart($emailTwigTemplate->renderBlock('body_text', $mailData), 'text/plain');
-        foreach ($attachments as $attachment) {
-            if (!$attachment instanceof \Swift_Attachment) {
-                throw new \InvalidArgumentException('Attachment is not an instance of Swift_Attachment');
-            }
-            $message->attach($attachment);
-        }
+        $email = (new Email())
+            ->from(new Address($this->fromAddress, $this->fromName))
+            ->to(...$toAddresses)
```
praneethpr, line 91: Do we need name for $toAddress as well?

config/packages/dev/mailer.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
```
praneethpr, line 1: Do we need to create one for test as well? so that all emails goto a single email address?

mickel1138: I'm not sure if this config will use that new style yaml file. But if there is no file in test, the default one is used.

docker-compose.override.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+
+services:
+###> symfony/mailer ###
+  mailer:
+    image: schickling/mailcatcher
+    ports: [1025, 1080]
+###< symfony/mailer ###
```
mickel1138, line 8: I wonder what this file is.

src/Command/ValidateRemotelyHostedLinksCommand.php
```diff
@@ -131,7 +131,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $this->mailer->sendEmailMessage(
                 $this->twig->load('Email/data-repository-managers.error-remotely-hosted.email.twig'),
                 array('errors' => $errors),
-                array('griidc@gomri.org' => 'GRIIDC')
+                array('griidc@gomri.org')
```
mickel1138, line 134: Here we may have to pass in a recipient instead. It called `Address`.

src/Event/EventListener.php
```diff
@@ -151,8 +149,7 @@ protected function sendMailMsg(
             $this->mailer->sendEmailMessage(
                 $twigTemplate,
                 $mailData,
-                array($person->getEmailAddress() => $person->getFirstName() . ' ' . $person->getLastName()),
-                $attachments
+                array($person->getEmailAddress()),
```
mickel1138, line 152: Same here use, `Address` instead.

config/packages/dev/mailer.yaml
praneethpr: Okay

src/Util/MailSender.php
```diff
@@ -41,22 +50,23 @@ class MailSender
...
         TwigEnvironment $twig,
         string $fromAddress,
         string $fromName,
         string $bccAddress
     ) {
         $this->twig = $twig;
         $this->mailer = $mailer;
-        $this->from = array($fromAddress => $fromName);
+        $this->fromAddress = $fromAddress;
+        $this->fromName = $fromName;
```
mickel1138, line 69: How about we pass in Address instead. And type hint it like that also.

src/Util/MailSender.php
mickel1138: yes

src/Command/AccountExpirationNotifyCommand.php
```diff
@@ -251,7 +252,7 @@ protected function notifyAccountHolders(array $accounts, \Twig\TemplateWrapper $
                 $emailTemplate,
                 $twigData,
                 array(
-                    $person->getEmailAddress() => $person->getFirstName() . ' ' . $person->getLastName(),
+                    new Address($person->getEmailAddress(), 'GRIIDC'),
```
mickel1138, line 255: use names

mickel1138 (dismissed):
LGTM

mickel1138 (approved):

PR #1309, opened 2022-11-11
Using plugin instead of contain supress

mickel1138:
Using the plugin instead: https://github.com/psalm/psalm-plugin-symfony


fightingtexasaggie (approved):
Psalm and psalm's vscode plugin seem to be working correctly.

PR #1313, opened 2022-11-17
Bugfix/pelagos 5042 elastic index mapping coordinates to int instead of float

mickel1138:


src/Event/DatasetIndexSubscriber.php
```diff
@@ -124,6 +130,28 @@ public function populateCalculatedFields(PostTransformEvent $event)
         }
     }
 
+    /**
+     * Forces array elements to float.
+     *
+     * @param array|float $coordinates
```
praneethpr, line 136: Missing return annotation

praneethpr (approved):
Tesed

src/Event/DatasetIndexSubscriber.php
mickel1138: So, with the new PHP analysis style guides, this no longer needed. You either have a return type on the function, or the the docblock.

PR #1314, opened 2022-11-18
Feature/pelagos 5041 add tag facets to multisearch

mickel1138:


fightingtexasaggie (approved):
Tested this out and it seems to work as expected.

PR #1315, opened 2022-11-21
Release/6.39.0

mickel1138:


config/packages/dev/mailer.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+    mailer:
+        dsn: 'null://null'
```
fightingtexasaggie, line 3: why not import from env like the main one? This would be easier to manage with a local file.

fightingtexasaggie (approved):
Seems ready to go.

praneethpr (approved):

PR #1316, opened 2022-11-21
Feature/pelagos 5049 add facets to template switcher

mickel1138:
This is the modified template switcher. If keys have the same value, one only has to specify it in 'default' and change the departures in each node.

### DIFF
--- a/assets/js/vue/utils/template-switch.js
+++ b/assets/js/vue/utils/template-switch.js
@@ -13,6 +13,7 @@ const templateVariables = {
     digitalTypeDesc: 'Resource Type',
     productTypeDesc: 'Product Type',
     tags: 'Tags',
+    tagsExpanded: false,
   },
   GRIIDC: {
   },
@@ -36,7 +37,6 @@ export default {
   getProperty: (name) => {
     if (
       global.baseTemplateName !== undefined
-            && global.baseTemplateName !== undefined
             && templateVariables[global.baseTemplateName] !== undefined
             && name in templateVariables[global.baseTemplateName]
     ) {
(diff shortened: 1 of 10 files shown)


assets/js/vue/utils/template-switch.js
```diff
@@ -14,50 +14,37 @@ const templateVariables = {
...
   getProperty: (name) => {
     if (
-    // eslint-disable-next-line no-undef
-      PELAGOS_TEMPLATE_PROPS !== undefined
-      // eslint-disable-next-line no-undef
-            && PELAGOS_TEMPLATE_PROPS.BaseTemplateName !== undefined
-      // eslint-disable-next-line no-undef
-            && templateVariables[PELAGOS_TEMPLATE_PROPS.BaseTemplateName] !== undefined
+      global.baseTemplateName !== undefined
+            && global.baseTemplateName !== undefined
```
fightingtexasaggie, line 39: repeated from above...as we found earlier

mickel1138: You didn't refresh, that's old.

templates/hri-base.html.twig
```diff
@@ -163,12 +164,6 @@
         };
       });
     </script>
-
-    <script>
-      window.PELAGOS_TEMPLATE_PROPS = {"BaseTemplateName": "HRI"};
-    </script>
-
 {% endblock %}
-{% endblock htmlbody %}
```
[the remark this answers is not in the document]
mickel1138: It's confusing, but now there is a nested block. I will clean it up a bit.

fightingtexasaggie (approved):
This looks good and is easy enough to use. Are we going to re-name RG's for HRI at some point?

mickel1138:
> This looks good and is easy enough to use. Are we going to re-name RG's for HRI at some point?

Don't know, but it would be possible.

PR #1317, opened 2022-11-23
On facet change, append query param to url

praneethpr:


assets/js/vue/pages/MultiSearchApp.vue
```diff
@@ -99,10 +104,27 @@ export default {
       this.form.perPage = noOfResults;
       this.onSubmit();
     },
+    detectHashChange() {
+      this.route = window.location.hash;
+      this.submitted = false;
+    },
```
mickel1138, line 110: Is this being used still?

fightingtexasaggie (commented):
It seems only a default search is being executed.  Both the query string and facets are being ignored (reload?) if pasted into the URL from a previous search. I had merged in a fresh develop, but just tested without it and it behaves the same way. The URL itself changes after an apparent reload where the values are being lost.

fightingtexasaggie (approved):
Yep, that fixed it! LGTM

PR #1321, opened 2022-12-06
Added hide spinner after GML change

mickel1138:


fightingtexasaggie (approved):
tested by replacing GML and re-rendering on dataset-review. Worked as expected. .always() seems like the right place to hide the spinner as I don't think there's a situation we'd not want to close it.

fightingtexasaggie (approved):

PR #1322, opened 2022-12-07
Catch exception and force page to 1

fightingtexasaggie:


src/Search/SearchResults.php
```diff
@@ -165,7 +166,11 @@ public function __construct(PagerfantaInterface $pagerFantaResults, SearchOption
     private function processResults(): void
     {
         $this->currentPage = $this->searchOptions->getCurrentPage();
-        $this->pagerFantaResults->setCurrentPage($this->currentPage);
+        try {
+            $this->pagerFantaResults->setCurrentPage($this->currentPage);
+        } catch (OutOfRangeCurrentPageException $exception) {
+            $this->pagerFantaResults->setCurrentPage(1);
```
mickel1138, line 172: Maybe also set `$this-currentPage` to 1. Because this does get serialized back out.
As an alternative you can see how many pages there are, and if it's less than currentpage, then set to 1. You could do that with a ternary if.

src/Search/SearchResults.php
```diff
@@ -164,13 +164,13 @@ public function __construct(PagerfantaInterface $pagerFantaResults, SearchOption
...
     {
-        $this->currentPage = $this->searchOptions->getCurrentPage();
-        $this->pagerFantaResults->setCurrentPage($this->currentPage);
         $this->pagerFantaResults->setMaxPerPage($this->searchOptions->getMaxPerPage());
 
         $this->numberOfResults = $this->pagerFantaResults->getNbResults();
         $this->numberOfPages = $this->pagerFantaResults->getNbPages();
         $this->resultsPerPage = $this->pagerFantaResults->getMaxPerPage();
+        $this->currentPage = ($this->searchOptions->getCurrentPage() <= $this->numberOfPages) ? $this->searchOptions->getCurrentPage() : 1;
+        $this->pagerFantaResults->setCurrentPage($this->currentPage);
```
mickel1138, line 173: You should still set current page to 1

fightingtexasaggie: I am setting $this->currentPage = 1 right here. What do you mean?

mickel1138: most have been looking at old code, NM

mickel1138 (approved):

PR #1325, opened 2022-12-12
Feature/pelagos 5057 re order dataland

fightingtexasaggie:


templates/Dataland/v2/index.html.twig
```diff
@@ -38,7 +38,7 @@
                             <div>
                                 <div>
                                     <div>
-                                        <strong>DOI:</strong> {{ dataset.doi.doi|default }}
+                                        <strong>DOI:</strong> {{ dataset.doi.doi|default|formatWebDoi }}
```
mickel1138, line 41: Why not just format the URL here? Rather than using a whole function?

fightingtexasaggie: yeah, you're right. We probably don't need to load more into twig.

templates/Dataland/v2/index.html.twig
```diff
@@ -38,7 +38,10 @@
                             <div>
                                 <div>
                                     <div>
-                                        <strong>DOI:</strong> {{ dataset.doi.doi|default }}
+                                        <strong>DOI:</strong>
+                                        {% if dataset.doi %}
+                                            https://doi.org/{{ dataset.doi.doi|default }}
```
mickel1138, line 43: default would be null, thats not a good url...

mickel1138 (approved):
LGTM

PR #1326, opened 2022-12-14
feature/pelagos-5060-fix-supervisord-paths

fightingtexasaggie:
This is needed if pelagos is intalled at a location other than the pelagos user's homedir. It was as early mistake making the pelagos homedir /opt/pelagos. On new server instances, it is /home/pelagos and the installation still happens at /opt/pelagos as normal. We just shouldn't rely on ~/ referring to the install location anymore. This keeps certain non-pelagos homedir related stuff out of our basedir.


mickel1138 (approved):
LGTM

PR #1327, opened 2022-12-17
Feature/pelagos 5062 update webpack encore

mickel1138:
Upgrading Webpack Encore, this makes it faster, but also mitigates 6 critical issues.


fightingtexasaggie (commented):
.

fightingtexasaggie (approved):
I think this'll blend ;-)

PR #1328, opened 2022-12-20
Added checker for done files

mickel1138:


src/Command/PelagosFileCheckerCommand.php
```diff
@@ -92,7 +92,16 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         /** @var File $file */
         foreach ($files as $file) {
-            $file->setStatus(File::FILE_IN_QUEUE);
+            $filePathName = $file->getFilePathName();
+            if (str_contains($file->getPhysicalFilePath(), $filePathName)) {
+                $file->setStatus(File::FILE_DONE);
+            } else {
+                // if filePathName contains "funkywhitespace"...
+                if (preg_match("/\p{C}+/", $filePathName)) {
```
mickel1138, line 100: You'd probably want and the `u` here too.,

fightingtexasaggie (approved):
I think we all worked on this, but after today's test of it, running it, letting it finish, and running it again cleared all errors. This gets us out of the bind we're in with S3.x932.000:0001

PR #1330, opened 2023-01-03
Putting Phpstan back, for tool use

mickel1138:


fightingtexasaggie (approved):
I discussed this with Mike this morning and agree it's a good idea.

PR #1348, opened 2023-01-18
Feature/pelagos 5092 add funder to submission and review forms

praneethpr:


mickel1138:
If you want to know the relationship of a collection from both sides, you have to set the relationship on each side.

This is may not be really correct, but instead we could maybe just have the relationship on one side.

fightingtexasaggie (commented):
Submission now fails as Funder is set as a required field.

assets/static/js/dataset-submission.js
```diff
@@ -33,6 +33,47 @@ $(function() {
...
+        return arr;
+    }
+
+    const addedFunders = [];
+
+    $.ajax({
+        url: Routing.generate('pelagos_api_datasets_get_collection') + '?udi=' + $("#regForm").attr("udi") + '&_properties=funders.id',
+        success: function(data){
+            addedFunders.push
+              let associatedFunders = data[0].funders;
```
fightingtexasaggie, line 52: this reference to data[0] in console shows "Uncaught TypeError: data[0] is undefined"

praneethpr: I should've added a check, good catch

praneethpr: Fixed it

assets/static/js/funderComponent.js
```diff
@@ -0,0 +1,62 @@
+//FOUC preventor
+$("html").hide();
```
mickel1138, line 2: This "Flash Of Unstyled Content" should be left in the submission form.

assets/static/js/datasetReview.js
```diff
@@ -2,11 +2,11 @@ var $ = jQuery.noConflict();
 var geowizard;
 
 //FOUC preventor
-$("html").hide();
+$("html").hide();
```
mickel1138, line 5: Leave this here.

assets/static/js/dataset-submission.js
```diff
@@ -157,10 +157,20 @@ $(function() {
         $("#largeFileUri").val($(this).val()).trigger("change");
     });
 
+    $("#funderList").on("keyup change", function() {
+        $(this).valid();
+    });
+
```
mickel1138, line 163: Is this needed, it will validate all the time, not just when the Red X is there.

praneethpr: Removed it

migrations/Version20230202151933.php
```diff
@@ -10,18 +10,21 @@
...
+        return '';
     }
 
     public function up(Schema $schema) : void
     {
         // this up() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
 
+        $this->addSql('ALTER TABLE dataset_submission ADD additional_funders TEXT DEFAULT NULL');
+        $this->addSql('ALTER TABLE dataset_submission ALTER cold_storage_total_unpacked_size TYPE INT');
```
fightingtexasaggie, line 26: We're gonna need a bigger boat, see https://www.postgresql.org/docs/14/datatype-numeric.html Since INT is -2147483648 to +2147483647, this would give us a max cold storage filesize of 2GB. (it's signed) We need bigint.

mickel1138: This doesn't belong here, and should be removed.

migrations/Version20230202151933.php
```diff
@@ -30,6 +33,9 @@ public function down(Schema $schema) : void
         // this down() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
 
+        $this->addSql('ALTER TABLE dataset_submission DROP additional_funders');
+        $this->addSql('ALTER TABLE dataset_submission ALTER cold_storage_total_unpacked_size TYPE BIGINT');
+        $this->addSql('ALTER TABLE dataset_submission ALTER cold_storage_total_unpacked_size DROP DEFAULT');
```
mickel1138, line 38: Remove

fightingtexasaggie (approved):
Ok, I think this is ready to merge into the epic for now.

mickel1138 (approved):
Merge

PR #1351, opened 2023-01-23
feature/PELAGOS-5099 (ldap robustness)

fightingtexasaggie:


fightingtexasaggie (commented):
Testing reveals still needs try/catch blocks also in LdapClient.php

PR #1357, opened 2023-02-03
Upgrades for PHP 8.1

mickel1138:
I did not run composer update, it will change some major versions, and we have to check if that is ok.


mickel1138:
https://github.com/phpstan/phpstan/issues/6668

fightingtexasaggie (commented):
Also one in datasetSubmissionController, line 203.

fightingtexasaggie (approved):
I think you got them all. LGTM.

PR #1364, opened 2023-02-15
Added funder facet

praneethpr:


assets/js/vue/components/search/ResultSet.vue
```diff
@@ -38,7 +38,8 @@
                     <div class="card card-filter">
                         <Facet :facet-info="results.facetInfo.statusInfo" :facet-name="facetLabels.status" v-on="$listeners" :formValues="formValues"/>
                         <Facet :facet-info="results.facetInfo.fundingCycleInfo" :facet-name="facetLabels.fundingCycle" v-on="$listeners" :formValues="formValues" v-show="showFundingCycleFacet"/>
-                        <Facet :facet-info="results.facetInfo.fundingOrgInfo" :facet-name="facetLabels.fundingOrg" v-on="$listeners" :formValues="formValues" v-show="showFundingOrgFacet"/>
+                        <Facet :facet-info="results.facetInfo.funderInfo" :facet-name="facetLabels.funder" v-on="$listeners" :formValues="formValues"/>
+                        <!-- <Facet :facet-info="results.facetInfo.fundingOrgInfo" :facet-name="facetLabels.fundingOrg" v-on="$listeners" :formValues="formValues" v-show="showFundingOrgFacet"/> -->
```
praneethpr, line 42: Should I go ahead and remove all funding Org related code?

src/Util/Search.php
```diff
@@ -486,6 +491,56 @@ private function getProjectDirectorInfo(array $aggregations): array
...
+        foreach ($funders as $funder) {
+            $fundersInfo[$funder->getId()] = array(
+                'id' => $funder->getId(),
+                'name' => $funder->getName(),
+                'count' => $aggregations[$funder->getId()]
+            );
+        }
+        //Sorting based on highest count
+        $array_column1 = array_column($fundersInfo, 'count');
+        $array_column2 = array_column($fundersInfo, 'name');
```
mickel1138, line 534: Maybe these could use better names, but it's not too important.

config/packages/fos_elastica.yaml
```diff
@@ -152,6 +152,11 @@ fos_elastica:
                     property_path: false
                     type: 'date'
                     format: 'yyyy-MM-dd'
+                funders:
+                    type: 'nested'
+                    properties:
+                        id:
```
fightingtexasaggie, line 158: Do we need to index name as well?

src/Util/Search.php
praneethpr: Yes, but i felt this was anyways going to be replaced with multi-search

config/packages/fos_elastica.yaml
praneethpr: No, we don't need to index name, because we are searching based on `id`

mickel1138 (approved):

PR #1370, opened 2023-02-28
Added specific field and map search button

praneethpr:


src/Search/MultiSearch.php
```diff
@@ -32,6 +32,31 @@ class MultiSearch
...
+
+    /**
+     * Elastic index mapping for theme keywords.
+     */
+    const ELASTIC_INDEX_MAPPING_THEME_KEYWORDS = 'datasetSubmission.themeKeywords';
+
+    /**
+     * Index boost for Title, Authors, Theme Keywords.
+     */
+    const BOOST = '^2';
```
mickel1138, line 58: Make private? Also, other maybe public?

src/Search/SearchOptions.php
```diff
@@ -119,6 +119,13 @@ class SearchOptions
      */
     private $rangeEndDate;
 
+    /**
+     * Specific field to be searched upon.
+     *
+     * @var string
+     */
+    private $field;
```
mickel1138, line 127: Shall we start type hinting here? I hear `@var` is a bad habit.

fightingtexasaggie (commented):
Search for 'all' doesn't seem to work right. Example: searching for 'dolphin' in title returns 19 results, but no results when set to 'all'.

src/Controller/UI/MultiSearchController.php
```diff
@@ -54,6 +55,7 @@ public function search(Request $request, MultiSearch $multiSearch, JsonSerialize
         $searchOptions->setRangeStartDate($rangeStartDate);
         $searchOptions->setRangeEndDate($rangeEndDate);
         $searchOptions->setSortOrder($sortOrder);
+        $searchOptions->setField($field);
```
fightingtexasaggie, line 58: perhaps only call if not empty string

fightingtexasaggie (approved):
Confirmed it is now working as expected.

PR #1371, opened 2023-03-01
Misc/test case fix funder

mickel1138:


fightingtexasaggie (approved):
Looks good, tests all run as expected.

PR #1372, opened 2023-03-01
Added tooltip, moved tags

praneethpr:


assets/js/vue/utils/template-switch.js
```diff
@@ -20,7 +20,7 @@ const templateVariables = {
     tagsExpanded: true,
   },
   GRIIDC: {
-    typeExpanded: false,
+    typeExpanded: true,
```
mickel1138, line 23: This can also be removed, as the default is true.

src/Entity/Dataset.php
```diff
@@ -59,17 +59,17 @@ class Dataset extends Entity
     /**
      * Cold Storage Tag
      */
-    const TAG_COLDSTORAGE = 'Coldstorage';
+    const TAG_COLD_STORAGE = 'Cold Storage';
```
mickel1138, line 62: Make public

assets/js/vue/utils/template-switch.js
praneethpr: Okay

src/Entity/Dataset.php
praneethpr: We can run them when are merging this epic into develop?

src/Entity/Dataset.php
```diff
@@ -81,9 +81,9 @@ class Dataset extends Entity
      * Valid Tags for a dataset.
      */
     const TAGS = [
```
fightingtexasaggie, line 83: Tag hover values need to match the strings in the ticket.

fightingtexasaggie (commented):
While it makes sense to put the hover text on the badges like was done here, I think the PO wanted them on the facet tags, instead of a repeat of the same text as hover.. It honestly makes sense in both places to me.

fightingtexasaggie (approved):

PR #1374, opened 2023-03-07
Misc/upgrade doctrine

mickel1138:


fightingtexasaggie (dismissed):
I reviewed and hammered on this for a while. It seems to be fine.

fightingtexasaggie (commented):
sidenote: 140 topic_keywords are stored as key/value pairs, but they were like this in the data beforehand, and they load up fine in review/view.

PR #1383, opened 2023-03-22
Feature/pelagos 5114 create gcmd keywords entity table

mickel1138:
This PR can be good to go, but if we don't want to commit to Enums, and PHP 8.1, we can use old Constants instead.

### DIFF
--- a/src/Entity/Keyword.php
+++ b/src/Entity/Keyword.php
@@ -32,7 +32,6 @@ class Keyword extends Entity
         return $this->json;
     }
 
-
     /**
      * Set the JSON for this Keyword.
      */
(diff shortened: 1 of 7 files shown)


src/Entity/Keyword.php
```diff
@@ -0,0 +1,62 @@
...
+ * Keywords class for Standardized keywords.
+ */
+#[ORM\Entity(repositoryClass: KeywordRepository::class)]
+class Keyword extends Entity
+{
+    /**
+     * JSON containing Standard Keyword Data.
+     */
+    #[ORM\Column]
+    private array $json = [];
```
mickel1138, line 19: Does should allow for `mixed` unless you turn every json into an `array` which seems not practical.

PR #1388, opened 2023-03-27
Bugfix/pelagos 5136 add empty callback

fightingtexasaggie:


assets/static/js/common.js
```diff
@@ -105,3 +105,7 @@ function showConfirmation(options) //title, message, yesText, noText
         }, options));
     });
 }
+
+function gmapsCallback()
```
fightingtexasaggie, line 109: This seems to not get loaded in time for dataland, or probably any of the vue apps.

praneethpr: If we are going to add an empty callback, why not just add `Function.prototype` in those script tags

PR #1392, opened 2023-03-30
Show datasets when search terms doesn't exist

praneethpr:


fightingtexasaggie (approved):
This works as intended after I merged it into the epic to test, otherwise I was getting DB errors.

mickel1138 (approved):
LGTM

PR #1393, opened 2023-03-30
Simply use Function.prototype as the passed callback.

fightingtexasaggie:
Re-did this without all the whitspace fixups that vscode did on MDE's index.php.


mickel1138 (approved):
LGTM

PR #1396, opened 2023-04-19
Epic/pelagos 5113 add standardized keywords

mickel1138:


composer.json
```diff
@@ -6,7 +6,7 @@
     "homepage": "https://github.com/griidc/pelagos/",
     "license": "BSD-2-Clause",
     "require": {
-        "php": "^8.0",
+        "php": "^8.1",
```
mickel1138, line 9: This should just be removed, there is no need for it. But we are going to do it later.

fightingtexasaggie (approved):
Ok, this looks ready to rock.

PR #1399, opened 2023-04-19
Order Funder by name in repo.

fightingtexasaggie:
This also fixes the order everywhere else (review, submission).  When we switch to shortname, we'll have to decide whether order should be based on name or short name, but for now it's name.


mickel1138 (approved):
LGTM

PR #1402, opened 2023-05-02
Hot Fix for Google Map Tiles not loading

mickel1138:


fightingtexasaggie (approved):
Seems pretty straight forward, working.

PR #1409, opened 2023-06-05
Feature/pelagos 5164 clear list on reset

mickel1138:
This has a previous ticket merged in.


fightingtexasaggie (approved):
Seems to work fine. The only other thing probably not of concern is the control is active even if a locked RG is selected, but I don't think this is an issue.

mickel1138:
> Seems to work fine. The only other thing probably not of concern is the control is active even if a locked RG is selected, but I don't think this is an issue.

This will be done in another ticket.

PR #1410, opened 2023-06-07
Feature/pelagos 5122 replace funding org with funder

fightingtexasaggie:


fightingtexasaggie:
Currently the funder filter works for datasets, but not for info products. I was thinking we'd base it on the InformationProduct->researchGroups->fundingCycle->findingOrganization->defaultFunder. Is this a reasonable approach?

mickel1138 (approved):
LGTM

PR #1414, opened 2023-06-15
Feature/pelagos 5167 add sort to initial paramaterless search results on current search tool

fightingtexasaggie:
The acceptance criteria cite acceptedDate, which is what the sortingDateForDisplay already gets for cases of accepted submission. In other cases it falls through to the submissiontimestamp, but those are already  grouped below all the accepted datasets per the 'accepted' bool I set up in the index, populated by datasetIndexSubscriber.

If there's a cleaner approach, please let me know.

### DIFF
--- a/src/Util/Search.php
+++ b/src/Util/Search.php
@@ -2,16 +2,18 @@
 
 namespace App\Util;
 
-use App\Entity\DatasetSubmission;
-use App\Entity\Funder;
-use App\Entity\FundingCycle;
-use App\Entity\Person;
-use App\Entity\ResearchGroup;
-use Doctrine\ORM\EntityManagerInterface;
-use Elastica\Aggregation;
 use Elastica\Query;
-use FOS\ElasticaBundle\Finder\TransformedFinder;
+use App\Entity\Funder;
+use App\Entity\Person;
+use Elastica\Aggregation;
 use Pagerfanta\Pagerfanta;
+use App\Entity\FundingCycle;
+use App\Entity\ResearchGroup;
+use RecursiveIteratorIterator;
+use App\Entity\DatasetSubmission;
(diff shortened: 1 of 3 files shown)


src/Util/Search.php
```diff
@@ -134,6 +136,25 @@ public function buildQuery(array $requestTerms): Query
...
+            && ($requestTerms['sortOrder'] == 'default')
+            && empty($requestTerms['collectionStartDate'])
+            && empty($requestTerms['collectionEndDate'])
+            && empty($requestTerms['options']['rgId'])
+            && empty($requestTerms['options']['status'])
+            && empty($requestTerms['options']['fundingCycleId'])
+            && empty($requestTerms['options']['projectDirectorId'])
+            && empty($requestTerms['options']['funderId'])
+        ) {
+            $defaultSearch = true;
```
mickel1138, line 153: Rename to to something like `$isInititalSearch`

src/Util/Search.php
```diff
@@ -173,7 +194,11 @@ public function buildQuery(array $requestTerms): Query
         $mainQuery->setQuery($subMainQuery);
 
         // Add sort order
-        if ('default' !== $sortOrder) {
+        if ($defaultSearch) {
+            // Show accepted datasets first. The 'accepted' is a boolean set where datasetSubmission
+            // status == 'Accepted' by the DatasetIndexSubscriber.
+            $mainQuery->addSort(['accepted' => ['order' => 'desc'], self::ELASTIC_INDEX_MAPPING_SORTING_DATE => ['order' => 'desc']]);
```
mickel1138, line 200: Drop second parameter in array,  and use accepted date (also add to directive).

mickel1138 (approved):
LGTM

PR #1418, opened 2023-07-03
Feature/pelagos 5172 add keywords to dataset submission

mickel1138:


fightingtexasaggie:
Only a single keyword shows up in dataset review (but all came across in the database from the DIF to the submission, and also the new submission that dataset review ends up making doesn't copy any of the values over.

src/Form/DatasetSubmissionType.php
```diff
@@ -149,6 +150,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                'label' => 'Keywords',
+                'entry_type' => EntityType::class,
+                'entry_options' => [
+                    'class' => Keyword::class,
+                ],
+                'by_reference' => true,
+                'allow_add' => true,
+                'allow_delete' => true,
+                'delete_empty' => true,
+                'required' => false,
```
fightingtexasaggie, line 163: Isn't this a required field now?

fightingtexasaggie (approved):

PR #1421, opened 2023-07-14
Add funder to IP

mickel1138:


src/Controller/Api/InformationProductController.php
```diff
@@ -207,14 +216,15 @@ public function deleteInformationProduct(Request $request, InformationProduct $i
         $entityManager = $this->getDoctrine()->getManager();
```
mickel1138, line 216: Remove getDoctrine before going to 6

fightingtexasaggie (approved):
Confirmed that the IP file upload is broken on griidc-test altogether on 6.48.0 and this isn't related to this PR. Otherwise, I couldn't really find anything odd.

PR #1427, opened 2023-07-19
Feature/pelagos 5185 re order results accepted first by accepted date all others by submit date

fightingtexasaggie:


src/Entity/Entity.php
```diff
@@ -73,6 +73,7 @@ abstract class Entity
      * The last modification time stamp (in UTC) for this Entity.
      *
      * @var \DateTime $modificationTimeStamp;
+     * @Serializer\Groups({"search", "card"})
```
fightingtexasaggie, line 76: Is there a way to do this just for the class that extends this base class attribute. I really don't want to expose it for everything.

config/packages/fos_elastica.yaml
```diff
@@ -226,4 +226,8 @@ fos_elastica:
                 publishedDate:
                     property_path: false
                     type: 'date'
-                    format: 'yyyy-MM-dd'                
+                    format: 'yyyy-MM-dd'
+                acceptedDate:
```
fightingtexasaggie, line 230: If there is a better way to treat both accepted date for datasets and published date on info products indexes the same, that would be a better approach than faking a redundant field.

assets/js/vue/components/information-product/InformationProductCard.vue
```diff
@@ -22,6 +22,9 @@
         <div v-if="informationProduct.publisher">
           Publisher: {{ informationProduct.publisher | truncate(100) }}
         </div>
+        <div>
+          Created/Updated: {{ informationProduct.modificationTimeStamp }}
```
fightingtexasaggie, line 26: I felt something like this is needed in the info product card, because it shows why the card is in the sort. Also, need to figure out the moment library to do date formatting both here and in the dataset published date field (same ugliness...in production like that).

mickel1138 (approved):
Tested, LGTM

PR #1429, opened 2023-07-26
Escape querystring to esacep Ampersand

mickel1138:


fightingtexasaggie (approved):
Reviewed with MVDE, tested, LGTM.

PR #1435, opened 2023-08-14
Misc/ip search bugs

mickel1138:


fightingtexasaggie (approved):
Changes reviewed. Counter/pager works correctly now, and tested non-published IPs don't display...all good.

PR #1441, opened 2023-08-29
Feature/pelagos 5194 add keywords to review

mickel1138:


fightingtexasaggie (approved):
Tested, including merging in 5119 and validated xml output in Oxygen 20. Review works as expected. Also tested adding/removing new keywords during review.

PR #1442, opened 2023-09-06
Feature/pelagos 5191 add anzsrc keywords to metadata generator

fightingtexasaggie:
Switched back to 'revisions' string in MD template.


mickel1138 (approved):
This is correct.

PR #1448, opened 2023-09-25
Feature/pelagos 5208 make keyword add tool to dataset

mickel1138:


fightingtexasaggie (approved):
This is a pretty slick little tool! The only thing I noticed is it's a little confusing with the 0 keywords filter showing first. I think it would make more sense just sorting ascending, but I understand that's what the PO wanted.

PR #1451, opened 2023-10-10
Feature/pelagos 5211 change list styling to look like tagbox 2

mickel1138:


fightingtexasaggie (approved):
LGTM, but then again, looked alright before too imho.

PR #1452, opened 2023-10-12
Feature/pelagos 5212 add keywords to dataland

mickel1138:


fightingtexasaggie (approved):
Reviewed. Tested both keyword sections. I'd of suggested a bulleted list over a comma separated list as some of the keywords might contain commas, but no biggie.

PR #1455, opened 2023-10-18
Temporary disable standard keywords

mickel1138:


fightingtexasaggie (approved):
Reviewed. DIF seems to still work just fine after the temporary changes. Also, it'll have to be non-mandatory on review otherwise datasets are going to get held up before they're ready to apply keywords. We should at least discuss with DRPM.

PR #1456, opened 2023-10-18
Don't show keywords on submission

mickel1138:


fightingtexasaggie (commented):
This currently breaks dataset-review because the same twig template is used by dataset-review.

fightingtexasaggie (approved):
LGTM!

PR #1463, opened 2023-11-09
Only count within 30 seconds of download

mickel1138:


fightingtexasaggie (approved):
reviewed, tested, also confirmed 30 second event window is working on dataland.

PR #1469, opened 2023-11-14
Release/6.51.0

mickel1138:


src/Util/Metadata.php
```diff
@@ -19,14 +19,39 @@ class Metadata
      */
     protected $twig;
 
+    /**
+     * The env variable for GCMD Keyword version.
+     *
+     * @param gcmdVersion
+     */
+    protected $gcmdVersion;
```
mickel1138, line 27: The @param is really needed, they is not type defined anyway.
Also I think it can be a private. Alternatively this can just be defined right in the constructor.

No action is needed, just for next time.

mickel1138:
Some docblocks were missed for titles of routes or functions. But not a big deal.
Also some extra lines or new lines here and there.

All seems to work just fine.

fightingtexasaggie (approved):
LGTM

PR #1470, opened 2023-11-17
Added USGS topo map

mickel1138:
Had to disable small map when wizard loads.


fightingtexasaggie (approved):
Gets us out of a bind fast

PR #1471, opened 2023-11-28
Corrected filename

mickel1138:
Because filename would say first filename, because zip is no longer available. Also remove SHA256 because zip is now dynamic.


fightingtexasaggie (approved):
LG

PR #1472, opened 2023-11-28
Release/6.52.0

mickel1138:


fightingtexasaggie (approved):
Confirmed this was only the epic we'd already approved, plus the hotfix for the tiles.

PR #1473, opened 2023-11-29
Misc/new griidc template

mickel1138:


templates/griidc.html.twig
```diff
@@ -0,0 +1,240 @@
...
+                    "triggerFunction": function(showCollectorDialog) {
+                        $("#atlwdg-suggest").click(function(e) {
+                            e.preventDefault();
+                            showCollectorDialog();
+                        });
+                    }
+                },
+            };
+        });
+    </script>
```
mickel1138, line 231: This is for issues collector, and can go.

PR #1475, opened 2023-12-01
Connect download all button in filemngr to zip streamer and log.

fightingtexasaggie:
hotfix - but we need to discuss "download all" on dataset submission as that is still broke.

Update: it's not an issue on submission because we don't show the button in file mangaer when called from dataset submission. All should be good.

### DIFF
--- a/src/Controller/Api/FileManager.php
+++ b/src/Controller/Api/FileManager.php
@@ -387,7 +387,7 @@ class FileManager extends AbstractFOSRestController
     public function doesZipFileExist(DatasetSubmission $datasetSubmission): Response
     {
         return new Response(
-            json_encode($datasetSubmission?->getFileset()?->getAllFiles()?->count() > 0),
+            json_encode($datasetSubmission?->getFileset()?->getProcessedFiles()?->count() > 0),
             Response::HTTP_OK,
             array(
                 'Content-Type' => 'application/json',
--- a/src/Controller/Api/FileManager.php
+++ b/src/Controller/Api/FileManager.php
@@ -332,7 +332,6 @@ class FileManager extends AbstractFOSRestController
         $udi = $dataset->getUdi();
         // Only log if this is downloaded from dataland. We don't log review downloads.
         if ($request->headers->get('referer') and preg_match("/^.*\/data\/$udi$/", $request->headers->get('referer'))) {
-
             $currentUser = $this->getUser();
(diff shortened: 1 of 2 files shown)


src/Controller/Api/FileManager.php
```diff
@@ -323,70 +326,35 @@ public function downloadFile(File $file, Datastore $datastore): Response
...
+                'file_download'
             );
-            $response->headers->set('Content-Disposition', $disposition);
-            return $response;
-        } else {
-            throw new BadRequestHttpException('No Zip file found');
         }
-    }
 
-    /**
```
mickel1138, line 355: Don't remove these functions, it's not part of this ticket. And they are still being used by other actions!

src/Controller/Api/FileManager.php
```diff
@@ -323,70 +326,35 @@ public function downloadFile(File $file, Datastore $datastore): Response
...
-                $fileStream = fopen($zipFilePath, 'r');
-                stream_copy_to_stream($fileStream, $outputStream);
-            });
-            $disposition = HeaderUtils::makeDisposition(
-                HeaderUtils::DISPOSITION_ATTACHMENT,
-                basename($zipFilePath)
+        $dataset = $datasetSubmission->getDataset();
+        $udi = $dataset->getUdi();
+        // Only log if this is downloaded from dataland. We don't log review downloads.
+        if ($request->headers->get('referer') and preg_match("/^.*\/data\/$udi$/", $request->headers->get('referer'))) {
```
mickel1138, line 334: Make a variable for Referer, it's at least being used twice.

src/Controller/Api/FileManager.php
```diff
@@ -323,70 +326,35 @@ public function downloadFile(File $file, Datastore $datastore): Response
      *
      * @return Response
      */
-    public function downloadZipAllFiles(DatasetSubmission $datasetSubmission): Response
+    public function downloadZipAllFiles(DatasetSubmission $datasetSubmission, LogActionItemEventDispatcher $logActionItemEventDispatcher, Request $request): Response
```
mickel1138, line 329: The docblock can be fixed up some. Order, remove unneeded (or add) params. @Route should be last.

mickel1138 (approved):

mickel1138 (approved):

PR #1483, opened 2023-12-18
Feature/pelagos 5250 fix data monitoring for new template

mickel1138:


assets/static/css/dataset-monitoring.css
```diff
@@ -49,3 +49,7 @@
 .awesome-icon {
     cursor: pointer;
 }
+
+img.bobble-img {
```
mickel1138, line 53: the first first could probably be removed.

PR #1485, opened 2023-12-20
Feature/pelagos 5251 styling fixes for stats on new template

mickel1138:


fightingtexasaggie (approved):
Reviewed and the only minor thing I noticed was ^M chars on the griidc.scss file, which isn't really a big deal.

PR #1486, opened 2023-12-20
Feature/pelagos 5250 fix data monitoring for new template

mickel1138:


fightingtexasaggie (approved):
I'm unsure if the horizontal scroll at typical default resolution is going to be acceptable to the PO, but I do feel it's more of a PO issue anyway.  Otherwise I think it's good to go. I merged in a fresh epic without  issue as well.

PR #1489, opened 2024-01-05
Update caniuse db

fightingtexasaggie:
Should this maybe be part of our build process instead of committing it?


mickel1138 (approved):
LGTM

PR #1497, opened 2024-01-11
Prevent enter on all types of input

mickel1138:


fightingtexasaggie (approved):
I did verify every different field type and the only thing I noticed 'enter' doing was expanding/collapsing the tree view of the keywords, which I think is fine. No submits were triggered.

PR #1498, opened 2024-01-17
Feature/pelagos 5253 dataset submission breaks menu s o m e t i m e s

mickel1138:


fightingtexasaggie (commented):
I'm no longer able to add keywords during dataset review. There is no console error, just nothing seems to happen.

mickel1138:
> I'm no longer able to add keywords during dataset review. There is no console error, just nothing seems to happen.

I can't find anything wrong, check again. This error happened when JMS serializer was not yet fixed.

fightingtexasaggie (approved):

PR #1502, opened 2024-01-24
Fixed tombstone style

mickel1138:


fightingtexasaggie (approved):
looks good in griidc-base...and still fine in hri, gomri, grp base.

PR #1504, opened 2024-01-25
Changed zoom to level to 8 (was 11)

mickel1138:


fightingtexasaggie (approved):
tested and map seems to load on every geometry I can find, especially the small footprint ones that had been problematic before. I think this is a good level.

PR #1506, opened 2024-01-25
Removed forced font

mickel1138:


fightingtexasaggie (approved):
Fonts identical Lato now per console tool inspect. Visually, the right hand pane on screen is still a slightly different color, but I think that's another issue.

PR #1507, opened 2024-01-26
Added DL counts, date optional. Filter aware.

fightingtexasaggie:


src/Command/GetGoMRIStatisticsCommand.php
```diff
@@ -98,6 +112,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $io->writeln("Total number of GoMRI datasets submitted since 2021-01-01 until current date " . $totalGomriDatasetSubmittedSince2021Count);
         $io->writeln("Total number of datasets (all data including GoMRI) submitted since 2021-01-01 until current date " . $totalDatasetSubmittedSince2021Count);
 
+        if ($start != '' and $end != '') {
+            $io->writeln("Total GoMRI Downloads from $start to $end: " . $this->logActionItemRepository->countDownloads($start, $end));
```
fightingtexasaggie, line 116: Turn into a proper tz containing datetime object and pass those instead of strings.

src/Command/GetGoMRIStatisticsCommand.php
```diff
@@ -27,28 +28,41 @@ class GetGoMRIStatisticsCommand extends Command
...
             ->setDescription(self::$defaultDescription)
+            ->addArgument('start', InputArgument::OPTIONAL, 'Starting date for DL count')
+            ->addArgument('end', InputArgument::OPTIONAL, 'Ending date for DL count')
         ;
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
+        $start = $input->getArgument('start');
```
fightingtexasaggie, line 64: Sanity check these for date format at least.

fightingtexasaggie:
Use of strtotime always assumes utc if unspecified, but it's a great time parser, I think worth the use limitation.

with central TZ example usage: 
pelagos:get-gomri-statistics  '2023-10-01 00:00:00-05' '12/31/2023 23:59:59-06'

src/Repository/LogActionItemRepository.php
```diff
@@ -35,20 +35,35 @@ public function __construct(ManagerRegistry $registry, FundingOrgFilter $funding
...
 
     /**
-     * Sum of all dataset file sizes.
+     * Count of downloads, optional date range and also filter aware.
      *
-     * @return integer Size of data in bytes.
+     * @return integer The count of the datasets downloaded, per FAIR guidelines.
      */
-    public function countDownloads(): int
+    public function countDownloads(string $start = '', string $stop = ''): int
```
mickel1138, line 42: Null as default is a better option.

src/Repository/LogActionItemRepository.php
```diff
@@ -35,20 +35,35 @@ public function __construct(ManagerRegistry $registry, FundingOrgFilter $funding
...
+            ->andWhere('log.creationTimeStamp >= ?3')
+            ->andWhere('log.creationTimeStamp <= ?4')
+            ->orderBy('log.subjectEntityId', 'ASC')
+            ->addOrderBy('log.creationTimeStamp', 'ASC')
+            ->setParameter(1, 'Pelagos\Entity\Dataset')
+            ->setParameter(2, 'File Download')
+            ->setParameter(3, $start)
+            ->setParameter(4, $stop);
+        } else {
+            $qb = $this->createQueryBuilder('log')
```
mickel1138, line 58: Instead of duplication this code, just do the same code before the `if` and add the extra `andWhere` and `setParameter` in the if.

src/Repository/LogActionItemRepository.php
```diff
@@ -35,20 +35,35 @@ public function __construct(ManagerRegistry $registry, FundingOrgFilter $funding
...
+     * Count of downloads, optional date range and also filter aware.
      *
-     * @return integer Size of data in bytes.
+     * @return integer The count of the datasets downloaded, per FAIR guidelines.
      */
-    public function countDownloads(): int
+    public function countDownloads(string $start = '', string $stop = ''): int
     {
-        $qb = $this->createQueryBuilder('log')
+        if ($start != '' and $stop != '') {
```
mickel1138, line 44: If default is `null`  as suggested, just `empty` instead to check.

src/Command/GetGoMRIStatisticsCommand.php
```diff
@@ -98,6 +114,27 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+            // will assume DST-aware central time if not specified.
+            $startDateTime->setTimezone(new \DateTimeZone('America/Chicago'));
+            $startDateTime->setTimestamp($startEpoch);
+            // will include tz in DB string
+            $dbFormatStartTime = $startDateTime->format('Y-m-d H:i:sO');
+
+            $endDateTime = new \Datetime();
+            $endDateTime->setTimezone(new \DateTimeZone('America/Chicago'));
+            $endDateTime->setTimestamp($endEpoch);
+            $dbFormatEndTime = $endDateTime->format('Y-m-d H:i:sO');
```
mickel1138, line 128: Why not pass DateTime as start end parameters, and let the function format it the way it wants to use it for the query.

You can also create a DateTime in one go from a string.

src/Command/GetGoMRIStatisticsCommand.php
```diff
@@ -27,29 +28,44 @@ class GetGoMRIStatisticsCommand extends Command
...
+            ->addArgument('end', InputArgument::OPTIONAL, 'Ending date for DL count')
         ;
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
 
+        $startEpoch = strtotime($input->getArgument('start'));
+        $endEpoch = strtotime($input->getArgument('end'));
```
mickel1138, line 66: Why create a new `Datetime` here instead.

mickel1138: I see, datetime with nothing becomes now. You could still do it, but just check if the argument is set.

mickel1138 (approved):

PR #1508, opened 2024-01-26
Added DOI to pubs api for RG

mickel1138:


fightingtexasaggie (approved):
Tested, title visible and linked card to the doi works.

PR #1512, opened 2024-01-29
Feature/pelagos 5255 adjust subsite template to match new templates

mickel1138:


fightingtexasaggie (approved):
Looks fine, size fine.

PR #1513, opened 2024-01-29
GOMRI homepage updates

mickel1138:


fightingtexasaggie (commented):
Are we also going to change the logo in the header on this? I only see the new griidc logo in the footer, and the old griidc logo in the header on the gomri template now.

fightingtexasaggie (approved):
Looks good, tested by merging in 5255.

PR #1520, opened 2024-02-01
Picked datapicker z-index and padding for select

mickel1138:


fightingtexasaggie (approved):
Tested, works and looks clean with no elements blocked. (Unrelated, copy/paste is working again after using map wizard)

PR #1527, opened 2024-02-09
Need field name in template.

fightingtexasaggie:
Server crashes without this! (causes nasty db-loop, which we should probably investigate deeper as sending this form in with a null from an un-authenticated field results in an easy-to-violate vulnerability resulting in a crashed instance.)

### DIFF
--- a/src/Controller/UI/AccountController.php
+++ b/src/Controller/UI/AccountController.php
@@ -109,7 +109,7 @@ class AccountController extends AbstractController
         $emailAddress = $request->request->get('emailAddress');
         $reset = ($request->request->get('reset') == 'reset') ? true : false;
 
-        // Don't trust the public untrusted form to not send nulls/empties. EH doesn't like nulls.
+        // Don't allow null/empty field.
         if (empty($emailAddress)) {
             $people = [];
         } else {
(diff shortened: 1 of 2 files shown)


src/Controller/UI/AccountController.php
```diff
@@ -109,7 +109,12 @@ public function sendVerificationEmail(Request $request, MailSender $mailer)
         $emailAddress = $request->request->get('emailAddress');
         $reset = ($request->request->get('reset') == 'reset') ? true : false;
 
-        $people = $this->entityHandler->getBy(Person::class, array('emailAddress' => $emailAddress));
+        // Don't trust the public untrusted form to not send nulls/empties. EH doesn't like nulls.
+        if (empty($emailAddress)) {
+            $people = [];
+        } else {
+            $people = $this->entityHandler->getBy(Person::class, array('emailAddress' => $emailAddress));
```
mickel1138, line 116: Let's rewrite this without using `entityHandler`.

src/Controller/UI/AccountController.php
```diff
@@ -109,7 +109,12 @@ public function sendVerificationEmail(Request $request, MailSender $mailer)
         $emailAddress = $request->request->get('emailAddress');
         $reset = ($request->request->get('reset') == 'reset') ? true : false;
 
-        $people = $this->entityHandler->getBy(Person::class, array('emailAddress' => $emailAddress));
+        // Don't trust the public untrusted form to not send nulls/empties. EH doesn't like nulls.
```
mickel1138, line 112: Please reword this comment.

fightingtexasaggie: Yeah, that's giving a bit too much info in retrospect

PR #1528, opened 2024-02-10
Feature/pelagos 5307 update links on profile and submission

mickel1138:


templates/DatasetReview/file-submit.html.twig
```diff
@@ -25,9 +25,9 @@
             <div id="tabs-2"> <!-- Upload via SFTP/Ingest-->
                 <p>Dataset files that are too large via the normal Upload mechanism can be uploaded to the griidc-ingest
                     server using Globus or SFTP and then the folder selected below.  See your
-<a href="{{ path('pelagos_app_ui_person_profile_page', {'id': app.user.id}) }}" TARGET="_NEW">user profile page</a>
+<a style="color: #2563EB" href="{{ path('pelagos_app_ui_person_profile_page', {'id': app.user.id}) }}" TARGET="_NEW">user profile page</a>
```
mickel1138, line 28: This needs a class, you want to control a:link a:visited a:hover etc... Also, don't use the color, just say `color:revert`

mickel1138: I had to deal with this myself, and just `style=color:revert` is enough, no class needed, it doesn't change much more.

PR #1534, opened 2024-02-16
Recreate LDAP on password changes

mickel1138:


fightingtexasaggie:
I've installed this on griidc-test and it appears to work, but I still haven't been able to get Apache Directory Server to connect to griidc-test to verify that the dn had indeed been re-generated.  Will continue to try...

fightingtexasaggie (dismissed):
Tested, and confirmed on Apache Directory Studio that the account had been re-created on password change.

It looks like existing POSIX-enabled accounts get regenerated with everything they need to authenticate griidc-ingest, so I'm pretty confident it'll work.

Not having a test ingest server, we can either hack the current one to authenticate on griidc-test instead of prod, or we can just carefully roll forward with this low-risk change.

PR #1537, opened 2024-02-26
Locates content and copies to clipboard on dblclick event.

fightingtexasaggie:
It does the write thing every time, but generates console error on double-write. It still works, but open for suggestions. dblclick.off has no effect. (maybe ok as is?)


fightingtexasaggie:
@mickel1138  I'm unable to review this per the github workflow, but am good with the changes you made.

mickel1138 (approved):
Works.

PR #1538, opened 2024-02-29
Removed ring shadow on funder.

mickel1138:
Don't show blue outline in input field for funder.


fightingtexasaggie (dismissed):
Tested, works with the ring removed in addt'l funder

fightingtexasaggie (approved):
Reviewed, tested.

PR #1539, opened 2024-03-04
wrap block, not dates

mickel1138:


fightingtexasaggie (approved):
LGTM. Date is now broken onto two lines.

PR #1540, opened 2024-03-04
Also allow enter on textarea

mickel1138:


fightingtexasaggie (approved):
tested, enter working as expected in multi-line fields in dsreview now.

PR #1543, opened 2024-03-06
Added custom assert for udi prefix

mickel1138:
Also added repository with prefix checker.
Custom contraint
custom validator


fightingtexasaggie (commented):
The assert needs to allow selection of the current, unchanged, prefix in the event other fields are being edited for a Funding Cycle. Currently I cannot change any of the other FC details under the FO pods page.

mickel1138:
> The assert needs to allow selection of the current, unchanged, prefix in the event other fields are being edited for a Funding Cycle. Currently I cannot change any of the other FC details under the FO pods page.

Good find, I made it much simpler when I finally remembered how it was done.

fightingtexasaggie (approved):
Now this simplier approach is working perfectly.

PR #1552, opened 2024-04-11
Misc/leaflet maps

fightingtexasaggie:
Leaflet proof-of-concept test - to test //leaflet/<UDI> I'll put real links on the Jira ticket for griidc-test links.


fightingtexasaggie:
can keep branch for future reference, but this was never to be merged in itself.

PR #1558, opened 2024-04-29
Epic/pelagos 4099 create v2 dataset monitoring page

mickel1138:


fightingtexasaggie (approved):
I went through the main and subsites in testing, and nothing looks out of place in the code. I filed a bug on the PDF, but was minor and PO is ok going forward fixing it post-deployment, so approving.

PR #1561, opened 2024-05-09
Feature/pelagos 5345 dataset monitoring status field

fightingtexasaggie:


src/Entity/Dataset.php
```diff
@@ -1035,4 +1044,30 @@ public function __toString(): string
...
+     /**
+     * Get the Dataset's Lifecycle Status
+     *
+     *  Lifecycle Status Definitions:
+     *  None - Dataset has no approved DIF, although UDI has been internally established.
+     *  Identified - Dataset has approved DIF, but no submission
+     *  Submitted - Dataset has a submission, not a draft. Does not have to be approved, nor in-review, etc.
+     *  Accepted - Dataset has a submission, and submission is excepted, and acceptance has not been revoked back to "in review"
+     *  Restricted (subset of Accepted) - Criterial for accepted, but also has the restricted flag set.
+     */
```
fightingtexasaggie, line 1057: Add test!

src/Entity/Dataset.php
```diff
@@ -1035,4 +1044,30 @@ public function __toString(): string
...
+    public function getDatasetLifecycleStatus(string $udi): string
```
mickel1138, line 1058: UDI not needed now.

src/Entity/Dataset.php
```diff
@@ -1035,4 +1044,30 @@ public function __toString(): string
...
+    {
+        if (($this->getDatasetStatus() === Dataset::DATASET_STATUS_ACCEPTED) and ($this->isRestricted() === false)) {
+            $datasetLifeCycleStatus = self::DATASET_LIFECYCLE_STATUS_ACCEPTED;
+        } elseif ($this->getDatasetStatus() === Dataset::DATASET_STATUS_ACCEPTED) {
```
mickel1138, line 1062: You can't guarantee restricted status is `true` here, I would flip them around, and test for Restricted first, it's it not `true` then, then the else must be not true, because now it could be "unset" or "null" or unknown.

src/Entity/Dataset.php
```diff
@@ -55,6 +55,15 @@ class Dataset extends Entity
...
 
+    /**
+     * Valid values for DatasetLifeCycleStatus.
+     */
+    public const DATASET_LIFECYCLE_STATUS_ACCEPTED = 'accepted';
+    public const DATASET_LIFECYCLE_STATUS_RESTRICTED = 'restricted';
+    public const DATASET_LIFECYCLE_STATUS_SUBMITTED = 'submitted';
+    public const DATASET_LIFECYCLE_STATUS_IDENTIFIED = 'identified';
+    public const DATASET_LIFECYCLE_STATUS_NONE = 'none';
+
```
mickel1138, line 66: Maybe use `enum` ?

PR #1564, opened 2024-05-28
Feature/pelagos 5350 make column width for udi and doi narrower

fightingtexasaggie:
shrank slightly, if we're ok with the longer DOIs breaking, I think this could work. The new-issued DOIs all are on a single line.

### DIFF
--- a/templates/DatasetMonitoring/index.html.twig
+++ b/templates/DatasetMonitoring/index.html.twig
@@ -22,7 +22,7 @@
       }
 
       .dsmonitoringcols {
-        white-space: nowrap!important;
+        white-space: nowrap;
       }
     </style>
 {% endblock stylesheets  %}
--- a/templates/DatasetMonitoring/index.html.twig
+++ b/templates/DatasetMonitoring/index.html.twig
@@ -20,6 +20,10 @@
       input[type='text'].dx-texteditor-input:focus {
         box-shadow: none!important;
       }
+
+      .dsmonitoringcols {
+        white-space: nowrap!important;
+      }
     </style>
 {% endblock stylesheets  %}
 
--- a/templates/DatasetMonitoring/index.html.twig
+++ b/templates/DatasetMonitoring/index.html.twig
@@ -24,6 +24,13 @@
       .dsmonitoringcols {
         white-space: nowrap;
       }
+
+      .dx-treelist [tabindex]:not([tabindex="-1"]):focus-visible:focus-visible {
+        outline: none;
+      }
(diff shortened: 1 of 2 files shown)


fightingtexasaggie:
I ended up having to widen UDI a tad to accomodate the NOAA UDIs, because we're not using a fixed-width font for the table.

templates/DatasetMonitoring/v2/researchGroups.html.twig
```diff
@@ -28,11 +28,11 @@
         columns: [{
             dataField: 'udi',
             caption: 'UDI',
-            width: '10rem',
+            width: '10.25rem',
             },{
                 dataField: 'doi',
                 caption: 'DOI',
-                width: '12rem',
+                width: '10.25rem',
```
mickel1138, line 35: In addition could set `white-space:nowrap` to prevent hyphen from wrapping.
Also maybe try: `columnAutoWidth`

fightingtexasaggie: columnAutoWidth worked perfectly. text-overflow: clip is destructive and makes the rest of a long string inaccessible (and non-displayed). With autowidth working right, no text was spilled ;-)

templates/DatasetMonitoring/index.html.twig
```diff
@@ -20,6 +20,10 @@
       input[type='text'].dx-texteditor-input:focus {
         box-shadow: none!important;
       }
+
+      .dsmonitoringcols {
+        white-space: nowrap!important;
```
mickel1138, line 25: Important shouldn't be necessary here.

mickel1138 (approved):

templates/DatasetMonitoring/v2/researchGroups.html.twig
```diff
@@ -28,11 +28,12 @@
         columns: [{
             dataField: 'udi',
             caption: 'UDI',
-            width: '10rem',
+            width: 180,
             },{
                 dataField: 'doi',
                 caption: 'DOI',
-                width: '12rem',
+                width: 252,
```
mickel1138, line 35:
```suggestion
                width: 201,
```

templates/DatasetMonitoring/v2/researchGroups.html.twig
```diff
@@ -28,11 +28,12 @@
         columns: [{
             dataField: 'udi',
             caption: 'UDI',
-            width: '10rem',
+            width: 180,
```
mickel1138, line 31:
```suggestion
            columnAutoWidth: true,
```

mickel1138 (approved):

PR #1566, opened 2024-05-30
Adjust font size on monitoring.

fightingtexasaggie:


fightingtexasaggie:
As we're not using fixed-width fonts for these columns, the NOAA FO datasets have the widest UDIs, so others may have a bit extra space.

templates/DatasetMonitoring/index.html.twig
```diff
@@ -8,9 +8,16 @@
     <link href="/css/dx.generic.griidc-scheme.css" rel="stylesheet">
 
     <style>
+      .dx-widget {
+        font-size: 1rem!important;
+        font-family: dm-sans, Helvetica, sans-serif!important;
```
mickel1138, line 13: I feel we should remove this, because the font is set by the template, and it's set to this.

templates/DatasetMonitoring/index.html.twig
```diff
@@ -8,9 +8,16 @@
...
+      }
+
       .dx-treeview li {
         margin-bottom: revert;
         line-height: 1rem;
+        font-size: 1rem!important;
+        font-family: dm-sans, Helvetica, sans-serif!important;
```
mickel1138, line 20: Same here, remove this.

PR #1570, opened 2024-06-12
Updated caniuse db.

fightingtexasaggie:


fightingtexasaggie:
to be done at deploy time, every time.

PR #1573, opened 2024-06-17
Added onRowClick event to open dataland.

fightingtexasaggie:


templates/DatasetMonitoring/v2/researchGroups.html.twig
```diff
@@ -70,6 +70,9 @@
                     tooltipInstance.hide();
                 });
             }
+        },
+        onRowClick: function(e) {
+            window.open(`${Routing.generate('pelagos_app_ui_dataland_default')}/${ e.data.udi }`);
```
mickel1138, line 75: Add target, I know by default it is blank, but we want to be explicit in case it changes.

mickel1138, line 75:
```suggestion
            window.open(`${Routing.generate('pelagos_app_ui_dataland_default')}/${ e.data.udi }`, '_blank');
```

mickel1138 (approved):

PR #1586, opened 2024-07-11
Update template copyright dates.

fightingtexasaggie:
I just found this, still on 2022. In the future, could we just use {{ "now"|date("Y") }} ?


mickel1138 (approved):
LGTM

PR #1587, opened 2024-07-11
Remove no longer used env vars.

fightingtexasaggie:
Please double check, but I could not find these used anywhere anymore. Furthermore, I think there are still a couple that are made available and referenced by the yaml configs, but no longer used. These here are just the low hanging fruits I think are safe to remove.

DATABASE_NAME is still around only because it's needed in a migration, so I guess it stays unless we update the old migration to reference the new style database DSN to get the db name from.


mickel1138 (approved):
All good to go. Checked.

PR #1590, opened 2024-07-18
Feature/pelagos 5363 test v2 square bobble graph for dataset montoring

mickel1138:


fightingtexasaggie (commented):
Instead of returning datasets, how about a method that just returns counts instead. We're just passing it to |length in the twig template anyway.

PR #1592, opened 2024-07-23
feature/PELAGOS-5368-Create-log-action-item-viewer

fightingtexasaggie:
It would be really nice to show UDI instead of Dataset ID in many cases, but I'm thinking this probably should be implemented as a front-end to API query to resolve into UDI...thoughts? It's functional enough now I think, but that would be nice to have.

### DIFF
--- a/src/Controller/Admin/LogActionItemCrudController.php
+++ b/src/Controller/Admin/LogActionItemCrudController.php
@@ -3,7 +3,6 @@
 namespace App\Controller\Admin;
 
 use App\Entity\LogActionItem;
-use DateTime;
 use EasyCorp\Bundle\EasyAdminBundle\Config\Action;
 use EasyCorp\Bundle\EasyAdminBundle\Config\Actions;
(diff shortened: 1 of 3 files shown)


src/Controller/Admin/LogActionItemCrudController.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+namespace App\Controller\Admin;
+
+use App\Entity\LogActionItem;
+use DateTime;
```
mickel1138, line 6: Where used use \ not use it here.

src/Entity/LogActionItem.php
```diff
@@ -196,4 +196,68 @@ public function getPayLoad()
     {
         return $this->payLoad;
     }
+
+    /**
+     * Virtual function that returns a stringified payload, making sense per json type.
+     *
+     */
```
mickel1138, line 203:
```suggestion

    /**
     * Virtual function that returns a stringified payload, making sense per json type.
     */
```

PR #1593, opened 2024-07-24
feature/PELAGOS-5240-log-session-ID-for-download-count

fightingtexasaggie:


fightingtexasaggie:
I made a significant change. I enabled sessions the moment the /search route is used for non-logged in users. This could have performance implications, but likely not much more than the logging of the search has added, and that's been in place for a long time. This is the only way we can separate multiple searches by the same anonymous user within a 30 second window.

mickel1138 (approved):

PR #1595, opened 2024-07-30
Release/6.59.0

mickel1138:


fightingtexasaggie (approved):
Tested this out on griidc-dev's /opt/pelagos dir in prod mode, all seems to be in order. Build was fine, and changes made appear to be in place.

PR #1613, opened 2024-09-05
Feature/pelagos 5374 grp attributes export

fightingtexasaggie:
This won't open in office, but I think it's syntactically correct.


src/Repository/DatasetRepository.php
```diff
@@ -338,4 +338,25 @@ public function getDatasetWithDoiSet(): array
...
+        if ($this->fundingOrgFilter->isActive()) {
+            $researchGroupIds = $this->fundingOrgFilter->getResearchGroupsIdArray();
+
+            $qb
+            ->innerJoin('dataset.researchGroup', 'rg')
+            ->andWhere('rg.id IN (:rgs)')
+            ->setParameter('rgs', $researchGroupIds);
+        }
+        return $qb->getQuery()->getResult();
+    }
```
mickel1138, line 361: This is not needed, `FundingOrgFilter` has a function that will return a list of ID's, and then you can just do Repo->findBy(arrayOfIds)

Example: https://github.com/griidc/pelagos/blob/develop/src/Controller/DefaultController.php#L82

src/Controller/UI/GRPMetadataExportController.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Class constructor
+     *
+     */
+    public function __construct(SerializerInterface $serializer, NormalizerInterface $normalizer, EncoderInterface $encoder)
+    {
+        $this->serializer = $serializer;
+        $this->normalizer = $normalizer;
+        $this->encoder = $encoder;
```
mickel1138, line 42: If these are not used globally, inject them at the "action".

src/Controller/UI/GRPMetadataExportController.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * This is a parameterless report, so all is in the default action.
+     *
+     * @Route("/grp/export", name="pelagos_app_ui_grpexport_default")
```
mickel1138, line 48: User attribute, not annotation.

src/Controller/UI/GRPMetadataExportController.php
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @return Response A Response instance.
+     *
+     */
+    public function defaultAction(EntityManagerInterface $entityManager)
```
mickel1138, line 53: Inject needed stuff here, not in constructor.
Also, don't inject EntityManager, the controller has one already. Instead inject the repository you need.

src/Controller/UI/GRPMetadataExportController.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        if (!$this->isGranted('ROLE_DATA_REPOSITORY_MANAGER')) {
+            return $this->render('template/AdminOnly.html.twig');
+        }
```
mickel1138, line 57: This is should probably just be available for anyone.

src/Controller/UI/GRPMetadataExportController.php
```diff
@@ -0,0 +1,82 @@
...
+        $datasetRepository = $entityManager->getRepository(Dataset::class);
+        $dataset = $datasetRepository->findAllFiltered();
+
+        /*
+        $serializedData = $this->serializer->serialize($dataset, 'json', [
+            'groups' => ['export'],
+            'json_encode_options' => JSON_PRETTY_PRINT
+        ]);
+        dd($serializedData);
+        */
```
mickel1138, line 70: Debug code.

src/Entity/DatasetSubmission.php
```diff
@@ -2174,6 +2189,27 @@ public function getKeywords(): ?Collection
...
+    {
+        $keywordArray = [];
+        foreach ($this->keywords as $keyword) {
+            $label = $keyword->getLabel();
+            if ($keyword->getType() == KeywordType::TYPE_ANZSRC) {
+                $type = 'anzsrc';
+            } elseif ($keyword->getType() == KeywordType::TYPE_GCMD) {
+                $type = 'gcmd';
+            } else {
+                $type = 'unknown';
```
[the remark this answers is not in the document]
fightingtexasaggie: Thanks. I was looking for that.

src/Controller/UI/GRPMetadataExportController.php
fightingtexasaggie: It also seems normalizer and encoder are not needed for csv only.

src/Controller/UI/GRPMetadataExportController.php
fightingtexasaggie: Calling the getDoctrine() from a controller has been deprecated in favor of Dependency Injection.

src/Controller/UI/GRPMetadataExportController.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * This is a parameterless report, so all is in the default action.
+     *
+     * @return Response A Response instance.
+     */
+    #[Route('/grp/export')]
+    public function defaultAction(EntityManagerInterface $entityManager, SerializerInterface $serializer, FundingOrgFilter $fundingOrgFilter)
+    {
+        $fileName = 'GRP-Metadata.csv' . (new DateTime('now'))->format('Ymd\THis') . '.csv';
```
mickel1138, line 26: Two times .csv in the file name.

src/Controller/UI/GRPMetadataExportController.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+class GRPMetadataExportController extends ReportController
+{
+    /**
+     * This is a parameterless report, so all is in the default action.
+     *
+     * @return Response A Response instance.
+     */
+    #[Route('/grp/export')]
+    public function defaultAction(EntityManagerInterface $entityManager, SerializerInterface $serializer, FundingOrgFilter $fundingOrgFilter)
```
mickel1138, line 24: Instead of EntityManager, is the Repository you need. In this case, DatasetRepository.

src/Entity/DOI.php
```diff
@@ -38,6 +39,7 @@ class DOI extends Entity
      *
      * @Serializer\Groups({"doi"})
      */
+    #[SGroups(['export'])]
```
mickel1138, line 42: I still would like this name changed. SGroups is confusing. Either use just Groups, and Maybe SerializerGroups

src/Entity/DatasetSubmission.php
```diff
@@ -2174,6 +2189,27 @@ public function getKeywords(): ?Collection
...
+        $keywordArray = [];
+        foreach ($this->keywords as $keyword) {
+            $label = $keyword->getLabel();
+            if ($keyword->getType() == KeywordType::TYPE_ANZSRC) {
+                $type = KeywordType::TYPE_ANZSRC->value;
+            } elseif ($keyword->getType() == KeywordType::TYPE_GCMD) {
+                $type = KeywordType::TYPE_GCMD->value;
+            } else {
+                $type = 'unknown';
+            }
```
mickel1138, line 2207: Is the if statement really needed? The type should never be null, or something else. But just in case, you can null coalesce.

```suggestion
            $type = $keyword->getType()?->value ?? 'unset';
```

src/Controller/UI/GRPMetadataExportController.php
fightingtexasaggie: yeah, but I can't inject that, because I need a specific one.

mickel1138: Yes you can, look at other controllers.

mickel1138:
Merge in develop, and remove that canIUse-lite update from here. So there won't be a conflict.

mickel1138 (approved):
LGTM

PR #1615, opened 2024-09-09
Footer removed from PDF base

mickel1138:


fightingtexasaggie (approved):
I know this is what the PO asked for, but without some sort of document footer there's no obvious indication of the last page if someone prints this out. I suggest a "This report run on: <date/time>" type last line for the PDF.

PR #1616, opened 2024-09-09
Add time to file name

mickel1138:


fightingtexasaggie (approved):
I think we're fine leaving timezone out. This sort of report doesn't change _that_ fast.

PR #1624, opened 2024-09-18
GetGML command added

mickel1138:


fightingtexasaggie (approved):
While this is probably not the "correct" way to generate output, it works for our purposes.

PR #1625, opened 2024-09-23
Don't use viewport height

mickel1138:


fightingtexasaggie (approved):
Confirmed this fixes it. It was visible on the dev branch at //research-group/about/846 before.

PR #1633, opened 2024-09-30
Add logged in vs anon in GoMRI stats cmd.

fightingtexasaggie:
This report is getting really complex. I think a better approach for these sorts of periodic reports might be to bite the bullet and use some sort of ETL operation into a format more conducive to writing clean reports, then just write them in clean easy to understand SQL. The combined effort of the ETL and the simplier reports would be less than writing single reports like this.


mickel1138 (approved):
LGTM

PR #1645, opened 2024-10-07
Feature/pelagos 5404 change custom validator annotations

fightingtexasaggie:


mickel1138:
This works. 

1. `NoAngleBrackets` is used in a test, and now fails.
2. It is assumed that `use Symfony\Component\Validator\Constraints as Assert;` exists.
3. The use statement for `NoAngleBrackets` is not removed. The PHP-cs-fixer could handle that.

PR #1651, opened 2024-10-09
Install symfony/doctrine-messenger

fightingtexasaggie:


mickel1138:
It seems this is longer included with symfony/messenger by default, so we have to install it manually.

https://symfony.com/doc/current/messenger.html#doctrine-transport

mickel1138 (approved):

PR #1654, opened 2024-10-10
Fix regex. Negate logic.

fightingtexasaggie:


mickel1138 (approved):
LGTM if this is all find and replace

PR #1658, opened 2024-10-22
Feature/pelagos 5417 fix filer visibility in s6

fightingtexasaggie:
This addresses the permissions on newly created directories created by the filer.

### DIFF
--- a/config/packages/oneup_flysystem.yaml
+++ b/config/packages/oneup_flysystem.yaml
@@ -7,4 +7,6 @@ oneup_flysystem:
     filesystems:
         datastore_flysystem:
             adapter: data_store_adapter
+            visibility: public
+            directory_visibility: public
             #alias: League\Flysystem\Filesystem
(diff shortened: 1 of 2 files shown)


config/packages/oneup_flysystem.yaml
```diff
@@ -7,4 +7,6 @@ oneup_flysystem:
     filesystems:
         datastore_flysystem:
             adapter: data_store_adapter
+            visibility: public
+            directory_visibility: public
```
mickel1138, line 11: This is a suggestion. I added these to the  default config, because it seemed easier to change, with else code, and easier to maintain in the future.

I have not tested, so please test it, with the other code removed.

fightingtexasaggie: I searched the codebase for any place else where we were doing a stream or file write and these are the only places. I'm unsure about the syntax here. I came across documentation showing for each of these sections, you could define the octals for what we want to be public vs private for both directories and files, but I didn't try like this. I'll give it a test.

fightingtexasaggie: Seems to work fine like this.

mickel1138 (approved):

PR #1664, opened 2024-11-05
PELAGOS-5375-When in polygon mode in spatial wizard, generating from line

mickel1138:
Need at least 3 point pairs for a polygon.


fightingtexasaggie (dismissed):
Reviewed with MvdE, it seems to work as expected now.

fightingtexasaggie (approved):
LGTM!

PR #1670, opened 2024-11-12
Use PHP 8.2

mickel1138:


fightingtexasaggie (dismissed):
LGTM!

fightingtexasaggie (dismissed):
LGTM still, probably shouldn't be reviewing my own changes, but this is just the README.md and I'm not sure github would let the original author review.

PR #1674, opened 2024-11-13
Add parameters to query sting and use the to load

mickel1138:


fightingtexasaggie (approved):
Highlighting is working now too! LGTM.

PR #1676, opened 2024-11-14
Added .5em left padding to text boxes and text areas.

fightingtexasaggie:
Seems fine on submission as there already is some padding there.


mickel1138 (approved):
LGTM

PR #1677, opened 2024-11-18
WIP funding cycle land

mickel1138:


fightingtexasaggie:
To test information products, see funding-cycle/about/822

templates/FundingCycleLand/people.html.twig
```diff
@@ -0,0 +1,13 @@
+<h3>People associated with {{ fundingCycle.name }}</h3>
```
mickel1138, line 1: Remove H3 from template

PR #1682, opened 2024-11-26
fix-pobbles

fightingtexasaggie:
I counted and only 68 seem to fit in a line, so pobbles 69-80 were not being shown per line. As an alternative, we could shrink pobbles if we wanted a nice even number.


mickel1138:
LGTM

mickel1138 (approved):

PR #1689, opened 2024-12-06
Updated some templates, added prose.

mickel1138:


fightingtexasaggie (approved):

fightingtexasaggie (approved):
reviewed, but we probably both should take a look as we both made lots of commits

src/Entity/Person.php
```diff
@@ -824,7 +824,7 @@ public function checkDeletable()
      */
     public function __toString()
     {
-        return (string) $this->id;
+        return (string) $this->getFirstName() . ' ' . $this->getLastName();
```
mickel1138, line 827: This is used for Project Director in new Research Group page.

PR #1709, opened 2024-12-20
Need to expose routes

mickel1138:


fightingtexasaggie (approved):
js-router now working. DIF previously was broken (and likely a ton more).

PR #1747, opened 2025-01-21
Add conditional DOI link rendering based on dataset status

mickel1138:


fightingtexasaggie (approved):
Tested and verified DOIs now removed for DIFs

PR #1755, opened 2025-01-23
Bugfix/pelagos 5433 cannot delete dataset with deletion and summary tool

mickel1138:


src/Entity/Dataset.php
```diff
@@ -147,7 +147,8 @@ class Dataset extends Entity
      *
      * @var DatasetSubmission
      */
-    #[ORM\OneToOne(targetEntity: 'DatasetSubmission')]
+    #[ORM\OneToOne(targetEntity: 'DatasetSubmission', cascade: ['remove'])]
```
fightingtexasaggie, line 150: I still don't understand why this is a OneToOne here. The relationship between dataset and dataset submission from any perspective was never a one to one.

mickel1138: Here it is OneToOne because one submission is the active/approved submission, the rest are kept in datasetSubmissionHistory.

fightingtexasaggie (approved):
Tested, double-checked only expected submissions and dif gets deleted. All seems to look good.

PR #1765, opened 2025-01-29
Misc/recipe update jan 2025

mickel1138:


fightingtexasaggie (approved):
Co-reviewed with MvdE earlier. Tested an entire dataset lifecycle...seems solid.

PR #1767, opened 2025-01-29
Add sitemap minimum date feature.

fightingtexasaggie:
This was pretty easy to implement, just stick your baseline date in the .env.local and that'll be reflected in sitemap.xml. We need to discuss whether this is what we need or not for Google.

### DIFF
--- a/src/Controller/DefaultController.php
+++ b/src/Controller/DefaultController.php
@@ -14,12 +14,20 @@ use App\Entity\FundingCycle;
 use App\Entity\PersonResearchGroup;
 use App\Util\FundingOrgFilter;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 
 /**
  * This is the default controller.
  */
 class DefaultController extends AbstractController
 {
+    private $params;
+
+    public function __construct(ParameterBagInterface $params)
+    {
+        $this->params = $params;
+    }
+
     /**
      * Get research groups array in the funding cycle.
      *
@@ -138,12 +146,10 @@ class DefaultController extends AbstractController
      *
      * @param EntityManagerInterface $entityManager    The Doctrine Entity Manager.
      * @param FundingOrgFilter       $fundingOrgFilter The funding organization filter utility.
-     *
-     *
-     * @return StreamedResponse
+     * @param string                 $sitemapMinDate   The minimum date for the sitemap.
      */
     #[Route(path: '/sitemap.xml', name: 'pelagos_sitemap')]
-    public function showSiteMapXml(EntityManagerInterface $entityManager, FundingOrgFilter $fundingOrgFilter)
+    public function showSiteMapXml(EntityManagerInterface $entityManager, FundingOrgFilter $fundingOrgFilter, string $sitemapMinDateParam): StreamedResponse
     {
         $criteria = array(
             'availabilityStatus' =>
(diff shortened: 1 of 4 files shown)


src/Controller/DefaultController.php
```diff
@@ -14,12 +14,20 @@
...
  */
 class DefaultController extends AbstractController
 {
+    private $params;
+
+    public function __construct(ParameterBagInterface $params)
+    {
+        $this->params = $params;
+    }
+
```
mickel1138, line 30: The `ParamaterBagInterface` already exists in the `AbstractController` (actually `ContainerBagInterface`). So no need to add it again.

Just use `$this->getParameter('SITEMAP_MIN_DATE')`

Also $params is not used anywhere.

templates/Default/sitemap.xml.twig
```diff
@@ -3,7 +3,7 @@
 {% for dataset in datasets %}
     <url>
         <loc>{{ vanityurl('pelagos_app_ui_dataland_default', {'udi': dataset.udi}) }}</loc>
-        <lastmod>{{ dataset.modificationTimeStamp|date('Y-m-d') }}</lastmod>
+        <lastmod>{{ (dataset.modificationTimeStamp > sitemapMinDate) ? dataset.modificationTimeStamp|date('Y-m-d') : sitemapMinDate|date('Y-m-d') }}</lastmod>
```
mickel1138, line 6: Consider using the twig function `date` to convert it a Datetime so it can be compared. Formats may vary.

fightingtexasaggie: It is a datetime from the controller.

mickel1138: > It is a datetime from the controller.

Yes, but sitemapMinDate is a `string`

src/Controller/DefaultController.php
```diff
@@ -162,11 +159,18 @@ public function showSiteMapXml(EntityManagerInterface $entityManager, FundingOrg
 
         $datasets = $entityManager->getRepository(Dataset::class)->findBy($criteria);
 
-        $response = new StreamedResponse(function () use ($datasets) {
+        $sitemapMinDateParam = $this->getParameter('sitemap_min_date');
+        if (!is_string($sitemapMinDateParam)) {
+            throw new \InvalidArgumentException('The parameter "sitemap_min_date" must be a valid string.');
```
mickel1138, line 164: Don't throw any exceptions, we want the sitemap to still work if this string is not there, just don't use it in that case. Leave it `null` or adjust your twig logic.

src/Controller/DefaultController.php
```diff
@@ -162,14 +159,34 @@ public function showSiteMapXml(EntityManagerInterface $entityManager, FundingOrg
...
-                )
-            );
-        });
+        try {
+            $sitemapMinDateParam = $this->getParameter('sitemap_min_date');
+        } catch (\Exception) {
+            $sitemapMinDateParam = null;
+        }
+
+        if ($sitemapMinDateParam !== null && is_string($sitemapMinDateParam) && $sitemapMinDateParam !== '') {
```
mickel1138, line 168: This is also known as `empty`

templates/Default/sitemap.xml.twig
```diff
@@ -3,7 +3,7 @@
 {% for dataset in datasets %}
     <url>
         <loc>{{ vanityurl('pelagos_app_ui_dataland_default', {'udi': dataset.udi}) }}</loc>
-        <lastmod>{{ dataset.modificationTimeStamp|date('Y-m-d') }}</lastmod>
+        <lastmod>{{ (sitemapMinDate is defined and sitemapMinDate > dataset.modificationTimeStamp) ? sitemapMinDate|date('Y-m-d') : dataset.modificationTimeStamp|date('Y-m-d') }}</lastmod>
```
mickel1138, line 6: The variable is defined, it can be null. Better to check for `is null` instead.

PR #1781, opened 2025-02-04
Enable display of funder facet in template switch

mickel1138:


fightingtexasaggie (approved):
LGTM, confirmed funder facet shows in HRI template now.

PR #1782, opened 2025-02-05
Add conditional bounding box to ld+json

fightingtexasaggie:
For discussion. We previously were on the DataONE map, so if we want to stay on it we need to include a bounding box in our ld+json. They don't do any other sorts of geometries at this point anyway, and we have a utility that calculates the bounding box, so this was easy enough to implement.  I started the ticket and branch so I could test with it on the DataONE test site.


templates/Dataland/metadata.markup.twig
```diff
@@ -32,5 +32,14 @@
    "datePublished" : "{{dataset.acceptedDate|date('c', 'UTC') }}",
    "@type" : "Dataset",
    "@context" : "http://schema.org"
+   {% if boundingBox is not empty %},
+   "spatialCoverage": {
+      "@type": "Place",
+      "geo": {
+         "@type": "GeoShape",
+         "box": "{{boundingBox|default}}"
```
mickel1138, line 40: The `|default` is not needed, you are checking it for `empty`, so `|default` would make it null is not set.

mickel1138 (dismissed):

fightingtexasaggie:
Waiting on merge until DataONE successfully imports the latest mapping file, that way we don't force a new version of everything on top of the cleanup operation currently in progress.

mickel1138 (approved):

PR #1790, opened 2025-02-10
Added leaflet map to dataland instead of GeoViz

mickel1138:


mickel1138:
Based on: https://github.com/griidc/pelagos/pull/1552

PR #1797, opened 2025-02-17
Update PHP version to 8.3 in workflow files and add matrix support fo…

mickel1138:
…r PHPUnit


mickel1138:
I think the correct upgrade path for this is to upgrade to Symfony 7.2+ then upgrade Psalm.

fightingtexasaggie (dismissed):
I feel this is a pretty safe change, as it really doesn't impact non-build components, besides adding symfony/string. Tested to be safe, all seems solid.

PR #1813, opened 2025-02-26
Refactor message handlers to use constructor property promotion and a…

mickel1138:
…ttributes


fightingtexasaggie (approved):
This looks and tests good. Ready.

PR #1814, opened 2025-02-26
Refactor command constructors to use readonly properties and improve …

mickel1138:
…return types in execute methods


fightingtexasaggie (approved):
LGTM

PR #1815, opened 2025-02-27
API get total size of fileset or cold stored files by udi.

fightingtexasaggie:
I don't know if we want to save this, but might be useful for reports, etc. P.O. had asked to add filesize to an external report and we didn't have a easy way to get it that was cold-storage aware. Ideally our API should support this in the future.


fightingtexasaggie:
I changed the existing API route for file info to add a couple more useful attributes, and also an optional udi to report just a single dataset's file information. Searching, this route isn't in use by anything anyway.

PR #1836, opened 2025-03-12
Made room for 1 more aggregator needed.

fightingtexasaggie:


src/Util/Search.php
```diff
@@ -621,7 +621,7 @@ private function getStatusAggregationQuery(): Aggregation\Terms
     {
         $availabilityStatusAgg = new Aggregation\Terms('status');
         $availabilityStatusAgg->setField('availabilityStatus');
-        $availabilityStatusAgg->setSize(5);
+        $availabilityStatusAgg->setSize(6);
```
mickel1138, line 624: We decided to take this to 7. Because there are 7 different status.

mickel1138 (approved):
LGTM

PR #1860, opened 2025-03-27
Feature/pelagos 5473 improve speed of loading geojson

fightingtexasaggie:


src/Event/DoctrineDatasetListener.php
```diff
@@ -23,14 +24,23 @@ class DoctrineDatasetListener
...
+     *
+     * @var \Memcached
+     */
+    protected $cache;
+
     /**
      * DoctrineDatasetListener constructor.
      *
      * @param MessageBusInterface $messageBus Symfony messenger bus interface.
+     * @param MemcachedAdapter    $cache      Memcached adapter instance.
```
mickel1138, line 38: Not a parameter. And because you calling a static method, you can construct this in the calling function.

src/Event/DoctrineDatasetListener.php
```diff
@@ -78,6 +88,9 @@ protected function updateDataset($entity, EntityManagerInterface $entityManager)
                     $entity->setDatasetFileSize($dataset->getTotalFileSize());
                 }
             }
+            // Invalidate the cache entry for the dataset
+            $cacheKey = 'gml2geojson' . $dataset->getUdi();
+            $this->cache->delete($cacheKey);
```
mickel1138, line 93: We could also recalculate it here again?

src/Util/Geometry.php
```diff
@@ -18,6 +19,11 @@ class Geometry
      */
     protected $entityManager;
 
+    /**
+     * An instance of the high-performance memcached server.
+     */
+    protected \Memcached $memcached;
```
mickel1138, line 25: Should just be private really.

PR #1893, opened 2025-04-15
Added some of the better basemaps.

fightingtexasaggie:


fightingtexasaggie (commented):
MvdE's changes look good to me. Gulf label toggles as expected.

PR #1920, opened 2025-05-08
Misc/add docker compose pelagos 5489

mickel1138:


fightingtexasaggie (approved):
tested against my Docker env, no issues. LGTM.

PR #1939, opened 2025-05-19
Feature/pelagos 5494 remove sftp from sub rev

fightingtexasaggie:


templates/DatasetSubmission/file-submit.html.twig
```diff
@@ -9,22 +9,18 @@
         <legend class="section-header">Dataset File Transfer Details</legend>
         <span>
             The majority of datasets can be uploaded using the File Manager in the "Upload" tab.
-            {% if largefile_tab != 'hidden' %}
-                Extremely large datasets should be uploaded via Globus or SFTP using the "Large Files" Tab.
-            {% endif %}
```
mickel1138, line 14: Is the large file tab going to be removed?

fightingtexasaggie: No, we're leaving test in it's place per the ticket.

PR #1968, opened 2025-06-10
Fix route definition for map search endpoint

mickel1138:


fightingtexasaggie (approved):
reasonable route name

PR #1970, opened 2025-06-11
update deprecations - formtype

mickel1138:
Remove symfony/templating dependency from composer files and update related configurations


fightingtexasaggie (approved):
seems fine

PR #1996, opened 2025-06-28
Misc/remove folder ingest

mickel1138:
I don't know why it does this, but this seriously break my development environment. And calling this always fails, and goes nowhere.


mickel1138:
> I merged this into the test ticket on griidc-test and something seems to be blocking filer.
> 
> I see a lot of the following recently, but this is not a new message, and pre-dates the removal ticket. I think unrelated, but including it here just in case.
> 
> [2025-07-01T10:02:47.286229-05:00] php.ERROR: Warning: require(/opt/data-test.griidc.org/var/cache/prod/ContainerMbNkpaK/getConsole_ErrorListenerService.php): Failed to open stream: No such file or directory {"exception":"[object] (ErrorException(code: 0): Warning: require(/opt/data-test.griidc.org/var/cache/prod/ContainerMbNkpaK/getConsole_ErrorListenerService.php): Failed to open stream: No such file or directory at /opt/data-test.griidc.org/var/cache/prod/ContainerMbNkpaK/App_KernelProdContainer.php:225)"} [

I think fully wiping the cache directory will fix this issue.

PR #2002, opened 2025-06-30
Add EasyAdmin for National Data Center.

fightingtexasaggie:


src/Controller/Admin/NationalDataCenterCrudController.php
```diff
@@ -0,0 +1,91 @@
...
+            })
+        ;
+    }
+
+    /**
+     * Check to see if entity is in use.
+     */
+    private function isNationalDataCenterInUse(NationalDataCenter $nationalDataCenter): bool
+    {
+        $datacenters = $this->entityManager->getRepository(DistributionPoint::class)->findBy(['dataCenter' => $nationalDataCenter]);
```
mickel1138, line 87: No need to construct EntityManager, this is a private function, and you can inject the Repository here instead.

PR #2011, opened 2025-07-11
feature/PELAGOS-5534-fix-doi-logger

mickel1138:
refactor: rename logger variable to doiIssueLogger so it will log to the named log for DOI


mickel1138:
https://symfony.com/doc/current/logging/channels_handlers.html#monolog-autowire-channels

fightingtexasaggie (approved):
I wish there was a way to track all these forced names / magic spells

PR #2016, opened 2025-07-16
Feature/pelagos 5537 change index properties of data repository crud

mickel1138:


fightingtexasaggie (approved):
These changes look good, tested fine.  I do wonder if creator/created/modified would also be useful as readonly in the edit screens, but it wasn't asked for.

PR #2031, opened 2025-07-22
use DOI URL instead of doi://<doi>.

fightingtexasaggie:
This also changes the "cite as:" section in the metadata product. If that is not desired, we'll have to add a parameter to the DOI utility to make the 'https://doi.org/' prefix an option.


mickel1138 (approved):
LGTM

PR #2035, opened 2025-07-30
Remove SFTP tab-select on review/submission. Add default of upload ta…

fightingtexasaggie:
This should prevent the "largefile" tab pre-selection issue. I corrected the tab selector so that "remotely hosted" pre-selects the correct tab. Previously, remotely hosted submissions were lighting up the last, large-files tab mistakingly.


mickel1138 (approved):
Works

PR #2038, opened 2025-07-31
Add cold-storage candidate report.

fightingtexasaggie:
In doing DMS-1149, I realized we need a mechanism to keep up with these cold-storage candidates.


mickel1138 (approved):
Merge solved

PR #2060, opened 2025-09-03
Centered elements over counts and labels.

fightingtexasaggie:
This is the quick fix. What we really need to do is re-write this without vue, but that's a bigger ticket.

### DIFF
--- a/assets/js/vue/components/stats/RepositorySummary.vue
+++ b/assets/js/vue/components/stats/RepositorySummary.vue
@@ -99,6 +99,5 @@ export default {
     display: flex;
     flex-direction: column;
     align-items: center;
-    margin-bottom: 6px;
 }
 </style>


assets/js/vue/components/stats/RepositorySummary.vue
```diff
@@ -93,4 +93,15 @@ export default {
...
+/* Center icons above counts/labels */
+#program-resources .col-sm {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+}
+
+#program-resources .col-sm img.inline-block {
+    margin-bottom: 6px;
+}
```
mickel1138, line 106: Maybe we can give this section a class name, classes tend to stick around longer, and id may be removed because it seems un-needed.

mickel1138 (approved):
LGTM

PR #2081, opened 2025-09-27
Refactor keyword deletion logic to streamline updates in keyword lists

mickel1138:
For some reason onDelete doesn't fire. Maybe because it's not really delete from the list. Instead on using the event, the code is ran on the click.


fightingtexasaggie (commented):
I think it's close, but if I remove a keyword and end a review, and re-open a new review, I initially don't see the keyword removal. If I refresh, I do. If I accept (or end) the review after I refresh (shift F5), the new values get stored as expected, but if I don't refresh the shown values get stored instead.

PR #2087, opened 2025-10-06
Feature/pelagos 5579 add status prose to dashboard

mickel1138:


src/Enum/DatasetLifecycleStatus.php
```diff
@@ -11,6 +11,8 @@ enum DatasetLifecycleStatus: string
     case RESTRICTED = 'Restricted';
     case SUBMITTED = 'Submitted';
     case IDENTIFIED = 'Identified';
+    case INCOMPLETE = 'Incomplete';
+    case PENDING = 'Pending';
```
fightingtexasaggie, line 15: Maybe something that describes the state better, like Difpending would work. I'm a bit concerned it would be confusing to remember that this is a dif-only phase without it being in the name.

PR #2130, opened 2025-12-09
Misc/yarn package update

mickel1138:


fightingtexasaggie (dismissed):
I tested in both dev and prod modes, and I don't see anything breaking. I removed yarnpkg from my dev system, which removed the /usr/bin/yarn command, so added a symlink.

fightingtexasaggie (dismissed):

fightingtexasaggie (approved):
LGTM

PR #2134, opened 2025-12-11
Bugfix/pelagos 5613 fix dif opening from url

fightingtexasaggie:


assets/static/js/dif.js
mickel1138, line 322: This does nearly the same as on line R972-R980. Which on actually fires? Maybe the other one is no longer needed?

fightingtexasaggie: They both fire if passing in ?id=<udi> in the URL, otherwise neither fires on normal load, nor if clicking a DIF on the tree.

mickel1138 (approved):
Removed extra function, tested, GTG

PR #2145, opened 2025-12-17
Constrain list of RGs to those within the funding org filter (subsite)

fightingtexasaggie:
There's likely some bad data already to clean up too. I'll query, generate the list, and send to the PO.


src/Controller/UI/InformationProductController.php
```diff
@@ -28,10 +28,10 @@ class InformationProductController extends AbstractController
      */
     #[IsGranted(Account::ROLE_DATA_REPOSITORY_MANAGER)]
     #[Route(path: '/information-product', name: 'pelagos_app_ui_information_product')]
-    public function index(SerializerInterface $serializer, EntityManagerInterface $entityManager): Response
+    public function index(SerializerInterface $serializer, EntityManagerInterface $entityManager, ResearchGroupRepository $researchGroupRepository): Response
```
mickel1138, line 31: Why not inject all the other repositories too?

PR #2162, opened 2026-01-13
downgrade react-dom to version 18.3.1 for compatibility

mickel1138:


fightingtexasaggie (approved):
yarn.lock matches what's in 6.81.0

PR #2163, opened 2026-01-13
upgrade React and ReactDOM to version 19

mickel1138:


fightingtexasaggie (approved):
I confirmed we only use React for the menu, and it tested fine under R19.

PR #2176, opened 2026-02-03
Moving to php 8.4

fightingtexasaggie:
I'm unsure we should update the composer requirement yet or not.


mickel1138 (approved):
Tested, looks good.

PR #2179, opened 2026-02-04
quote restart directive for podman-compose compatibility.

fightingtexasaggie:


mickel1138 (approved):
This seems to be the recommended way.

More things could be quoted, but this works.

PR #2180, opened 2026-02-04
Some suggested optimizations to speed up webpack.

fightingtexasaggie:
This gets our webpack builds down to about 4 seconds. This was totally co-pilot suggested, so please make sure it's sane.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -57,7 +57,10 @@ Encore
   .addEntry('dashboard', './assets/js/entry/dashboard.js')
 
   // enables Sass/SCSS support
-  .enableSassLoader()
+  .enableSassLoader((options) => {
+    // Speed up Sass by using the embedded implementation
+    options.implementation = require('sass-embedded');
+  })
   .enablePostCssLoader()
 
 // enables the Symfony UX Stimulus bridge (used in assets/bootstrap.js)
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -57,7 +57,10 @@ Encore
   .addEntry('dashboard', './assets/js/entry/dashboard.js')
 
   // enables Sass/SCSS support
-  .enableSassLoader()
+  .enableSassLoader((options) => {
+    // Speed up Sass by using the embedded implementation
+    options.implementation = require('sass-embedded');
+  })
```
mickel1138, line 63: Doesn't seem needed, sass-embedded is installed instead regular sass, and has already speed up the build.

webpack.config.js
```diff
@@ -119,5 +121,18 @@ Encore
...
+
+// Retrieve and tweak the final Webpack config for performance
+const webpackConfig = Encore.getWebpackConfig();
+webpackConfig.cache = {
+  type: 'filesystem',
+  buildDependencies: {
+    config: [__filename],
+  },
+};
+if (!Encore.isProduction()) {
```
mickel1138, line 135: Seems ok, but doesn't gain much time. Also unsure of effects.

https://webpack.js.org/configuration/devtool/

mickel1138:
It speeds up the consecutive `yarn dev` builds, but -- watch build seem to as fast as before.

We should probably read more on this before committing this.

fightingtexasaggie:
maybe later...lots to read up on

PR #2189, opened 2026-02-10
Epic/pelagos 5577 new dif page

mickel1138:


fightingtexasaggie:
not at this time

assets/js/modules/templates/spatialWizard.html
```diff
@@ -0,0 +1,118 @@
+<div
+    id="modalEl"
```
mickel1138: Probably should have better name.

PR #2202, opened 2026-02-17
Rolling back JQuery-UI to restory prior styling behaviour.

fightingtexasaggie:


assets/static/css/dif.css
```diff
@@ -4,7 +4,7 @@ select,textarea {
 
 label:not(.button)
 {
-    font-weight: bold;
+    font-weight: normal;
```
mickel1138: This will override ALL labels, the other label SHOULD be bold, only the labels in the Wizard should not be bold when hovered.

mickel1138 (approved):
Fine now

PR #2207, opened 2026-02-25
Thinner, even-spaced, even-width buttons.

fightingtexasaggie:
Proposed thinner buttons...just a stab at it

### DIFF
--- a/assets/scss/components/buttons.scss
+++ b/assets/scss/components/buttons.scss
@@ -32,4 +32,12 @@
   .button-pill {
     @apply rounded-full bg-blue-500 text-white focus-visible:bg-blue-800 hover:bg-blue-800 py-[0.78rem] px-[1.875rem] h-[2.688rem] text-sm;
   }
+
+  /*
+  Button form, for GRIIDC-styled buttons
+  */
+  .button-griidc-form,
+  a.button-griidc-form {
+    @apply box-border h-[2.275rem] min-h-[2.275rem] max-h-[2.275rem] py-0;
+  }
 }
--- a/assets/scss/dif.scss
+++ b/assets/scss/dif.scss
@@ -4,12 +4,3 @@
 
 @import 'components/inputs.scss';
 @import 'components/buttons.scss';
-
-.dif-buttons {
-	box-sizing: border-box;
-	height: 2.275rem !important;
-	min-height: 2.275rem !important;
-	max-height: 2.275rem !important;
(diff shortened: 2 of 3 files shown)


assets/scss/dif.scss
```diff
@@ -3,4 +3,13 @@
...
+@import 'components/buttons.scss';
+
+.dif-buttons {
+	box-sizing: border-box;
+	height: 2.275rem !important;
+	min-height: 2.275rem !important;
+	max-height: 2.275rem !important;
+	padding-top: 0 !important;
+	padding-bottom: 0 !important;
+}
```
mickel1138: Instead of this, open `buttons.scss` and add another defined class, like `button-form` or something, and use tailwind to style it, instead of straight CSS. This way it can be re-used.

assets/scss/components/buttons.scss
```diff
@@ -32,4 +32,12 @@
   .button-pill {
     @apply rounded-full bg-blue-500 text-white focus-visible:bg-blue-800 hover:bg-blue-800 py-[0.78rem] px-[1.875rem] h-[2.688rem] text-sm;
   }
+
+  /*
+  Button form, for GRIIDC-styled buttons
+  */
+  .button-griidc-form,
+  a.button-griidc-form {
```
mickel1138, line 40: `a` is not needed, there is no hyperlink button.

assets/scss/components/buttons.scss
```diff
@@ -32,4 +32,12 @@
...
+    @apply box-border h-[2.275rem] min-h-[2.275rem] max-h-[2.275rem] py-0;
```
mickel1138, line 41: Instead of using absolute REM, you could use `h-8` the 8 is calculate from the Spacing variable. Spacing 8 is about 2rem. Spacing is part of the Theme.

assets/scss/components/buttons.scss
```diff
@@ -32,4 +32,12 @@
   .button-pill {
     @apply rounded-full bg-blue-500 text-white focus-visible:bg-blue-800 hover:bg-blue-800 py-[0.78rem] px-[1.875rem] h-[2.688rem] text-sm;
   }
+
+  /*
+  Button form, for GRIIDC-styled buttons
+  */
+  .button-griidc-form,
```
mickel1138, line 39: I would remove the word `griidc`, these are all our buttons.

PR #2237, opened 2026-03-25
Feature/pelagos 5646 confirmation page when dif is created saved

mickel1138:


templates/DIF/dif-confirmation.html.twig
```diff
@@ -0,0 +1,72 @@
...
+        {% for message in app.flashes('error') %}
+            <div class="p-4 mb-4 text-sm text-rose-900 rounded-xl bg-rose-50" role="alert">
+                <span class="font-medium">{{ message }}</span>
+            </div>
+        {% endfor %}
+        <div>
+            <section class="bg-white py-8 antialiased md:py-16">
+            <div class="mx-auto max-w-2xl px-4 2xl:px-0">
+                <h2 class="text-xl font-semibold text-gray-900 sm:text-2xl mb-4">
+                {% if dataset.dif.status == constant('App\\Entity\\DIF::STATUS_SUBMITTED') %}
```
fightingtexasaggie, line 28: breaks if the udi param is null, which is allowed on the route. Also breaks if udi passed is not found.

fightingtexasaggie (approved):

PR #2243, opened 2026-03-31
Add approve and Unlock buttons to DIF for DRPMs and show current state.

fightingtexasaggie:


templates/DIF/dif-confirmation.html.twig
mickel1138, line 1: With this many changes, might as well a separate drpm-confirmation page.  This looks confusing.

src/Controller/UI/DIFController.php
```diff
@@ -141,17 +141,54 @@ public function difTwo(Request $request, FormFactoryInterface $formFactory, Data
...
+                        $submitAction = $buttonName;
+                        break;
+                    }
+                }
+            }
+
+            if ($saveAndSubmit->isClicked()) {
+                $submitAction = 'saveAndSubmit';
+            }
+
```
mickel1138: Can be removed, this mechanism doesn't work because of the validator.

src/Controller/UI/DIFController.php
```diff
@@ -141,17 +141,54 @@ public function difTwo(Request $request, FormFactoryInterface $formFactory, Data
...
             }
 
             $entityManager->persist($dif);
             $entityManager->persist($dataset);
 
             $entityManager->flush();
 
-            $this->addFlash('success', 'DIF Successfully Submitted. Your UDI is: ' . $dataset->getUdi());
-
+            $this->addFlash('success', $successMessage);
```
mickel1138: Do this conditionally.

fightingtexasaggie:
Merging in now as we can address the mouseover/title stuff later in the epic or as an improvement. We need this for DIF stuff to be testable.

PR #2257, opened 2026-04-16
Feature/pelagos 5652 dif field placeholder text

fightingtexasaggie:


src/Entity/DIF.php
```diff
@@ -39,7 +39,7 @@ class DIF extends Entity
     /**
      * Valid data sizes for $dataSize.
      */
-    public const DATA_SIZES = ['< 1GB', '1GB-10GB', '10GB-200GB', '200GB-1TB', '1TB-5TB', '>5TB'];
+    public const DATA_SIZES = ['< 10GB', '10GB-200GB', '200GB-1TB', '1TB-5TB', '>5TB'];
```
mickel1138: The ticket same remove DEFAULT selection, and remove the entire option. The default for this is set int the DIF constructor.

migrations/Version20260416193932.php
mickel1138, line 1: Not needed, read next comment. Also read ticket closely.

assets/static/js/dif.js
mickel1138, line 1: This is the OLD DIF.

assets/js/entry/dif.js
```diff
@@ -48,15 +48,32 @@ document.addEventListener('DOMContentLoaded', () => {
...
     plugins: {
-      clear_button: {
-        title: 'Remove all selected options',
+      remove_button: {
+        title: 'Remove this funder',
       },
     },
   });
 
+  function syncFunderPlaceholder() {
```
mickel1138: Use: https://tom-select.js.org/docs/#hideplaceholder

templates/DIF/dif.v2.html.twig
```diff
@@ -29,6 +31,13 @@
             color: rgb(17 24 39 / var(--tw-text-opacity, 1));
         }
 
+        .ts-control .ts-placeholder,
```
mickel1138: Use only the selectors that are relevant.
Maybe using just .ts-control would be ok here.

templates/DIF/dif.v2.html.twig
```diff
@@ -29,6 +31,12 @@
             color: rgb(17 24 39 / var(--tw-text-opacity, 1));
         }
 
+        .ts-control input.items-placeholder,
+        .ts-control input#funders-ts-control {
+            font-size: 0.875rem !important;
+            line-height: 1.25rem !important;
+        }
```
mickel1138, line 42: Check these statements. Especially where a `comma` is used. Look redundant. Also, where you used `flex` maybe instead of adding that to `.ts-control` it needs to be in it's on more specific selector, so it doesn't change every one of them.

mickel1138: Github only marked one section, but you should all the CSS here.

templates/DIF/dif.v2.html.twig
```diff
@@ -42,6 +40,11 @@
             font-size: 0.875rem !important;
             line-height: 1.25rem !important;
         }
+        .ts-control:has(#funders-ts-control) {
+            display: flex !important;
+            flex-wrap: wrap !important;
+        }
+
```
mickel1138, line 47: If you are targeting an ID, just use and ID selector.

PR #2271, opened 2026-04-23
Add container definition for developer enviroment.

fightingtexasaggie:
This will build a full symfony-enabled container with PHP 8.4 on a Rocky-9 base image for development use.
Port 8000 maps out to view in a browser.


fightingtexasaggie:
This isn't needed in the codebase. I added some local files for now, and updated the .git/info/exclude for a localized .gitignore.

PR #2279, opened 2026-04-29
Feature/pelagos 5649 add text to spatial extent wizard

mickel1138:


assets/js/modules/templates/spatialWizard.html
```diff
@@ -0,0 +1,117 @@
+<div
+    id="modalEl"
+    tabindex="-1"
+    aria-hidden="true"
```
mickel1138, line 4: This seems not correct for ARIA, throws a console warning.

PR #2287, opened 2026-05-06
Feature/pelagos 5651 pop up on dif

fightingtexasaggie:


assets/js/entry/dif.js
```diff
@@ -11,6 +11,8 @@ import JustValidatePluginDate from 'just-validate-plugin-date';
 import Routing from '../../../vendor/friendsofsymfony/jsrouting-bundle/Resources/public/js/router.min';
 
 import GeoViz from '../modules/geoViz';
+import FullScreenModal from '../modules/fullScreenModal';
+import { cookie } from 'jquery';
```
fightingtexasaggie: avoid using jquery if at all possible!

assets/js/modules/fullScreenModal.js
```diff
@@ -0,0 +1,112 @@
...
+import modalTemplate from './templates/fullScreenModal.hbs';
+import Handlebars from 'handlebars';
+
+export default class FullScreenModal {
+  constructor(options = {}) {
+    this.modalInstance = null;
+    this.cookieName = options.cookieName || 'generic-modal-acknowledged';
+    this.title = options.title || 'Modal Title Placeholder';
+    this.content = options.content || 'Modal Content Placeholder';
+    this.supplementalCss = options.supplementalCss || null;
```
mickel1138: All the `this` will be available to the class, that mean that `FullScreenModal.title` is accessible from the caller. Which is ok if you want that, but it's better to use setters. So I would recommend not using `this` but `const` or `let` (if needed)

assets/js/modules/fullScreenModal.js
```diff
@@ -0,0 +1,112 @@
...
+      .split(';')
+      .map((cookie) => cookie.trim())
+      .some((cookie) => cookie.startsWith(`${this.cookieName}=`));
+
+    if (isAcknowledged) {
+      return;
+    }
+
+    const newElement = document.createElement('div');
+    newElement.innerHTML = Handlebars.compile(modalTemplate)({ title: this.title, content: this.content });
```
mickel1138: Maybe construct the options elsewhere for readability.

assets/js/modules/fullScreenModal.js
```diff
@@ -0,0 +1,112 @@
...
+    const modalElement = newElement.firstChild;
+    document.body.appendChild(modalElement);
+
+    const fontSource = document.getElementById('pelagos-content') || document.body;
```
mickel1138: You should not expect `pelagos-content` to be there, this class has no knowledge of this. I wouldn't even bother setting the font, just inherit what is there, or set it explicitly.

assets/js/modules/templates/fullScreenModal.hbs
```diff
@@ -0,0 +1,23 @@
+<div class="bg-white rounded-lg shadow-lg flex flex-col h-full">
```
mickel1138: There is no ID here?

assets/js/modules/fullScreenModal.js
```diff
@@ -0,0 +1,112 @@
...
+
+    modalElement.querySelector('#modal-close-button-for-good').addEventListener('click', () => {
+      this.hideModal(modalElement, modalInstance);
+      document.cookie = `${this.cookieName}=1; path=/; max-age=31536000; SameSite=Lax`;
+    });
+
+    showModal();
+  }
+
+  applySupplementalCss(modalElement) {
```
mickel1138: Huh?

assets/js/modules/fullScreenModal.js
```diff
@@ -0,0 +1,112 @@
...
+        return;
+      }
+
+      modalElement.querySelectorAll(selector).forEach((element) => {
+        Object.assign(element.style, ruleSet);
+      });
+    });
+  }
+
+  hideModal(modalElement, modalInstance) {
```
mickel1138: Make `Modal` (from Flowbite) do this.

assets/js/entry/dif.js
```diff
@@ -8,9 +8,8 @@ import 'tom-select/dist/css/tom-select.css';
 import JustValidate from 'just-validate';
 import JustValidatePluginDate from 'just-validate-plugin-date';
 
-import Routing from '../../../vendor/friendsofsymfony/jsrouting-bundle/Resources/public/js/router.min';
-
```
mickel1138, line 15: Why was this commented out, but you are using it?

assets/js/entry/dif.js
```diff
@@ -310,4 +309,39 @@ document.addEventListener('DOMContentLoaded', () => {
         document.getElementById('spatialExtentGeometry').value = gmlOutput;
       });
   });
+
+  const dashboardUrl = window.Routing.generate('app_ui_dashboard');
```
mickel1138: Just because you know it's available from somewhere, you can't rely on it being here. Uncomment routing, use that one. You don't know what other modules might do. Also webpack will make sure Routing is only loaded once, even if every module needs it, and includes it.

assets/js/modules/fullScreenModal.js
fightingtexasaggie: 🤦🏼 removed that mess

assets/js/entry/dif.js
```diff
@@ -310,4 +310,33 @@ document.addEventListener('DOMContentLoaded', () => {
...
+        If you would like to work on a previously saved DIF, please use the
+        <a href="${dashboardUrl}" class="pagelink" target="_blank">Dashboard</a> to load
+        the DIF.
+      </p>
+    `,
+  });
+
+  window.addEventListener('load', () => {
+    fullScreenModal.show();
+  });
```
mickel1138, line 381: You are already in content loaded event.

PR #2303, opened 2026-05-21
Feature/pelagos 5666 ensure dif emails work

fightingtexasaggie:


assets/js/entry/dif.js
```diff
@@ -207,6 +207,13 @@ document.addEventListener('DOMContentLoaded', () => {
       successEvent.currentTarget.submit();
     });
 
+  const saveAndContinueButton = document.getElementById('saveAndContinue');
```
mickel1138: By doing this, you bypass the validator, depending on the order where this code is.

If you look at the `.onSuccess` of the validator, you can see it already does something like this.

PR #2324, opened 2026-05-28
Added extra Doctrine directive to prevent Symfony 8 upgrade issues.

mickel1138:


fightingtexasaggie (approved):
No deprecations show, so It seems leaving enable_lazy_ghost_objects in place doesn't hurt anything.

PR #2325, opened 2026-05-28
Collective Symfony update to 7.4.12+

mickel1138:


fightingtexasaggie (approved):
test merge back into develop showed some peer dependency "incorrectly met" messages, but we can fix this after the merge.

PR #2342, opened 2026-06-11
Widen out UDI column.

fightingtexasaggie:


fightingtexasaggie:
image
Here's sized for the widest UDI ever possible

fightingtexasaggie:
image

The only difference is the column widths. Zoom is sticky per-hostname and is a browser setting.

mickel1138 (approved):
LGTM

PR #2343, opened 2026-06-12
Feature/pelagos 5692 grp infoprod report

fightingtexasaggie:
Added WIP tag as Rosalie is communicating with GRP on this.

### DIFF
--- a/.env
+++ b/.env
@@ -57,6 +57,7 @@ GCMD_VERSION="16.9"
 GCMD_VERSION_YEAR="2023"
 ANZSRC_VERSION="2020"
 EXPORT_REVIEW_PATH='/mnt/inspect-review-dataset'
+FUNDING_ORG_RGRPS_DSREPORT='National Academies of Science, Engineering, Medicine: Gulf Research Program'
 
 ###> INDEX NAMES ###
 DISCOVERY_INDEX=pelagos_instancename
--- a/.env
+++ b/.env
@@ -57,7 +57,6 @@ GCMD_VERSION="16.9"
 GCMD_VERSION_YEAR="2023"
 ANZSRC_VERSION="2020"
 EXPORT_REVIEW_PATH='/mnt/inspect-review-dataset'
-FUNDING_ORG_RGRPS_DSREPORT='National Academies of Science, Engineering, Medicine: Gulf Research Program'
 
 ###> INDEX NAMES ###
 DISCOVERY_INDEX=pelagos_instancename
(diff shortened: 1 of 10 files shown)


src/Controller/UI/ReportResearchGroupDatasetStatusController.php
```diff
@@ -201,6 +203,57 @@ public function getGrpResearchGroupReport(Request $request, EntityHandler $entit
...
+        if (!$this->isGranted('ROLE_DATA_REPOSITORY_MANAGER')) {
+            return $this->render('template/AdminOnly.html.twig');
+        }
+
+        $csv = [
+            ['Research Group ID', 'Research Group Name', 'Information Products Count', 'Information Product Tags'],
+        ];
+
+        // Fetch all Research Groups.
+        $fundingOrganization = $entityManager->getRepository(FundingOrganization::class)->findOneBy(['name' => 'National Academies of Science, Engineering, Medicine: Gulf Research Program']);
```
fightingtexasaggie: Use the .env ecosystem instead of hardcoding!

src/Entity/InformationProduct.php
```diff
@@ -612,4 +612,30 @@ public function removeFunder(Funder $funder): self
 
         return $this;
     }
+
+    /**
+     * Show strings of tags used - this is the product type descriptor and digital resource type descriptor names.
+     */
+    #[Serializer\VirtualProperty]
```
fightingtexasaggie: Remove this until such time as it is needed to be serialized.

src/Entity/InformationProduct.php
```diff
@@ -612,4 +612,30 @@ public function removeFunder(Funder $funder): self
...
+    public function getTags(): string
```
fightingtexasaggie: This conflates producttype and digital resourcetype. Check with PO.  User simply asked for "tags used" in the request.

src/Entity/InformationProduct.php
```diff
@@ -612,4 +612,27 @@ public function removeFunder(Funder $funder): self
 
         return $this;
     }
+
+    /**
+     * Show tags used as single string, combining the product type descriptor and digital resource type descriptor names.
+     */
+    public function getTags(): string
```
mickel1138: Maybe return an Array, or Collection (ArrayCollection)

mickel1138: Or move to Report.

.env
```diff
@@ -57,6 +57,7 @@ GCMD_VERSION="16.9"
 GCMD_VERSION_YEAR="2023"
 ANZSRC_VERSION="2020"
 EXPORT_REVIEW_PATH='/mnt/inspect-review-dataset'
+FUNDING_ORG_RGRPS_DSREPORT='National Academies of Science, Engineering, Medicine: Gulf Research Program'
```
mickel1138: Probably not need here. Also name is volatile.  Just hardcode it in the report may be better.

fightingtexasaggie: Yeah, I was on the fence about doing this anyway. I'll hardcode the ID in the report instead, and add a comment.

mickel1138 (approved):
Looks good.

PR #2361, opened 2026-06-17
fix: remove redundant jQuery require to avoid redeclaration with auto…

fightingtexasaggie:
This fixes the issue with the JQuery re-declaration and the visibility needed so the CDN-based includes still see what they're expecting. This does not address the merge conflict in yarn.lock, which can now just be resolved by re-generating a new yarn.lock on this epic after merging this in.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,6 +31,9 @@ module.exports = {
     ],
     'import/no-cycle': 'off',
   },
+  globals: {
+    "$": "readonly"
+  },
   settings: {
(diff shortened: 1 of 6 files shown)


fightingtexasaggie:
Once this is merged into the epic, merge in develop then regenerate the yarn.lock. There should be no errors, nor any eslint errors after the merge.

mickel1138:
Not needed right now, instead disable `provideGlobaljQuery' in webpack-encore.

PR #2386, opened 2026-07-07
Darker lines, increased top padding to center.

fightingtexasaggie:


templates/DatasetSubmission/datasetSubmissionForm.v2.html.twig
```diff
@@ -43,7 +43,7 @@
                         {{ form_widget(contact.primaryContact) }}
                         {{ form_row(contact.role, { attr: { class: 'input-base' } }) }}
                     </div>
-                    <hr class="my-2 border-gray-300">
+                    <hr class="mt-5 border-gray-600">
```
mickel1138: I usually stick with even numbers, but I don't know how it should look.

fightingtexasaggie: 5 seemed perfectly centered as I was matching the existing space. 4 was still a little off.

PR #2387, opened 2026-07-08
Used tippy.js to display definitions.

fightingtexasaggie:
Implementation partially suggested by Claude Opus 4.5. We might want to hold off on merging until discussing to prevent a merge conflict with other submission-related branches, however changes are localized and fairly minimal.


mickel1138 (changes requested):
Going with Tom-Select

PR #2395, opened 2026-07-13
Misc/update deploy scripts

fightingtexasaggie:
bin/deploy takes forever, and if there's no good reason to reload elasticsearch, we should not be. Also, I removed the old drupal-related code we've not needed since the Drupal Divorce.


mickel1138 (approved):
Good

PR #2417, opened 2026-07-30
Feature/pelagos 5719 move all boosts from es index to es queries

fightingtexasaggie:
This ES7 backwards-compatible update migrates boosts to the Elasticsearch queries. We'll need this before we can switch to ES8.

### DIFF
--- a/src/Search/MultiSearch.php
+++ b/src/Search/MultiSearch.php
@@ -96,7 +96,32 @@ class MultiSearch
         $queryString = $searchOptions->getQueryString();
         $specifiedField = empty($searchOptions->getField()) ? [] : [$searchOptions->getField()];
 
-        $simpleQuery = new Query\SimpleQueryString($queryString, $specifiedField);
+        $simpleQuery = new Query\SimpleQueryString($queryString);
+
+        // Boosts have moved from index to the queries as index boosts are no longer supported since
+        // moving to Elasticsearch 8. This boosts title, Place/Theme keywords, & DOIs to 3x weight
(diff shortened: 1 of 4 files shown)


src/Search/MultiSearch.php
```diff
@@ -96,7 +96,32 @@ public function search(SearchOptions $searchOptions): SearchResults
...
 
-        $simpleQuery = new Query\SimpleQueryString($queryString, $specifiedField);
+        $simpleQuery = new Query\SimpleQueryString($queryString);
+
+        // Boosts have moved from index to the queries as index boosts are no longer supported since
+        // moving to Elasticsearch 8. This boosts title, Place/Theme keywords, & DOIs to 3x weight
+        // IFF a specific field isn't being queried. Boost only makes sense to for broad queries.
+        if (empty($specifiedField)) {
+            $simpleQuery->setFields([
+                'title^3',
```
mickel1138, line 106: A little more complicated, but it's better to use the `FunctionScoreQuery` boosting can be done with: 
https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-function-score-query#function-field-value-factor

src/Search/MultiSearch.php
mickel1138, line 212:
```suggestion
        $functionScoreQuery = new FunctionScore();
        $functionScoreQuery->setQuery($query->getQuery());
        $functionScoreQuery->addFieldValueFactorFunction(
            field: 'title',
            factor: 2.0
        );
        $resultsPaginator = $this->finder->findPaginated($functionScoreQuery);
```

PR #2427, opened 2026-08-05
Remove stale pnpm lockfile as we're yarn 4.x now.

fightingtexasaggie:
This is a leftover from the Dec 9th yarn 4 conversion, and contains stale entries I suspect Dependabot is keying in on. In any event, it's not needed in the repo currently.


mickel1138 (approved):
LGTM

PR #2428, opened 2026-08-05
We can update existing postcss 8, leaving the vue-2 bound 7.x ver unt…

fightingtexasaggie:
…ouched

We could do this so the 8.x version we have is current, leaving only vue-locked version to itself to eventually go away with vue? This is all build-time only (7.x and 8.x) as user input doesn't touch it, so we're not vulnerable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,13 @@
     },
     "license": "UNLICENSED",
     "private": true,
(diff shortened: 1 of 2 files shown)


fightingtexasaggie:
Upon discussion, decided to leave it be.

PR #2439, opened 2026-08-10
Remove 'Move' option from Vue2 file manager and supporting code.

fightingtexasaggie:
removes the (broken) 'move to' option in vue2-based file manager


mickel1138:
It seems we have the code for it behind this function, why does it not work?

mickel1138:
We could just merge this to Develop instead, this is not just an EPIC bug.

PR #2441, opened 2026-08-12
Update dev container to Postgis-18

fightingtexasaggie:


docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   database:
-    image: postgis/postgis:14-3.5
+    image: postgis/postgis:18-master
```
mickel1138: The version should probably use `18.4` for the version. So the version is stable. `18-master` has more frequent updates.

mickel1138 (approved):
LGTM

PR #2442, opened 2026-08-12
Update status API to include PHP and database version information

mickel1138:


src/Controller/StatusController.php
```diff
@@ -80,9 +83,10 @@ private function getDatabaseEngineStatus(): ServiceStatus
         $serviceStatus = new ServiceStatus();
         try {
             $connection = $this->entityManager->getConnection();
-            $connection->executeQuery('SELECT 1');
+            $result = $connection->executeQuery('SELECT version();');
```
fightingtexasaggie: In case you just wanted just the '18.4' without the platform, container, and compilation details:
```suggestion
            $result = $connection->executeQuery('SELECT split_part(current_setting(\'server_version\'), \' \', 1);');
```

fightingtexasaggie (dismissed):
added suggestion if you wanted to shorten the DB version number down, but good either way

src/Controller/StatusController.php
mickel1138: It recommends we user `server_version_num` and don't parse.
https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-INFO-VERSION

mickel1138:
If we are concerned about giving away "secret" information, we exposing some paths to our file system.

fightingtexasaggie (dismissed):
Tested, LGTM!

src/Controller/StatusController.php
```diff
@@ -77,12 +53,15 @@ public function index(): Response
...
         try {
             $connection = $this->entityManager->getConnection();
-            $connection->executeQuery('SELECT 1');
-            $serviceStatus->setStatus(ServiceStatus::STATUS_OK);
-            $serviceStatus->setData(['Database connection' => 'Successful']);
+            $result = $connection->executeQuery("SELECT current_setting('server_version_num')");
+            $isConnected = $connection->isConnected();
+            $fetchedVersion = (int)$result->fetchOne();
+            // server_version_num is formed by multiplying the server's major version number by 10000 and adding the minor version number.
+            $decodedVersion = (int)round($fetchedVersion / 10000) . '.' . ($fetchedVersion % 10000);
```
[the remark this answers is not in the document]
mickel1138: Applications might use this function to determine the version of the database server they are connected to. The result is formed by multiplying the server's major version number by 10000 and adding the minor version number. For example, version 10.1 will be returned as 100001, and version 11.0 will be returned as 110000. Zero is returned if the connection is bad.

Prior to major version 10, PostgreSQL used three-part version numbers in which the first two parts together represented the major version. For those versions, [PQserverVersion](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQSERVERVERSION) uses two digits for each part; for example version 9.1.5 will be returned as 90105, and version 9.2.0 will be returned as 90200.

Therefore, for purposes of determining feature compatibility, applications should divide the result of [PQserverVersion](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQSERVERVERSION) by 100 not 10000 to determine a logical major version number. In all release series, only the last two digits differ between minor releases (bug-fix releases).

fightingtexasaggie (dismissed):

fightingtexasaggie (approved):

PR #2444, opened 2026-08-13
Insure RHEL-9 systemd service compatability for pelagos-messengemq.

fightingtexasaggie:
proteus.tamucc.edu is RHEL-9 based, and we should update all our RHEL-8 systems to RHEL-9 as it's supported until 2032. Proteus is slated to replace griidc-dev (RHEL-8). For RHEL-9 compatibility, we needed to update our .service files accordingly. This will also work in RHEL-8.


mickel1138 (approved):
LGTM

PR #2447, opened 2026-08-14
Misc/add port to loaddump

fightingtexasaggie:
The current loaddump outright ignores the specified port and just assumes 5432. This is incredibly dangerous if you have multiple DBs running, as commonly used during migrations.

I also removed some really old scripts long since defunct.


mickel1138 (approved):
LGTM

PR #2455, opened 2026-08-25
Release/6.88.0

mickel1138:


fightingtexasaggie (approved):
Cursory testing is good, and confirmed the only difference from the test branch (minus the epic) is eslint. (compared release/6.88.0  with commit hash 42722e627bee06a86289904a0767720aa1376dc7)

PR #2456, opened 2026-08-25
Rolling back eslint/js from 10 to 9

mickel1138:
Could lead to issues without having eslint 10


fightingtexasaggie (approved):
lg, safe change

PR #2471, opened 2026-09-01
Feature/pelagos 5724 allow same start end dates - submission

fightingtexasaggie:


assets/js/entry/dataset-submission.js
```diff
@@ -516,9 +516,9 @@ document.addEventListener('DOMContentLoaded', () => {
       {
         plugin: JustValidatePluginDate((fields) => ({
           format: 'yyyy-MM-dd',
-          isAfter: fields['#temporalExtentBeginPosition'].elem.value,
+          isAfterOrEqual: fields['#temporalExtentBeginPosition'].elem.value,
         })),
-        errorMessage: 'Date must be after start date.',
+        errorMessage: 'Date must be after (or on) start date.',
```
mickel1138: Silly change maybe: `Date can not be before start date?`

assets/js/entry/dataset-submission.js
```diff
@@ -656,6 +656,15 @@ document.addEventListener('DOMContentLoaded', () => {
...
   });
 
+  const revalidateDatesOnBlur = () => {
+    if (estimatedStartDate.value.trim() && estimatedEndDate.value.trim()) {
+      formValidate.revalidateField('#temporalExtentBeginPosition');
+      formValidate.revalidateField('#temporalExtentEndPosition');
+    }
+  };
+  estimatedStartDate.addEventListener('blur', revalidateDatesOnBlur);
+  estimatedEndDate.addEventListener('blur', revalidateDatesOnBlur);
```
mickel1138, line 666: These are now the third and fourth event that to the same thing. `blur` is similar to `change` as an event, maybe this could all just be be one event. And instead of using ID's, a class can be used. But you would have to iterate trough them (`each`). Maybe change can be removed?

PR #2475, opened 2026-09-02
Feature/pelagos 5720 add locked mode to geoviz spatial wizard

mickel1138:


fightingtexasaggie (approved):
Seems like a reasonable method to lock

